PR #73, opened 2023-04-12
Folding sidebar functionality

hunar1312:
@DonKoko, Please review this PR and let me know your feedback.
Thanks.


app/components/layout/sidebar/index.tsx
```diff
@@ -14,7 +17,12 @@ interface Props {
   user: User;
 }
 
+const sidebarCollapseStatus = atom(true);
```
DonKoko: This is good, but I try to follow a general rule that when creating the atom that i append Atom to the name. So in this case it will be `const sidebarCollapseStatusAtom = atom(true);`

app/components/layout/sidebar/index.tsx
```diff
@@ -26,6 +40,11 @@ export default function Sidebar({ user }: Props) {
       to: "categories",
       label: "Categories",
     },
+    {
```
DonKoko: I think you can remove that for now as we have no route to link it to. It will be the next thing to build but not needed for now.

app/components/layout/sidebar/index.tsx
```diff
@@ -34,44 +53,117 @@ export default function Sidebar({ user }: Props) {
...
+  const collapseSidebar = useCallback(() => {
+    setIsSidebarCollapsed(true);
+  }, []);
+  const uncollapseSidebar = useCallback(() => {
+    setIsSidebarCollapsed(false);
+  }, []);
+
+  useEffect(() => {
+    if (sidebarSwitchIsClicked === false) {
+      const sidebar = document.getElementById("main-navigation") as HTMLElement;
```
DonKoko: No accessing of the dom with document. You should use a ref.

app/components/layout/sidebar/index.tsx
```diff
@@ -34,44 +53,117 @@ export default function Sidebar({ user }: Props) {
...
+      sidebar.addEventListener("mouseenter", uncollapseSidebar);
+      sidebar.addEventListener("mouseleave", collapseSidebar);
+    }
+  });
```
DonKoko: You dont have a depencency, so that useEffect will run on every render, which it shouldn't. Only needs to run when `sidebarSwitchIsClicked` is changing. Also I think we can actually avoid those use-effects by using the jotai atoms better and doing some setter/getter atoms.

Also if you are adding event listeners in a useEffect you have to clean them up in the return, but yeah in general you should not be adding them like this. Should use `onMouseEnter` and `onMouseLeave`. 
With this in mind, we can create a few derived attoms which are setters and either fold or unfold the sidebar. https://jotai.org/docs/core/atom#atom

So to give some examples, here is what I would do.


```ts
/** Defining atoms */
const sidebarCollapsedAtom = atom(false);
/** Writtable only attom that uses a setter to change toggle the state */
const toggleSidebarAtom = atom(null, (_get, set) =>
  set(sidebarCollapsedAtom, !sidebarCollapsedAtom)
);

  /** How to use them */
  const isCollapse = useAtomValue(sidebarCollapsedAtom);
  const [,toggleSidebar] = useAtom(toggleSidebarAtom)

```

And in the JSX
```jsx
<aside
  onMouseEnter={toggleSidebar}
  onMouseLeave={toggleSidebar}
```

Doing it like this you can remove all of the event handlers you are adding, you can remove the useEffects and useCallbacks where you are defining your functions. A few reasons to do it like this:

1. Logic stays outside of the component
2. You are not using useEffect which can cause some undesired stuff and based on latest React conventions should only be used when interacting with 3rd party data.


Moreover we have to manage the button that maintains the collapsed state. I am pretty sure we can do it with some derived atoms as well, maybe we dont even need its own atom. Would be easier for me to figure that out by testing so I would suggest lets make the suggestion i have work first, and then after we can do this. I think it will be also a great exercise for you in the art of Jotai and atomic state.

app/styles/layout/sidebar.css
```diff
@@ -4,3 +4,20 @@
 .bottom {
   @apply flex justify-between border-t border-t-slate-400 pt-6;
 }
+
+#main-navigation {
```
DonKoko: We don't use id's for styling. Also I am pretty sure all of this can be achieved with some tailwind classes.

app/components/layout/sidebar/index.tsx
```diff
@@ -14,7 +20,18 @@ interface Props {
   user: User;
 }
 
+const componentMountedAtom = atom(false);
+const sidebarCollapseStatus = atomWithStorage("isSidebarCollapsed", true);
+const isSidebarSwitchClicked = atomWithStorage("sidebarSwitchIsClicked", false);
+
 export default function Sidebar({ user }: Props) {
+  const [hasMounted, setHasMounted] = useAtom(componentMountedAtom);
```
DonKoko: After reviewing this and thinking a bit further about it, i realize i made a mistake. This will never work as I want it to because of SSR and we will always have a flickr. Even if we use `<Suspense/>` we have no way of predicting if the sidebar will be open or closed so it will never be as smooth as I would like it to be. 
That being said, we drop this functionality for now. I know you spent some time on it but sometimes that happens.

app/components/layout/sidebar/index.tsx
```diff
@@ -34,45 +56,135 @@ export default function Sidebar({ user }: Props) {
...
+        <Link to="." title="Home" className="flex items-center">
           <img
-            src="/images/logo-full-color(x2).png"
+            src="/images/shelf-symbol.png"
             alt="Shelf Logo"
-            className="h-[30px]"
+            className="mx-1.5 inline h-[32px]"
           />
+          <span className="logo-text transition duration-200 ease-linear">
+            <svg
```
DonKoko: Lets move this to a separate component as those big svg's make the component look very messy.

DonKoko (changes requested):
Okey so as I mentioned on discord I am 85% happy, just want to improve and clean up the code as well. You can see my comments here and resolve them one by one. 
I think we will have some more improvements to do in the next review but I find it easier like this step by step.

app/components/layout/sidebar/index.tsx
hunar1312: Should I move this to /components/icons/library or create a separate component in /components/icons

app/components/layout/sidebar/index.tsx
hunar1312: Yes, I can understand. Will drop this functionality

app/components/layout/sidebar/index.tsx
hunar1312: Yes, you're right. Sounds great. Will utilise Jotai resources to achieve this functionality. Thanks for guiding

app/components/layout/sidebar/index.tsx
DonKoko: Move the SVG to the library and just render it within the span in the view. That should be a good enough solution.

DonKoko (approved):
Looks good. We can merge this.

Issue #86, opened 2023-04-24
Download QR code UI


DonKoko:
Related to #79 we need to also do the ui for `Download QR tag`.

<img width="755" alt="Screenshot 2023-04-24 at 14 38 22" src="https://user-images.githubusercontent.com/7840007/233985572-b4b29400-699c-4b41-b135-b3867727aefb.png">

Works with a sidebar that opens when you click the "Download QR tag" button.
Needs to be animated with framer-motion.
Here is just a quick and dirty example of how it would work via chat gpt:

```ts
import React, { useState } from "react";
import { motion } from "framer-motion";

const Sidebar = () => {
  const [isOpen, setIsOpen] = useState(false);

  const toggleSidebar = () => {
    setIsOpen(!isOpen);
  };

  return (
    <div>
      <button onClick={toggleSidebar}>Toggle Sidebar</button>
      <motion.div
        initial={{ x: "-100%" }}
        animate={{ x: isOpen ? 0 : "-100%" }}
        transition={{ duration: 0.5 }}
        style={{
          width: "300px",
          height: "100%",
          backgroundColor: "#333",
          position: "fixed",
          top: 0,
          left: 0,
          zIndex: 999,
          padding: "20px",
          boxSizing: "border-box",
        }}
      >
        <h2 style={{ color: "#fff" }}>Sidebar Content</h2>
        <p style={{ color: "#fff" }}>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed auctor
          felis at mauris vehicula eleifend. Sed vitae dapibus arcu. Praesent
          ut ipsum et velit faucibus dapibus a at risus. Nulla facilisi.
        </p>
      </motion.div>
    </div>
  );
};

export default Sidebar;
```

Needs to of course be improved to use atom's and to also render the overlay that shows on the side of the QR code thingy.
Let me know if you have any questions. Should be a pretty easy and straightforward thing to build.

Once #83 is finished, please create a new branch as copy of `79-view-item` and work on it there.

PR #90, opened 2023-04-26
initial UI development of #86 and merged dev branch

hunar1312:
@DonKoko UI development of #86 is completed. Please review it

Thanks.


app/components/icons/library.tsx
```diff
@@ -634,3 +634,24 @@ export function HorizontalDotsIcon(props: SVGProps<SVGSVGElement>) {
     </svg>
   );
 }
+
+export function CrossIcon(props: SVGProps<SVGSVGElement>) {
```
DonKoko: We already have an icon for this. Its called `XIcon`

app/components/items/download-qr-code.tsx
```diff
@@ -0,0 +1,70 @@
...
+      document.body.classList.remove("overflow-hidden")
+    }
+  }, [isSidebarOpen]);
+
+  return (
+    <div>
+      <Button icon="barcode" onClick={toggleSidebar} variant="secondary">
+        Download QR Tag
+      </Button>
+      <div className={tw("fixed right-0 top-0 z-10 h-screen w-screen bg-gray-25/70 backdrop-blur transition duration-300 ease-in-out", isSidebarOpen ? "visible" : "invisible opacity-0")}></div>
```
DonKoko: Clicking on this should also toggle the sidebar, so if its open and the user clicks on the blury overlay, it should close it.

app/components/items/download-qr-code.tsx
```diff
@@ -0,0 +1,70 @@
...
+  (get) => get(sidebarStatusAtom),
+  (get, set) => set(sidebarStatusAtom, !get(sidebarStatusAtom))
+);
+
+const DownloadQrCode = () => {
+  const [isSidebarOpen, toggleSidebar] = useAtom(toggleSidebarAtom);
+
+  useEffect(()=>{
+    if(isSidebarOpen) {
+      document.body.classList.add("overflow-hidden")
```
DonKoko: What is the purpose of this? Is it to disabled the scrolling when the sidebar is open? 
I any case, using document is not okey. Especially considering we will have a URL that corresponds to the sidebar being open, so that means there will be hydration issues. Can you please just comment that line and once i know how the back-end will be build we will solve it. Thanks

DonKoko (changes requested):
Some small changes required.

app/components/items/download-qr-code.tsx
hunar1312: Yes, the purpose is to disable scrolling when the sidebar is open.
Ahhh yes, we will be having a URL that corresponds to the sidebar being open. so, okay will comment this out for now.
Thanks.

app/components/items/download-qr-code.tsx
hunar1312: okay

DonKoko (approved):
Good to merge with `dev`. This is just the front-end.

Issue #91, opened 2023-04-26
Implementing mobile designs on main-navigation Sidebar


hunar1312:
The main-navigation sidebar is already developed for desktop. Needs to be responsive for all screens.

Design to be implemented: https://www.figma.com/file/2XkiKJ4II7Mux6JJL9hYTm/Shelf-App-%7C-Production-Ready?node-id=5534-343707&t=BbNGgKv7oDqD9Cpt-4

PR #92, opened 2023-04-27
made main-navigation-sidebar responsive

hunar1312:
Hello @DonKoko,
The desktop main navigation sidebar is now responsive for all screen sizes as described in the issue #91
Please Review this. Thanks

### DIFF
--- a/app/components/layout/sidebar/atoms.ts
+++ b/app/components/layout/sidebar/atoms.ts
@@ -1,8 +1,8 @@
 import { atom } from "jotai";
 
-export const sidebarCollapseStatusAtom = atom(false);
+export const sidebarCollapseStatusAtom = atom(true);
 
-export const keepSidebarUncollapsedAtom = atom(true);
+export const keepSidebarUncollapsedAtom = atom(false);
 
 /* Controls the state for whether the sidebar is collapsed or not */
 export const toggleSidebarAtom = atom(
--- a/app/components/layout/sidebar/atoms.ts
+++ b/app/components/layout/sidebar/atoms.ts
@@ -1,8 +1,8 @@
 import { atom } from "jotai";
 
-export const sidebarCollapseStatusAtom = atom(true);
+export const sidebarCollapseStatusAtom = atom(false);
 
-export const keepSidebarUncollapsedAtom = atom(false);
+export const keepSidebarUncollapsedAtom = atom(true);
 
 /* Controls the state for whether the sidebar is collapsed or not */
 export const toggleSidebarAtom = atom(
@@ -20,4 +20,12 @@ export const maintainUncollapsedAtom = atom(
     set(keepSidebarUncollapsedAtom, !get(keepSidebarUncollapsedAtom));
     set(sidebarCollapseStatusAtom, !get(keepSidebarUncollapsedAtom));
   }
+);
+
+/* Using different atoms for mobile navigation sidebar as they conflict with desktop's sidebar atoms state*/
+export const isMobileNavOpenAtom = atom(false);
+
+export const toggleMobileNavAtom = atom(
+  (get) => get(isMobileNavOpenAtom),
+  (get,set) => set(isMobileNavOpenAtom, !get(isMobileNavOpenAtom))
 );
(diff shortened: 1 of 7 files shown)


app/components/layout/sidebar/atoms.ts
```diff
@@ -1,8 +1,8 @@
 import { atom } from "jotai";
 
-export const sidebarCollapseStatusAtom = atom(false);
```
DonKoko: I had changed those as I prefer the default state to be open.

app/components/layout/sidebar/index.tsx
```diff
@@ -48,68 +46,114 @@ export default function Sidebar({ user }: Props) {
...
-          </span>
         </Link>
-        <button
-          className="transition-all duration-200 ease-linear"
-          onClick={manageUncollapsedSidebar}
-        >
-          <i className="icon text-gray-500">
-            {maintainUncollapsedSidebar ? <ActiveSwitchIcon /> : <SwitchIcon />}
-          </i>
+        <button className="menu-btn relative z-50" onClick={toggleSidebar}>
```
DonKoko: Please make a component for this.

app/components/layout/sidebar/index.tsx
```diff
@@ -48,68 +46,114 @@ export default function Sidebar({ user }: Props) {
...
-      <div className="mt-auto">
-        <SidebarBottom
-          user={{
-            ...user,
-          }}
-        />
-      </div>
-    </aside>
+      </header>
+      <div
```
DonKoko: This works really good but its super hard to read like this, as its difficult to figure which html is which element. Can you please place this in a separate component.

app/components/layout/sidebar/index.tsx
```diff
@@ -48,68 +46,114 @@ export default function Sidebar({ user }: Props) {
...
+              {maintainUncollapsedSidebar ? (
+                <ActiveSwitchIcon />
+              ) : (
+                <SwitchIcon />
+              )}
+            </i>
+          </button>
+        </div>
+        <div className="flex-1">
+          <ul className="menu mt-6 md:mt-10">
```
DonKoko: Can you please make a component MenuItems and put this part of the code there.

app/routes/_layout+/_layout.tsx
```diff
@@ -4,7 +4,7 @@ import type {
   LoaderFunction,
 } from "@remix-run/node";
 import { json } from "@remix-run/node";
-import { Outlet, useLoaderData } from "@remix-run/react";
+import { Link, Outlet, useLoaderData } from "@remix-run/react";
```
DonKoko: Here is a linting error.

DonKoko (changes requested):
Overall it looks and works good @hunar1312. I really like it.
Just 1 thing, I see you have used the same atom to control the "keep opened" on desktop and the mobile menu. This works but I would rather use different atoms for it for a few reasons:

1. I want the default state to be 'open' which only makes sense on desktop and if we do it like this, on mobile the menu will be open by default and we dont want that.
2. In the near future when we add some functionality to save this in the user settings, we will again have a conflicts with that.

app/components/layout/sidebar/index.tsx
```diff
@@ -48,68 +46,114 @@ export default function Sidebar({ user }: Props) {
...
+            {menuItems.map((item) => (
+              <li key={item.label}>
+                <NavLink
```
DonKoko: Here we need to close the menu on click on mobile. So if the user navigates they actually see the page and dont have to manually close the menu.

app/components/layout/sidebar/atoms.ts
hunar1312: Okay wasn't aware of that. Will use different atoms for mobile.

app/components/layout/sidebar/index.tsx
hunar1312: @DonKoko do you want me to convert this overlay into a component or the whole header?

DonKoko: @hunar1312 The overlay

DonKoko (approved):
Love it! Great job @hunar1312.

PR #101, opened 2023-05-04
100 qr scanning

DonKoko:


DonKoko:
@carlosvirreira @jurrejansen @hunar1312 merging the qr scanning functionality with `dev` branch. Some parts might still be kinda jank, so it needs some thorough testing.

PR #268, opened 2023-08-01
138 implementing optimistic UI

hunar1312:
Made adding notes UI follow Optimistic UI approach


app/components/assets/notes/notes.tsx
```diff
@@ -1,18 +1,43 @@
...
 import type { NoteWithDate } from "./note";
 import { Note } from "./note";
 
 export const Notes = () => {
   const { asset } = useLoaderData();
+  const user = useUserData();
   const hasNotes = asset?.notes.length > 0;
-
+  const fetcher = useFetcher();
+  let onSubmissionContent = "";
```
DonKoko, line 16: Can you add a comment here with what is going on. The code is not immediately obvious so I think its important for people who look at this code to understand it.

app/components/assets/notes/notes.tsx
```diff
@@ -1,18 +1,43 @@
...
+  if (fetcher.formData) {
+    for (const data of fetcher.formData.entries()) {
+      onSubmissionContent = data[1].toString();
+    }
+  }
   return (
     <div>
-      <NewNote />
+      <div>{JSON.stringify(fetcher)}</div>
```
DonKoko: This has to be cleaned up.

DonKoko (changes requested):
Some changes are required. Overall great job.

app/components/assets/notes/notes.tsx
hunar1312: sure

app/components/assets/notes/notes.tsx
hunar1312: sorry for this one

DonKoko (approved):

DonKoko (approved):

PR #298, opened 2023-08-11
updating readme which had the wrong name for assets

DonKoko:


README.md
```diff
@@ -222,8 +222,8 @@ For File storage we use the S3 buckets service provided by supabase. We do this
 
 ### Items
```
timvdalen, line 223:
```suggestion
### Assets
```

(for clarity?)

DonKoko: You are correct. I re-adjusted it and deploying now again.

PR #300, opened 2023-08-11
including roles in seeds as well as changing how user is created in seed

DonKoko:


timvdalen:
Does this create the org? I'm not super familiar with the codebase yet, but I don't see how it does.

Long term, there should probably be a startup script that interactively lets you configure this stuff (username, org name, etc), but a seeder would have helped me avoid a 500 during initial setup!

DonKoko:
> Does this create the org? I'm not super familiar with the codebase yet, but I don't see how it does.

Yes, it uses the `createUser` function that we use when someone registers, that function also creates a related organization.

> Long term, there should probably be a startup script that interactively lets you configure this stuff (username, org name, etc), but a seeder would have helped me avoid a 500 during initial setup!

Yeah I agree. I have lots of ideas for it like a setup wizzard but we haven't yet invested that much in the DX on that side, but it will come as well.

Issue #296, opened 2023-08-10
[Feature request]: Updated appearance 'Actions' dropdown on mobile + Activate after scan


jurrejansen:
### Contact Details

jurre@whale-agency.com

### Is your feature request related to a problem? Please describe?

In most cases users want to perform an action after scanning a code. This will be one less action to perform an action.

### Describe the solution you'd like

After scanning a QR code a user will land on the asset page. In most cases they scan the code to perform an action. For that reason it would be good to already have the 'Actions' menu activated after they scan.

Currently the 'Actions' dropdown on mobile is styled like this:

https://github.com/Shelf-nu/shelf.nu/assets/76112272/164999b5-3de5-457f-8b20-f048501140d3

A more mobile friendly appearance would be:

![actions-dropdown-mobile](https://github.com/Shelf-nu/shelf.nu/assets/76112272/d837f0b1-59f3-494e-8c73-658e9a3f2f71)

rishi-raj-jain:
I'm passing on this one.

Reason: Too much to do to even setup.

PR #304, opened 2023-08-19
Issue 296 active dropdown when scan qr

evonloch:
/claim #296 
1. Active dropdown when scan qr, only on mobile screen. 
2. Follow the design guide provided.

### DIFF
--- a/app/components/assets/actions-dropdown.tsx
+++ b/app/components/assets/actions-dropdown.tsx
@@ -1,6 +1,12 @@
-import { useEffect } from "react";
+import { useState } from "react";
 import type { Asset } from "@prisma/client";
-import { ChevronRight } from "~/components/icons";
+import {
+  ChevronRight,
+  TrashIcon,
+  PenIcon,
+  UserIcon,
+  UserXIcon,
+} from "~/components/icons";
 import {
   DropdownMenu,
   DropdownMenuContent,
@@ -19,24 +25,13 @@ interface Props {
   isQr: boolean;
 }
 
-export const ActionsDopdown = ({ asset, isQr }: Props) => {
+const ActionsDopdown = ({ asset, isQr }: Props) => {
   const assetIsAvailable = asset.status === "AVAILABLE";
-  var isMobileScreen;
-
-  useEffect(() => {
-    if (window.innerWidth <= 768) {
-      isMobileScreen = true;
-    } else {
-      isMobileScreen = false;
-    }
-  }, []);
-
-  console.log("from actions-dropdown.tsx: ");
-  console.log("isMobileScreen: ", isMobileScreen);
-  console.log("isQr: ", isQr);
+  const defaultOpen = window.innerWidth <= 640 && isQr;
+  const [open, setOpen] = useState(defaultOpen);
 
   return (
-    <DropdownMenu modal={false} defaultOpen={isQr && isMobileScreen}>
+    <DropdownMenu modal={false} defaultOpen={open}>
       <DropdownMenuTrigger className="asset-actions">
         <Button variant="secondary" to="#" data-test-id="assetActionsButton">
           <span className="flex items-center gap-2">
@@ -46,7 +41,7 @@ export const ActionsDopdown = ({ asset, isQr }: Props) => {
       </DropdownMenuTrigger>
       <DropdownMenuContent
         align="end"
-        className="order w-[180px] rounded-md bg-white p-0 text-right "
+        className="order relative w-screen rounded-lg bg-white p-0 text-right md:static md:w-[180px]"
       >
         <DropdownMenuItem className="border-b px-6 py-3">
           {!assetIsAvailable ? (
@@ -57,7 +52,9 @@ export const ActionsDopdown = ({ asset, isQr }: Props) => {
               className="justify-start text-gray-700 hover:text-gray-700"
               width="full"
             >
-              Release Custody
+              <span className="flex items-center gap-2">
+                <UserXIcon /> Release Custody
+              </span>
             </Button>
           ) : (
             <Button
@@ -67,7 +64,9 @@ export const ActionsDopdown = ({ asset, isQr }: Props) => {
               className="justify-start text-gray-700 hover:text-gray-700"
               width="full"
             >
-              Give Custody
+              <span className="flex items-center gap-2">
+                <UserIcon /> Give Custody
+              </span>
             </Button>
           )}
         </DropdownMenuItem>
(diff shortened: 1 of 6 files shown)


evonloch:
[screen-recorder-sat-aug-19-2023-18-10-31.webm](https://github.com/Shelf-nu/shelf.nu/assets/22785213/0e72fdd7-a073-433f-9aa9-defaad4c12cf)

app/components/assets/delete-asset.tsx
```diff
@@ -27,7 +27,7 @@ export const DeleteAsset = ({
       <Button
         variant="link"
         data-test-id="deleteAssetButton"
-        className="justify-start rounded-sm px-6 py-3 text-sm font-semibold text-gray-700 outline-none data-[disabled]:pointer-events-none data-[disabled]:opacity-50 hover:bg-slate-100 hover:text-gray-700"
+        className="justify-start rounded-sm px-2 py-3 text-sm font-semibold text-gray-700 outline-none data-[disabled]:pointer-events-none data-[disabled]:opacity-50 hover:bg-slate-100 hover:text-gray-700"
```
DonKoko: I think those changes broke the desktop view. The spacing is wrong, the icons are cropped out.
Screenshot 2023-08-21 at 11 09 38

Please check how its done on the `location` page. There the icons are styled and sized properly.

app/components/assets/actions-dropdown.tsx
```diff
@@ -15,12 +22,16 @@ interface Props {
     mainImage: Asset["mainImage"];
     status: Asset["status"];
   };
+  isQr: boolean;
 }
 
-export const ActionsDopdown = ({ asset }: Props) => {
+const ActionsDopdown = ({ asset, isQr }: Props) => {
   const assetIsAvailable = asset.status === "AVAILABLE";
+  const defaultOpen = window.innerWidth <= 640 && isQr;
```
DonKoko: This causes an error. What I would suggest we do is use state for this with default value of false. Then you need an effect that runs, checks if `window` exists and only then set it to the correct value.
Screenshot 2023-08-21 at 11 11 25

app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -136,6 +140,11 @@ export const links: LinksFunction = () => [
 ];
 
 export default function AssetDetailsPage() {
+  const queryLocation = useLocation();
+  const queryParams = new URLSearchParams(queryLocation.search);
```
DonKoko: This works but I would like to refactor it to do it the 'proper' Remix way. Please check this: https://remix.run/docs/en/main/hooks/use-search-params

So you can do:
```ts

const [searchParams] = useSearchParams()
const isQr = searchParams.get("ref") === "qr"
```

app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -179,7 +188,9 @@ export default function AssetDetailsPage() {
           View QR code
         </Button>
 
-        <ActionsDopdown asset={asset} />
+        <Suspense fallback={null}>
```
DonKoko: This is a proper solution. I saw you asked about it but I couldn't asnwer before you solved it. We usually use the `<ClientOnly>` component from remix-utils: https://github.com/sergiodxa/remix-utils
However the suspense solution requires to lazy load as well as it can have issues with React 18 and SSR. See this: https://remix.run/docs/en/1.19.3/guides/migrating-react-router-app#reactlazy-and-reactsuspense and this https://remix.run/docs/en/1.19.3/guides/migrating-react-router-app#client-only-components. So having said this I would prefer to replace it with the ClientOnly component.


One thing I want you to do is add the fallback. It should visually look exactly the same as the actions button. We need it because without it we are getting flickering and layout shift.

Also, you dont need to do this in this component. You can do all of this inside the `<ActionsDropdown>` component as you can use the hooks there. Then you dont need to pass `isQr` you can just do it inside the component. 
Last but not least, i think `isQr` is a kinda confusing name. Its too generic. Better do a longer name but that is more descriptive for example: `refIsQrScan`

app/components/assets/actions-dropdown.tsx
```diff
@@ -15,12 +22,16 @@ interface Props {
     mainImage: Asset["mainImage"];
     status: Asset["status"];
   };
+  isQr: boolean;
 }
 
-export const ActionsDopdown = ({ asset }: Props) => {
+const ActionsDopdown = ({ asset, isQr }: Props) => {
```
DonKoko: When I test it on my phone, it looks like this:
IMG_19916787F46E-1

The menu should be fixed to the bottom of the screen.

DonKoko (approved):
Overall good but some changes are required. Let me know on discord if you have some questions.

app/components/assets/delete-asset.tsx
evonloch: Fixed it.

jurrejansen:
https://github.com/Shelf-nu/shelf.nu/assets/76112272/181b8e0e-ed27-43a5-bc57-b629113beef9

- [ ] - close actions dropdown after selecting an action
- [ ] - 'select team member' dropdown is floating at top, would be nice if it opens below the 'select team members' select field

evonloch:
[screen-recorder-wed-aug-23-2023-03-45-01.webm](https://github.com/Shelf-nu/shelf.nu/assets/22785213/baed96a7-35f2-455d-b18c-ef1f89245a1e)

evonloch:
I think we need to work with give custody modal also. May in another issue.

app/components/assets/actions-dropdown.tsx
```diff
@@ -28,75 +28,103 @@ const ConditionalActionsDropdown = ({ asset }: Props) => {
...
-        asChild
-        align="end"
-        className="order static w-screen rounded-lg bg-white p-0 text-right md:static md:w-[180px]"
-      >
-        <div className="order fixed bottom-0 left-0 w-screen rounded-lg bg-white p-0 text-right md:static md:w-[180px]">
-          <DropdownMenuItem className="border-b px-6 py-3">
-            {!assetIsAvailable ? (
+      <figure>
+        {open && (
+          <style
```
DonKoko: Okey so I don't like this at all so I checked the code to see why what we were trying to do was not working. 
You mentioned that this code:

```css
@media (max-width: 640px) {
  [data-radix-popper-content-wrapper] {
    transform: none !important;
  }
}
```

Is breaking the dialog we use for custody that Jurre mentioned. I did check it and thats not the case at all. Its just that the modal doesnt have a max-width so on mobile it was overflowing.

So please do the following:
 - [ ] ~restore the css code in `asset.css`~
 - [ ] in `custom-modal.css` add `max-w-full` to the dialog class.

This will resolve the issue.

DonKoko (changes requested):
I added a comment in the file with the actions. Some changes are required.

app/components/assets/actions-dropdown.tsx
DonKoko: @evonloch so I checked it further based on your comment on github and you are right. Was my mistake I was mistaken what the error referred to. So honestly I dont like the solution with the inline styles but at the moment due to how radix works I cannot find another easy solution so we can leave it like that. So you can ignore my previous comment and please do the following:
 - [ ] Not sure why you added the `<figure>`. I think that could be removed. If its needed for some specific reason, please let me know
 - [ ] Can you still do this: in `custom-modal.css` add `max-w-full` to the dialog class.
 - [ ] Please add a comment in the code where we have the style explaining why it works like this with a link to this PR, so when I see it in 1 year and I am wondering "What is this?" I can easily view the PR it was related to.

Once this is done, we can merge and close the issue.

evonloch: fixed

DonKoko (approved):

PR #308, opened 2023-08-21
developed UI for subscriptions page and created successful subscription modal

hunar1312:


app/components/subscription/prices.tsx
```diff
@@ -47,56 +55,97 @@ export const Price = ({
   const activePlan = activeSubscription?.items.data[0]?.plan;
   const isFreePlan = price.id != "free";
   const features = price.product.metadata.features?.split(",") || [];
+
+  // icons to be mapped with different plans
+  const plansIconsMap: { [key: string]: JSX.Element } = {
```
DonKoko, line 61: 🤟🏼

app/components/subscription/prices.tsx
```diff
@@ -47,56 +55,97 @@ export const Price = ({
...
+    free: <SingleLayerIcon />,
+    tier_1: <DoubleLayerIcon />,
+    tier_2: <MultiLayerIcon />,
+  };
+
+  function shortenIntervalString(str: string): string {
```
DonKoko: This is nice, but can you move it to a util file and import it from there. i dont like defining it inside the component.

app/components/subscription/prices.tsx
```diff
@@ -47,56 +55,97 @@ export const Price = ({
...
+              </div>
             ) : null}
           </div>
-        </>
-      ) : null}
-      <div>
-        <i>{price.product.metadata.slogan}</i>
+          {price.unit_amount != null ? (
+            <div className=" mb-3 text-4xl font-semibold text-gray-900">
+              {price.currency === "usd"
```
DonKoko: I think there is a better way to do it. You should be able to use `toLocaleString` and pass a currency and this will format it accordingly by placing the correct char in front or behind.

app/components/subscription/successful-subscription-modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { Button } from "../shared/button";
+
+export default function SuccessfulSubscriptionModal() {
+  const [params] = useSearchParams();
+  const navigate = useNavigate();
+  const success = params.get("success") || false;
+  const handleBackdropClose = useCallback(
+    (e: React.MouseEvent<HTMLDivElement>) => {
+      if (e.target !== e.currentTarget) return;
+      navigate("/settings/subscription");
```
DonKoko: This works and is a really good way to do it in remix, but I tend to use navigate when the actual path changes. And searchParams, even tho they are part of the link they don't cause navigation. 
Thankfully remix has a super easy way to do it. the `useSearchParams` hook also has a setter.
`const [searchParams, setSearchParams] = useSearchParams()`

and then you can use `setSearchParams` to clean them up instead of using navigate. Both will work the same but this seems a better way as what we want to logically do is simply clear the params, not navigate from 1 url to another.

hunar1312: that's super cool! setting params using remix hooks. I'll check that out and will update that. Thanks for letting me know @DonKoko.

Issue #310, opened 2023-08-22
[Feature request]: Eye icon to see a password 


carlosvirreira:
### Is your feature request related to a problem? Please describe?

"As I type my password, I sometimes do now know what I had written" it would be helpful for me to be able to toggle

### Describe the solution you'd like

<img width="224" alt="Screenshot 2023-08-22 at 17 24 47" src="https://github.com/Shelf-nu/shelf.nu/assets/12449255/88e18a7e-414b-4e50-85d3-ebf09d98cc1f">

zcesur:
hey @DonKoko, zaf from algora here. currently only the [shelf members](https://console.algora.io/org/Shelf-nu/members) are authorized to create bounties

you can sign up on algora with a @shelf.nu address to auto join, or sign up with a regular email and have @carlosvirreira add you to the org

DonKoko:
Thanks @zcesur . I was just writing the team that I assume thats the problem. I will make an account now so I can place bounties.

jurrejansen:
@evonloch here you go
[icons.zip](https://github.com/Shelf-nu/shelf.nu/files/12410494/icons.zip)

Here they are in code if you rather use an embed.

`<svg width=" 100%" height=" 100%" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M2.42012 12.7132C2.28394 12.4975 2.21584 12.3897 2.17772 12.2234C2.14909 12.0985 2.14909 11.9015 2.17772 11.7766C2.21584 11.6103 2.28394 11.5025 2.42012 11.2868C3.54553 9.50484 6.8954 5 12.0004 5C17.1054 5 20.4553 9.50484 21.5807 11.2868C21.7169 11.5025 21.785 11.6103 21.8231 11.7766C21.8517 11.9015 21.8517 12.0985 21.8231 12.2234C21.785 12.3897 21.7169 12.4975 21.5807 12.7132C20.4553 14.4952 17.1054 19 12.0004 19C6.8954 19 3.54553 14.4952 2.42012 12.7132Z" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"/>
[...]

PR #313, opened 2023-08-23
310 eye icon to see a password

evonloch:
Screen Shot 2023-08-23 at 17 38 06
image

/claim #310


carlosvirreira:
Super nice, thank you. Would make the icon slightly less big and did you by any chance test this on mobile?

evonloch:
Hey! I can't push the code due to merge may be. I have tested it in browser's mobile device.

app/components/icons/library.tsx
```diff
@@ -875,3 +875,48 @@ export const ArrowLeftIcon = (props: SVGProps<SVGSVGElement>) => (
...
+    height=" 8%"
+    viewBox="0 0 24 24"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
+  >
+    <path
+      d="M2.42012 12.7132C2.28394 12.4975 2.21584 12.3897 2.17772 12.2234C2.14909 12.0985 2.14909 11.9015 2.17772 11.7766C2.21584 11.6103 2.28394 11.5025 2.42012 11.2868C3.54553 9.50484 6.8954 5 12.0004 5C17.1054 5 20.4553 9.50484 21.5807 11.2868C21.7169 11.5025 21.785 11.6103 21.8231 11.7766C21.8517 11.9015 21.8517 12.0985 21.8231 12.2234C21.785 12.3897 21.7169 12.4975 21.5807 12.7132C20.4553 14.4952 17.1054 19 12.0004 19C6.8954 19 3.54553 14.4952 2.42012 12.7132Z"
+      stroke="currentColor"
+      stroke-width="2"
```
DonKoko: This is not valid JSX. Please check your terminal and you will see the errors. EIther fix them manually or use something like this: https://react-svgr.com/playground/

app/routes/_auth+/login.tsx
```diff
@@ -116,18 +124,28 @@ export default function IndexLoginForm() {
...
             data-test-id="password"
             name={zo.fields.password()}
-            type="password"
+            type={showPassword ? "text" : "password"}
             autoComplete="new-password"
             disabled={disabled}
             inputClassName="w-full"
             error={zo.errors.password()?.message || data?.errors?.email}
           />
+          <span
```
DonKoko: Please set the width of this span to 20px and the color to `text-gray-500`. Also when you resize make sure its properly aligned vertically.

DonKoko (changes requested):
Its almost completely good, however we have a few more things.

1. This is not the only password field we have. We also have the password field in:
 - `onboarding.tsx`
 - `reset-password.tsx`
 
 2. Lets create a component for it `PasswordInput` and use that in all the different places. For the props of `PasswordInput` you can use the same props as `InputProps`. Looking at it I don't think it needs any special props by itself, it can just drill them all directly to `Input`.

app/routes/_auth+/login.tsx
```diff
@@ -97,6 +99,12 @@ export default function IndexLoginForm() {
   const navigation = useNavigation();
   const disabled = isFormProcessing(navigation.state);
 
+  const [showPassword, setShowPassword] = useState(false);
+
+  const togglePasswordVisibility = () => {
+    setShowPassword(!showPassword);
```
DonKoko: Can you please adjust this to: `setShowPassword((prevState) => !prevState)` Its safer and makes sure we are always using the prevState.

app/components/icons/library.tsx
evonloch: fixed

app/routes/_auth+/login.tsx
evonloch: done

evonloch:
Screenshot from 2023-08-24 15-36-23

evonloch:
Screenshot from 2023-08-24 15-09-37

evonloch:
Screenshot from 2023-08-24 15-10-13

DonKoko (approved):

Issue #149, opened 2023-06-06
User request: dedicated app


DonKoko:
Create an dedicated app that can be installed on windows or mac. 
We can do it using electron: https://github.com/lucasrabiec/electron-remixed

carlosvirreira:
This would be amazing!

rishi-raj-jain:
Just a starter electron app?

rishi-raj-jain:
Also, @neo773 how do you even get there so quickly!

neo773:
> Also, @neo773 how do you even get there so quickly!

I could ask you the same thing lol

DonKoko:
> Just a starter electron app?

I don't have experience with electron so I am not 100% sure what you mean. When I open an electron app I have, for example Discord I basically see the same as in the browser. 
What I assume is that you refer weather we need like a more powerful menu that integrates with the app features, something like vscode:
<img width="1338" alt="Screenshot 2023-08-23 at 17 45 09" src="https://github.com/Shelf-nu/shelf.nu/assets/7840007/1d21a1b6-c173-430e-a7bf-17af303a1f79">

Then the answer is no. Can you let me know if i understand you correctly?

PR #315, opened 2023-08-23
feat: add dedicated desktop App

neo773:
This PR adds a dedicated App for macOS & Windows 

https://github.com/Shelf-nu/shelf.nu/assets/62795688/bb0356f0-b887-4c73-897e-de1fb7ce09d9

/claim #149


DonKoko:
hey @neo773. Thanks for the PR. I have checked it and have done some testing locally on my laptop. We didn't expect you to use something like nativefier however we don't know enough about it to say weather its good or bad.
It does have some dependencies 

So here are some questions I have concerning this.

1. From what I can see this does nothing with our codebase. It just uses the url passed to create the Electron app. Does that mean that we don't have to make a new build when there is a new version of Shelf deployed? Will it always use the most up to date version of the live app?
2. How does security work? I generated the app and sent it to @carlosvirreira and @jurrejansen for their Mac and they get the standard security error from OSX:

image

So how can we solve this? We will make an account for Apple developers, but can we actually avoid this error for an app made with nativefier.

3. I couldn't generate the windows app on my Mac.  Here is the error I got:

```sh
Packaging... This will take a few seconds, maybe minutes if the requested Electron isn't cached yet...
Error during build. Run with --verbose for details. ExitSignalError: Command terminated via a signal (SIGKILL):
wine64 /Users/donkoko/devs/JS/shelf/node_modules/nativefier/node_modules/rcedit/bin/rcedit-x64.exe desktop-app/Shelf-win32-arm64/Shelf.exe --set-version-string FileDescription Shelf --set-version-string InternalName Shelf --set-version-string OriginalFilename Shelf.exe --set-version-string ProductName Shelf --set-version-string CompanyName Jia Hao --set-file-version 1.0.0 --set-product-version 1.0.0 --set-icon public/images/shelf-symbol-desktop.ico
    at ChildProcess.<anonymous> (/Users/donkoko/devs/JS/shelf/node_modules/nativefier/node_modules/@malept/cross-spawn-promise/src/index.ts:170:16)
    at ChildProcess.emit (node:events:513:28)
    at ChildProcess.emit (node:domain:489:12)
    at maybeClose (node:internal/child_process:1100:16)
    at Socket.<anonymous> (node:internal/child_process:458:11)
    at Socket.emit (node:events:513:28)
    at Socket.emit (node:domain:489:12)
    at Pipe.<anonymous> (node:net:301:12) {
  cmd: 'wine64',
  args: [
    '/Users/donkoko/devs/JS/shelf/node_modules/nativefier/node_modules/rcedit/bin/rcedit-x64.exe',
    'desktop-app/Shelf-win32-arm64/Shelf.exe',
    '--set-version-string',
    'FileDescription',
    'Shelf',
    '--set-version-string',
    'InternalName',
    'Shelf',
    '--set-version-string',
    'OriginalFilename',
    'Shelf.exe',
    '--set-version-string',
    'ProductName',
    'Shelf',
    '--set-version-string',
    'CompanyName',
    'Jia Hao',
    '--set-file-version',
    '1.0.0',
    '--set-product-version',
    '1.0.0',
    '--set-icon',
    'public/images/shelf-symbol-desktop.ico'
  ],
  stdout: '',
  stderr: '',
  signal: 'SIGKILL'
}
```

4. Can we generate actual installers for windows and mac. Our end users who will download this need to have it easy to 'install' it. For a lot of users the idea of having a portable app is not understandable. They don't know where to place it on windows, or that they need to move it to applications folder on mac.

5. Can we get the generated output inside a zip file, then it will be easier for our users to download it all.

I am doing some more testing now and trying to figure out the practicality of this and how we can ship it to our users.

neo773:
@DonKoko 
To address your questions
- Nativfier uses Electron under the hood
1. Yes, you don't have to make a new build it will always show the latest version/
2. That's just standard OS X Code Signing check even if you don't use nativifier you'll still get that Warning. Only way to dismiss that Warning is to sign the App with an Apple Developer account.

carlosvirreira:
@neo773 feel free to join us here: 

https://discord.gg/nJYKvpcf

DonKoko:
Okey so after our discussions on discord and some testing we managed to make both installers work. Thanks for the work @neo773. However we still have some stuff that need to be fixed:

1. Currently building both installers is in the same command. Lets separate them because like this the command will fail when used on both systems as each system cannot build for the other
2. We have a package called `"appdmg": "^0.6.6",`. Installing with `npm` fails on windows because of this package. I am not sure if you can somehow scope this package only to one OS but i know package.json supports `optionalDependencies`. Maybe thats a good approach?
3. The windows installation app has a few things we need to improve.
So here is how the folder looks after the build:
image

 - Lets rename `dist` to `portable` and `binaries` to `shelf-installer`.
 - inside binaries I see the following:
 
image

So this was a bit confusing for me. Which one do I have to use? Why are there 2 files called setup.  The correct file to install it with is the second one which is of type `Application`. So can we name this file something like `Install shelf`. Also it doesn't ask where to install it. Can we adjust that? When I check its installed in:
image
I would say even if we can't change the location, it should install in a folder called Shelf.

Last but not least, if I go to uninstall it, it doesn't have an icon. It does have one when starting the app tho.
image

I think this is everything that I found.

desktop-app/build.js
```diff
@@ -0,0 +1,93 @@
+const createDMG = require("electron-installer-dmg");
+const { signAsync } = require("@electron/osx-sign");
+const electronInstaller = require("electron-winstaller");
```
DonKoko: Here there are some esLint errors because of the order of imports. Can you fix them please. You can run `npm run lint` to see them.
Also can you please change the `.eslintignore` to not include the built apps. 
You can also add `/* eslint-disable no-console */` for the whole file `build.js`

desktop-app/build.js
DonKoko, line 1: ## For Mac

So when I run npm run build:mac i get an error for signing(that makes sense). Can you explain me what do i need to do to sign it once I have the developer ID?
also I got this output, but i remember you managed to make a dmg file. Not sure if thast also failing because of the signing but I dont get that file here.
image



## For windows

I get this in the installer. I can rename the file thats not an issue but it shows this placeholder description. Can we change it to `Install Shelf`.
image

DonKoko (changes requested):
When running `npm install` there are some warnings that can be solved easily. Can you please rename those packages.
image

Everything except the `start-server-and-test@1.15.5:` should be easily fixable. I also got some error, not sure why it happens maybe some permissions stuff related to how the `optionalDependencies` works.

Also I tried again the build in windows. The description worked but the icon didn't.
image

DonKoko (approved):

PR #321, opened 2023-08-28
custom fields

hunar1312:


app/components/assets/form.tsx
```diff
@@ -53,6 +54,21 @@ export const AssetForm = ({
   const [, validateFile] = useAtom(validateFileAtom);
   const [, updateTitle] = useAtom(updateTitleAtom);
 
+  const fields = [
```
DonKoko: We should not forget to clean this up. Just leaving a note.

app/components/assets/form.tsx
```diff
@@ -163,7 +180,51 @@ export const AssetForm = ({
...
+              key={field.id}
+              rowLabel={field.name}
+              subHeading={field.helpText && <p>{field.helpText}</p>}
+              className="border-b-0"
+            >
+              <Input
+                hideLabel
+                type={field.type}
+                label={field.name}
+                name={field.name}
```
DonKoko: We are missing the required. Also a bit of a challenge. We have some schema for creating/editing assets form data. We need to see how we can dynamically add to the schema based on the custom fields config.

app/components/assets/form.tsx
```diff
@@ -163,7 +180,51 @@ export const AssetForm = ({
...
+                className="w-full"
+              />
+            </FormRow>
+          ))
+        ) : (
+          <div>
```
DonKoko: Lets move this to its own component so we can keep the file cleaner. Its already quite big.

app/components/custom-fields/form.tsx
```diff
@@ -0,0 +1,132 @@
...
+interface Props {
+  title?: string;
+  helpText?: string;
+}
+
+export const CustomFieldForm = ({ title, helpText }: Props) => {
+  const navigation = useNavigation();
+  const zo = useZorm("NewQuestionWizardScreen", NewCustomFieldFormSchema);
+  const disabled = isFormProcessing(navigation.state);
+  const fieldTypes = ["Text"];
```
DonKoko: This will need to be cleaned up.

hunar1312: yes, this all static content will be cleaned up. This was added just to make frontend work and backend integration easier

app/components/assets/form.tsx
hunar1312: sure

app/components/custom-fields/form.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const [, updateTitle] = useAtom(updateTitleAtom);
+
+  const organizationId = useOrganizationId();
+  return (
+    <Form
```
DonKoko, line 63: Submitting with cmd+enter doesn't seem to work and it works on example on asset.new. Can you please check why?

app/routes/_layout+/settings.custom-fields.index.tsx
```diff
@@ -0,0 +1,119 @@
...
+    totalPages,
+    perPage,
+    prev,
+    next,
+    modelName,
+  });
+}
+
+export const action = async ({ request }: ActionArgs) => {
+  await requireAuthSession(request);
```
DonKoko, line 74: Here we only have a delete action so I would use the `isDelete()` helper to make sure the request has the correct method.

DonKoko (changes requested):
Overall really good. We discussed some stuff on discord and I also added some comments here. What is also missing the edit view.

DonKoko (changes requested):
Clicking on the actions dropdown of the custom field, navigates to the field page. Not sure why it happens

PR #327, opened 2023-08-29
feat: add workflow with permission

### LINKED ISSUE #319: [Feature request]: Special Action workflow for PRs
### Contact Details

_No response_

Currently when someone opens a PR to main from a fork, the e2e tests fail because of the required env variables. Here is the only context I could find: https://stackoverflow.com/questions/75873833/how-to-protect-github-secrets-in-pull-request-actions-from-malicious-pull-reques

So ideally, the e2e tests can run, however I am not sure if that's possible. We have the required secrets in the shelf repository but from what I read those are not passed onto PRs running the tests.

Someone mentioned to me that we can do a workflow that runs instead of the default one when a PR is opened from a fork. Then we can run all the steps from our deploy workflow except the e2e tests. 

In any case this needs to be looked into and any suggestions how to solve this are welcome.

_No response_

### Additional context

_No response_

kunal00000:
@DonKoko please [create an environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#creating-an-environment) named `external` and add some trusted users there.

Note: I have'nt removed original playwright job from the Deploy workflow. Once this one works I'll make pull request for that

Resolves #319
/claim #319


.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
+name: 🧪 Test
```
DonKoko: Can you name it something like: `Test PR from fork`
I rather we are more explicit so its clear for new comers what this is about

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+
+on: 
+  push:
+    branches:
+      - main
+      - dev
+  pull_request_target:
```
DonKoko, line 8: I am curious how this works? Shouldn't there be some value there?

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+    environment: 
+      ${{ github.event_name == 'pull_request_target' &&
+      github.event.pull_request.head.repo.full_name != github.repository &&
+      'external' || 'internal' }}
+
+    runs-on: ubuntu-latest
+    steps:
+    - run: echo ✓
+  
+  playwright:
```
DonKoko, line 97: This is good but we should add the other tests as well (lint, ts and vitest).

.github/workflows/test.yml
```diff
@@ -0,0 +1,68 @@
...
+    needs: authorize
+    name: 🎭 Playwright
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+      - name: 🛑 Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
```
DonKoko, line 104: Can we check if this is scoped to the workflow. I just want to make sure it doesn't cancel a deploy workflow that is running as this workflow starts.

DonKoko:
hey @kunal00000 . A bit late to review it because I had to handle some issues with other projects that took almost the whole week. 
So i checked the workflow you created and added a review with some small comments.

Also I think the playwright shouldn't be removed from the deploy as I want it to run also on deploy. Even if the tests passed in the PR.

.github/workflows/test.yml
kunal00000: Okay.

.github/workflows/test.yml
kunal00000: yes, that seems good.

.github/workflows/test.yml
kunal00000: As of my knowlegde, If no value is given to `pull_request_target:`, by default the workflow will be triggered by all events(open, sync, reopen) related to fork pull requests targeting the base repository.
Although we can give types and  branches filter but we won't be needing those.

.github/workflows/test.yml
kunal00000: It is scoped to a particular workflow according to https://github.com/marketplace/actions/cancel-workflow-action and there's particular configuration to stop other workflows.

kunal00000:
hi @DonKoko , replied to reviews and pushed the changes accordingly. 

As far as this is considered,
>  playwright shouldn't be removed from the deploy

it would fail everytime as secrets will not be passed until `pull_request_target` is used as trigger.
What we can do is use [workflow_call](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call) to trigger Test workflow whenever Deploy is triggered.

kunal00000:
Hello @DonKoko, any updates here?

DonKoko (approved):

DonKoko:
> Hello @DonKoko, please review the changes.

hey @kunal00000 . Changes look good and are approved. Only thing I still would like to solve is your suggestions from here:

> As far as this is considered,

> > playwright shouldn't be removed from the deploy

> it would fail everytime as secrets will not be passed until pull_request_target is used as trigger.
> What we can do is use [workflow_call](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call) to trigger Test workflow whenever Deploy is triggered.

I like that approach. Can you adjust that as well.

DonKoko (commented):
Okey this looks good to me. I think at this point all we have to do is test it. It's hard to say anymore if it will work by just looking at the code.
@kunal00000 how would you suggest we do it? I can merge it but shall we make a testing plan?

kunal00000:
hello @DonKoko . made the changes. 
I am worried about some things that may cause failure in this job:
- Test jobs called by Deploy workflow may or may not get access to secrets.
- and some more which I am not sure of.

We need to test this by merging and making test pull requests.
Note: Before merging please be sure to create environment named `external`.

kunal00000:
I think you should merge it and then I can make a pull request to remove jobs from Deploy workflow or some readme changes just to test the workflow.

kunal00000:
@DonKoko . Let's test this now.

DonKoko:
@kunal00000 its merged and this is what I see in the action:
image

PR #334, opened 2023-09-06
UI updates in custom fields

hunar1312:


app/components/custom-fields/form.tsx
```diff
@@ -48,6 +49,11 @@ export const CustomFieldForm = ({ name, helpText, required, type }: Props) => {
   const fieldTypes = CustomFieldType;
 
   const [, updateTitle] = useAtom(updateTitleAtom);
+  const fieldTypeDefaultValue = "TEXT";
+
+  const [hintTextVisibility, setHintTextVisibility] = useState(
+    fieldTypeDefaultValue === "TEXT" || type === "TEXT" ? true : false
```
DonKoko: Expressions already evaluate to true or false. you dont need to set them specifically.

app/components/custom-fields/form.tsx
```diff
@@ -49,6 +50,12 @@ export const CustomFieldForm = ({ name, helpText, required, type }: Props) => {
 
   const [, updateTitle] = useAtom(updateTitleAtom);
 
+  // keeping text field type by default selected
+  const [selectedFieldType, setSelectedFieldType] = useState("TEXT");
+  const fieldTypeDescription = {
```
DonKoko: This works however I want to change it a bit. Its a constant, it will never change so there is 2 things:
1. We should define it outside the component
2. For constants I want to start using the convention of naming them in all caps. So it should be `FIELD_TYPE_DESCRIPTION`

DonKoko (changes requested):

PR #335, opened 2023-09-06
Integration with Stripe for premium features + some new features

DonKoko:
New features include:
- Exporting assets
- Importing assets from template
- Importing assets from backup
- Custom fields


carlosvirreira:
Some feedback: 

- **Custom fields** | We have a toggle for required / not required on custom field create. However, when creating a new asset, the user gets no hint of what was a required field or not. Only on submit. If we could use an asterisk for required fields this would be great. 
- New asset button on asset index should be our orange primary button
- Deleting a field that has been used by an asset returns the following error (not required) ```Invalid `db.customField.delete()` invocation in /Users/donkoko/devs/shelf/app/app/routes/_layout+/settings.custom-fields.index.tsx:80:26 77 const formData = await request.formData(); 78 const customFieldId = formData.get("customFieldId") as string; 79 → 80 await db.customField.delete( Foreign key constraint failed on the field: `AssetCustomFieldValue_customFieldId_fkey (index)`
```
- Deleting a field that has been used baby an asset (required) ```Invalid `db.customField.delete()` invocation in /Users/donkoko/devs/shelf/app/app/routes/_layout+/settings.custom-fields.index.tsx:80:26 77 const formData = await request.formData(); 78 const customFieldId = formData.get("customFieldId") as string; 79 → 80 await db.customField.delete( Foreign key constraint failed on the field: `AssetCustomFieldValue_customFieldId_fkey (index)`
```
- On free tier I can add 3 custom fields, however, I do not currently have a limit
- (Improvement) > on the pricing table we have on free tier Github support - it would be nice if we can directly link to Github (https://github.com/Shelf-nu/shelf.nu)
- Trying to upgrade to Plus (monthly) returns this issue > ```No such customer: 'cus_OPJoKvtx808D7o'```
-

DonKoko:
thanks @carlosvirreira for the feedback. Here I am adding some extra points:
 - [ ] Make text on subscription success modal dynamic to which package they upgraded to
 - [ ] Trying to export when there are no assets, breaks the app
 - [ ] Add custom fields to import/export

jurrejansen:
@DonKoko

- [ ] When custom field is set to inactive, it should not show on asset pages even if it is filled

- [ ] When clicking 'Export' using Chrome browser it bricks Chrome

Uploading Screen Recording 2023-09-13 at 15.08.30.mov…

PR #338, opened 2023-09-07
custom fields activation/deactivation

hunar1312:
disabled deletion of custom fields and added the concept of active/inactive custom field


app/components/assets/custom-fields-inputs.tsx
```diff
@@ -35,29 +35,35 @@ export default function AssetCustomFields({
...
-                customFieldsValues?.find(
-                  (cfv) => cfv.customFieldId === field.id
-                )?.value || ""
-              }
-              className="w-full"
-            />
-          </FormRow>
-        ))
+        customFields.map((field: CustomField) => {
+          if (field.active) {
```
DonKoko: Yeah this is not good. We should change the query in `getAllRelatedEntries` to only get active ones.

app/components/custom-fields/form.tsx
```diff
@@ -31,6 +31,10 @@ export const NewCustomFieldFormSchema = z.object({
     .string()
     .optional()
     .transform((val) => (val === "on" ? true : false)),
+  active: z
+    .string()
+    .optional()
```
DonKoko, line 36: Shouldn't be optional, right?

app/components/custom-fields/form.tsx
```diff
@@ -131,6 +144,27 @@ export const CustomFieldForm = ({ name, helpText, required, type }: Props) => {
...
       </FormRow>
 
+      <FormRow rowLabel="" className="border-b-0 pt-2">
+        <div className="flex items-center gap-3">
+          <Switch
+            name={zo.fields.active()}
+            disabled={disabled}
+            defaultChecked={true}
+            checked={isActive}
+            onCheckedChange={(checked) => setIsActive(checked)}
```
DonKoko: I don't understand why we control this field? The other switch is uncontrolled and this is our preferred approach.

app/routes/_layout+/settings.custom-fields.index.tsx
```diff
@@ -79,7 +78,8 @@ export async function loader({ request }: LoaderArgs) {
     }),
   });
 }
-
+/*
+* disabling delete functionality for now
```
DonKoko: You can just remove this code. I don't see us adding this soon again.

DonKoko (changes requested):

DonKoko (approved):

PR #343, opened 2023-09-08
do not merge

kunal00000:


DonKoko:
@kunal00000 this looks better, its just showing the authorize waiting to be done. Shall I approve so we see what runs?

kunal00000:
Yes

kunal00000:
Finally things are working as expected 😌

DonKoko:
@kunal00000 I think we have success. I just want to see which error the playwright fails with. If its the SERVER_URL error that means I have to add the vars to that environment in github, if its the other error, thats just the test being broken.

kunal00000:
> @kunal00000 I think we have success. I just want to see which error the playwright fails with. If its the SERVER_URL error that means I have to add the vars to that environment in github, if its the other error, thats just the test being broken.

Yes we need to confirm this.

kunal00000:
It worked fine.
It was really interesting issue to work on, learnt alot of things on the go. 
Thanks @DonKoko for help through the process.

PR #363, opened 2023-09-15
355 store per page settings in user prefs cookie

### LINKED ISSUE #355: [Feature request]: Store per_page settings for indecies in user prefs cookie
### Contact Details

_No response_

Currently when you change the per_page setting on an index, and then refresh it resets. This is because we dont store it a anywhere.

Store it per index in the `userPrefs` cookie we have setup, the same way we store for example weather the sidebar is opened or folded. You can check the routes for examples:
- user.minimized-sidebar.tsx
- user.dismiss-support-banner.tsx

Lets make a few things:
 - we are starting to get more of those endpoints so its nice to re-organize them a bit. Currently the route structure is `api/user/pref-name.ts`. I would like to add in the middle a route to make it clear those are the user preferences routes and they are different than for example `user.upload-user-photo.tsx`. So lets put them in a `prefs` parent route. So I think we can do `user.preferences.any-route-we-need.ts`
 - Lets create a central and generic way to store this data in a cookie per index. So you should be able to configure it differently for assets index and categories index.

_No response_

### Additional context

_No response_

hunar1312:
resolves #355


app/modules/asset/service.server.ts
```diff
@@ -593,16 +596,16 @@ export const getPaginatedAndFilterableAssets = async ({
   const { assets, totalAssets } = await getAssets({
     userId,
     page,
-    perPage,
+    perPage: cookie.perPage,
```
DonKoko: Let's make a perPage variable and use it in all 3 places. You can just destructure it from `cookie`

app/modules/asset/service.server.ts
```diff
@@ -593,16 +596,16 @@ export const getPaginatedAndFilterableAssets = async ({
...
     search,
     categoriesIds,
     tagsIds,
   });
-  const totalPages = Math.ceil(totalAssets / perPage);
+  const totalPages = Math.ceil(totalAssets / perPageParam);
```
DonKoko: Lets use the perPage var here.

app/modules/organization/service.server.ts
```diff
@@ -67,26 +68,29 @@ export const getPaginatedAndFilterableTeamMembers = async ({
...
+  const { page, perPageParam, search } = getParamsValues(searchParams);
   const { prev, next } = generatePageMeta(request);
 
+  const cookie = await updateCookieWithPerPage(request, perPageParam);
+
   const { teamMembers, totalTeamMembers } = await getTeamMembers({
     organizationId,
     page,
-    perPage,
+    perPage: cookie.perPage,
```
DonKoko: Same as in assets.

DonKoko (changes requested):
Small changes required for code consistency.

DonKoko (approved):

DonKoko (approved):

Issue #309, opened 2023-08-22
[Feature request]: Scan codes from inside the website/web application


carlosvirreira:
### Contact Details

carlos@shelf.nu

### Is your feature request related to a problem? Please describe?

When I added shelf to my Home Screen as an 'web app' (https://support.apple.com/en-gb/guide/iphone/iph42ab2f3a7/ios) 

I now have to open my camera app all the time to start scanning codes. 

I would like to do this in the mobile version of the website.

This issue is in progress, a bounty will be placed for it later.

### Describe the solution you'd like

<img width="495" alt="Screenshot 2023-08-22 at 13 28 07" src="https://github.com/Shelf-nu/shelf.nu/assets/12449255/73391226-8949-45aa-ab14-fc07aa9ce8e0">

jurrejansen:
@carlosvirreira let's change this placement of button to the top navigation so it is always visible.

We'll have to discuss what the button it actually going to do
- Open default Camera App
- Design/Develop MVP scanner

carlosvirreira:
Let's go for > Open default Camera App

jurrejansen:
It would be great if it could work like this:

OPTION 1

- Tap 'Scan' icon
- Opens Default Camera app within the PWA
- Scan code
- Redirects to asset page

Instead of:

OPTION 2

- Tap 'Scan' icon
- Opens Default Camera App
- switches from browser app to Default Camera app
- Scan code
- Switches to browser app and opens new tab with asset page

With the first option having a kind of UX like this, but with scanning capability

https://github.com/Shelf-nu/shelf.nu/assets/76112272/cd80f75d-f4a9-41f6-a25e-de930188e31d

But if we want to keep it simple, let it be a fast improvement and open a bounty for it, the second option may be better

@DonKoko thoughts?

DonKoko:
I dont know for sure but I am pretty certain that option 1 is not possible.

carlosvirreira:
Example website opening camera in - website


https://github.com/Shelf-nu/shelf.nu/assets/12449255/b31d4432-2f27-46e0-9307-d17ff94661db

carlosvirreira:
@ap172x I see you are attempting this. Perhaps it is handy for you to join us on Discord? 

https://discord.gg/4ypbGTRJ

carlosvirreira:
hey @ap172x how is it going with this issue? have you made some progress?

Hitesh-sisara:
should i give it try ? 
[...]

PR #364, opened 2023-09-15
feature : Scan codes from inside the website/web application #309

Hitesh-sisara:
added a new feature : Scan codes from inside the website

when user presses QR code button, QR code scanner will be displayed using the device's native camera if the qr code contains valid asset links, the user will be redirected to the assets page 

https://github.com/Shelf-nu/shelf.nu/assets/72780174/93076566-bf65-4f36-9666-af0b0f9e70b5

### DIFF
--- a/app/atoms/notifications.ts
+++ b/app/atoms/notifications.ts
@@ -3,16 +3,17 @@ import { atom } from "jotai";
 import type { Icon } from "~/components/shared/icons-map";
 
 export type NotificationVariant = "primary" | "gray" | "success" | "error";
+export type NotificationIcon = {
+  name: Icon;
+  variant: NotificationVariant;
+  className?: string;
+};
 
 export interface NotificationType {
   open: boolean;
   title: string;
   message: string;
-  icon: {
-    name: Icon;
-    variant: NotificationVariant;
-    className?: string;
-  };
+  icon: NotificationIcon;
   time?: number;
   senderId: User["id"] | null;
 }
--- a/app/hooks/index.ts
+++ b/app/hooks/index.ts
@@ -4,3 +4,4 @@ export * from "./use-current-route-data";
 export * from "./use-user-data";
 export * from "./use-focus-search";
 export * from "./use-position";
+export * from "./use-client-notification";
--- a/app/hooks/use-client-notification.ts
+++ b/app/hooks/use-client-notification.ts
@@ -1,0 +1,25 @@
+import { useFetcher } from "@remix-run/react";
+import type { NotificationType } from "~/atoms/notifications";
+
+export function useClientNotification() {
+  const fetcher = useFetcher();
+  const submitter = ({
+    title,
+    message,
+    icon,
+  }: Omit<NotificationType, "open" | "senderId">) => {
+    fetcher.submit(
+      {
+        title,
+        message,
+        icon,
+      },
+      {
+        action: "/api/client-notification",
+        method: "POST",
+      }
+    );
+  };
+
+  return [submitter];
+}
--- a/app/routes/_layout+/_layout.tsx
+++ b/app/routes/_layout+/_layout.tsx
@@ -20,6 +20,7 @@ import type { CustomerWithSubscriptions } from "~/utils/stripe.server";
 import {
   getCustomerActiveSubscription,
   getStripeCustomer,
+  stripe,
 } from "~/utils/stripe.server";
 
 export const links: LinksFunction = () => [{ rel: "stylesheet", href: styles }];
@@ -45,7 +46,7 @@ export const loader: LoaderFunction = async ({ request }: LoaderArgs) => {
       })
     : undefined;
   let subscription = null;
-  if (user?.customerId) {
+  if (user?.customerId && stripe) {
     // Get the Stripe customer
     const customer = (await getStripeCustomer(
       user.customerId
--- a/app/routes/_layout+/assets._index.tsx
+++ b/app/routes/_layout+/assets._index.tsx
@@ -171,6 +171,8 @@ export default function AssetIndexPage() {
     clearTagFilters();
   };
 
+  // const [sendNotification] = useClientNotification();
+
   return (
     <>
       <Header>
--- a/app/routes/api+/client-notification.ts
+++ b/app/routes/api+/client-notification.ts
@@ -1,0 +1,35 @@
+import { json, type LoaderArgs } from "@remix-run/node";
+import { parseFormAny } from "react-zorm";
+import { z } from "zod";
+import type { NotificationIcon } from "~/atoms/notifications";
+import { requireAuthSession } from "~/modules/auth";
+import { sendNotification } from "~/utils/emitter/send-notification.server";
+
+export const ClientNotificationSchema = z.object({
+  title: z.string().min(4, { message: "Title is required" }),
+  message: z.string().min(10, { message: "Message is required" }),
+  icon: z.custom<NotificationIcon>(),
+});
+
+export async function action({ request }: LoaderArgs) {
+  const authSession = await requireAuthSession(request);
+
+  const formData = await request.formData();
+  const result = await ClientNotificationSchema.safeParseAsync(
+    parseFormAny(formData)
+  );
+
+  if (!result.success) {
+    return json({ error: result.error.message }, { status: 400 });
+  }
+  const { title, message, icon } = result.data;
+
+  sendNotification({
+    title,
+    message,
+    icon,
+    senderId: authSession.userId,
+  });
+
+  return json({ success: true });
+}


app/components/layout/sidebar/menu-button.tsx
```diff
@@ -2,11 +2,10 @@ import { useAtom } from "jotai";
 import { tw } from "~/utils";
 import { toggleMobileNavAtom } from "./atoms";
 
-
 const MenuButton = () => {
   const [isMobileNavOpen, toggleMobileNav] = useAtom(toggleMobileNavAtom);
   return (
-    <button className="menu-btn relative z-50" onClick={toggleMobileNav}>
+    <button className="menu-btn relative z-50 pt-1.5" onClick={toggleMobileNav}>
```
DonKoko: Is this so the button is perfectly aligned with the QR button as well?

package.json
DonKoko, line 1: You updated some stuff here that shouldn't be updated such as remix revision and prisma. Please reverse those

package.json
```diff
@@ -100,13 +102,14 @@
     "eslint": "^8.44.0",
     "eslint-config-prettier": "^8.8.0",
     "eslint-plugin-tailwindcss": "^3.13.0",
+    "eslint-plugin-unused-imports": "^3.0.0",
```
DonKoko: Not sure what this is.

package.json
```diff
@@ -50,9 +50,11 @@
     "@stripe/react-stripe-js": "^2.1.1",
     "@stripe/stripe-js": "^1.54.2",
     "@supabase/supabase-js": "^2.26.0",
+    "@zxing/library": "^0.20.0",
```
DonKoko: Can you please write on the PR the reasoning behind the decision to use this?

yarn.lock
DonKoko, line 1: We use npm in this project by default. If you prefer to use yarn thats fine but lets remove this file from the commits.

app/components/scan-qr/noop.ts
DonKoko, line 1: Can you please add some comments to this file? What's `noop`? Also I can see its inside the components folder and it doesn't seem like a component to me.

app/components/scan-qr/use_boolean_state.tsx
DonKoko, line 1: Please change file name to follow our convention: `use-boolean-state.tsx`. Also this seems like a generic hook that we could re-use in different places. We have a folder where we place our hooks.

app/components/layout/sidebar/sidebar.tsx
```diff
@@ -11,12 +11,56 @@ import SidebarBottom from "./bottom";
...
   const [isMobileNavOpen, toggleMobileNav] = useAtom(toggleMobileNavAtom);
   const mainNavigationRef = useRef<HTMLElement>(null);
 
+  const [showScanner, setShowScanner] = useState(false);
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [scannedSuccessfully, setScannedSuccessfully] = useState(false);
+
+  const { startMediaStream, stopMediaStream } = useMediaStream();
+
+  const isValidQRCode = (data: string) => {
```
DonKoko: Can you explain this? What's this regex validating?

app/components/scan-qr/QRCodeScanner.tsx
DonKoko, line 1: Please follow our file naming conventions. `qr-code-scanner.tsx` is the correct format.

app/components/scan-qr/media_stream_provider.tsx
DonKoko, line 1: Please follow our file naming conventions. Also what we are exporting here is basically a hook. Lets name the file accordingly and place it in our hooks folder.

app/components/layout/sidebar/sidebar.tsx
```diff
@@ -11,12 +11,56 @@ import SidebarBottom from "./bottom";
...
+    const regex = /^(https?:\/\/)([^/:]+)(:\d+)?\/qr\/[a-zA-Z0-9]+$/;
+    return regex.test(data);
+  };
+
+  const handleQRScan = (data: string | null) => {
+    if (data) {
+      if (isValidQRCode(data)) {
+        setScannedSuccessfully(true);
```
DonKoko: On a successful scan, you need to navigate the user to the route.
To do that you can do the following:
```ts
const navigate = useNavigate()
navigate(`/qr/${qr.id}`)
```

app/components/layout/sidebar/sidebar.tsx
```diff
@@ -42,8 +86,51 @@ export default function Sidebar() {
...
+            onScan={(data) => handleQRScan(data)}
+            onClose={() => setShowScanner(false)}
+          />
+          {/* Add a close button on top right corner of the scanner */}
+          <button
+            className="absolute right-2 top-2 flex h-8 w-8 items-center justify-center rounded-full bg-red-500 text-white"
+            onClick={handleScannerClose}
+            title="Close Scanner"
+          >
+            &times; {/* This is an "X" character */}
```
DonKoko: This works good but we already have an icon we use for this. It's called `XIcon`. You can check the `dialog` component to see how we use it.

app/components/layout/sidebar/sidebar.tsx
```diff
@@ -42,8 +86,51 @@ export default function Sidebar() {
...
+          </button>
+        </div>
+      )}
+
+      {scannedSuccessfully && (
```
DonKoko: This I think can be removed as we will trigger a navigate so we don't need to do this.

app/components/layout/sidebar/sidebar.tsx
```diff
@@ -42,8 +86,51 @@ export default function Sidebar() {
...
+        <div className="my-4 flex items-center justify-center rounded border border-green-300 bg-green-100 p-4 text-green-700">
+          Assets scanned successfully...
+        </div>
+      )}
+
+      {errorMessage && (
```
DonKoko: This is fine but I would like to hear a bit more about it. What type of errors are we handling here? Can you show some examples of how the error will look?

DonKoko (commented):
Some changes requested as well as some questions. Overall the code looks good and professionally written, we just need to re-organize everything a bit and then I can spend some more time checking the details of the code.

app/components/layout/sidebar/menu-button.tsx
Hitesh-Sisara: to align both button in center of header this was needed

Hitesh-sisara: i will revert those

package.json
Hitesh-Sisara: I think it must have been updated because I had tried lot of plugins to get that specific qr scanning  functionality

app/components/layout/sidebar/sidebar.tsx
Hitesh-Sisara: to check if qr code URLs are valid, I had designed it to work with Shelf.nu URLs, local hots URLs, and any other self-hosted instances's URLs -- basically it verifies that scanned QR has valid asset URLs

app/components/scan-qr/noop.ts
Hitesh-Sisara: The purpose of defining noop as a VoidOrPromiseFunction is to have a function that can be used as a placeholder or as a default function when you need to provide a function with the specified signature (i.e., a function that returns void or a promise that resolves to void). This is often used in scenarios where you want to provide a safe, no-op (no operation) function as a default callback to avoid errors or handle cases where an actual callback is not provided.

app/components/layout/sidebar/sidebar.tsx
Hitesh-Sisara: this will show error if qr code data validation fails

package.json
Hitesh-Sisara: i have tested 3 other packges some don't work in safari  and some don't work at all for our use case finally I settled with @zxing/library for qr scanning functionality and used html5 based native camera access , used detectrtc package to simplify camera connection and stream

package.json
DonKoko: @Hitesh-sisara please ignore this comment. I merged a PR that had some eslint changes I missed. Please sync with main and you shouldn't have to use that package anymore.

package.json
DonKoko: Thats good @Hitesh-sisara. I just wanted to know a bit more. This is the best library out there so perfect choice.

app/components/scan-qr/noop.ts
DonKoko: Clear. Thanks for explaining. I never used it before so its always nice to learn something.

package.json
DonKoko: Yeah happens. Can you reverse it to its previous versions. We need to update it at some point but I want to do it as a separate scope as I want to do more testing and double check some stuff.

app/components/layout/sidebar/sidebar.tsx
DonKoko: Okey. Can you abstract it into a function with a very clear name, so we know what it does. Like `checkQrLinkValidity` or sth like this. 
Also on successful scan we would need to navigate to that url where we do some server side validations, but good to know beforehand if its even valid. Just to confirm, the correct structure is: `/qr/${qrId}`

app/components/layout/sidebar/sidebar.tsx
Hitesh-Sisara: Uploading Screen Recording 2023-09-18 at 6.01.22 PM.mov…

PR #368, opened 2023-09-19
Feature: QR Scan

Hitesh-sisara:
added a new feature: Scan codes from inside the website

when the user presses the QR code button, a QR code scanner will be displayed using the device's native camera if the qr code contains valid asset links, the user will be redirected to the assets page

used @zxing/library and detector package to achieve functionality 

/claim #309

This PR is update over my original pr 

https://github.com/Shelf-nu/shelf.nu/assets/72780174/267a6433-6213-4878-97e5-0e8fa3c18175


app/components/layout/sidebar/sidebar.tsx
```diff
@@ -17,6 +24,54 @@ export default function Sidebar() {
...
+
+      if (match) {
+        // const qrId = match[4]; // Get the last segment of the URL as the QR id
+
+        setScannedSuccessfully(true);
+
+        setShowScanner(false);
+        window.location.href = data;
+
+        // navigate(`/qr/${qrId}`); (using this way will not close camera access)
```
DonKoko: I don't understand this. Can't you just close it yourself?

app/components/scan-qr/media-stream-provider.tsx
```diff
@@ -0,0 +1,330 @@
...
+        });
+      }
+
+      if (constraints?.aspectRatio) {
+        advanced.push({
+          aspectRatio: constraints.aspectRatio,
+        });
+      }
+
+      alert(JSON.stringify(constraints));
```
DonKoko: This shouldnt be here right? We have the sendNotification thing already.

app/components/scan-qr/media-stream-provider.tsx
```diff
@@ -0,0 +1,330 @@
...
+      setCurrentVideoDeviceId("");
+
+      alert("Stop Media Stream");
+    } catch (error: any) {
+      sendNotification({
+        title: "camera acces error",
+        message: error.toString(),
+        icon: { name: "trash", variant: "error" },
+      });
+      alert(error.toString());
```
DonKoko: No alert. We are sending notification.

DonKoko (changes requested):
Overall its good but there is 1 thing I really don't like, and that is the code for the scanner being in the `<Sidebar/>` component. I just dont think it fits there. Any specific reason to not have a dedicated component for it?

app/components/layout/sidebar/sidebar.tsx
Hitesh-Sisara: then we need to display QR scanner  stand-alone page  when that particular page unmounts camra acces stops, i have tried many way to stop it self but i think it's limitation from package itself

app/components/scan-qr/media-stream-provider.tsx
Hitesh-Sisara: added new commit

app/components/scan-qr/media-stream-provider.tsx
Hitesh-Sisara: added new commit

Hitesh-sisara:
> Overall its good but there is 1 thing I really don't like, and that is the code for the scanner being in the `<Sidebar/>` component. I just dont think it fits there. Any specific reason to not have a dedicated component for it?

added new commit

app/components/layout/sidebar/sidebar.tsx
```diff
@@ -17,6 +22,17 @@ export default function Sidebar() {
...
 
+  const [showScanner, setShowScanner] = useState(false);
+
+  const { stopMediaStream } = useMediaStream();
+  useStopMediaStream(stopMediaStream);
+
+  const handleScannerClose = () => {
+    stopMediaStream();
+    setShowScanner(false);
+    window.location.reload(); // Clear error messages when closing the scanner
```
DonKoko: I don't understand why we have to reload the page. Can you clarify what is going on here?

app/components/layout/sidebar/sidebar.tsx
```diff
@@ -17,6 +22,17 @@ export default function Sidebar() {
   const [isMobileNavOpen, toggleMobileNav] = useAtom(toggleMobileNavAtom);
   const mainNavigationRef = useRef<HTMLElement>(null);
 
+  const [showScanner, setShowScanner] = useState(false);
```
DonKoko, line 25: I still don't like this code being in the sidebar component. I want to move all related code in its own component, not just the icon. You can just use an atom for the state.

DonKoko (changes requested):
Requested some changes. Also when i run eslint it gives some warnings. Can you fix them please.

app/components/layout/sidebar/sidebar.tsx
Hitesh-Sisara: if we don't reload then it's also fine, I have added that to make sure camera access get's terminated properly because on some browser packages don't support some features

Issue #370, opened 2023-09-19
[Feature request]: Signup/login with email/password


DonKoko:
### Is your feature request related to a problem? Please describe?

We currently only have a magic link sign-up. We used to have a email/password but we removed it as it was giving us some issues at that point of time.

### Describe the solution you'd like

1. Implement username/password signup. We actually had it but we removed it at some point. We used the `supa-fly-stack`. The idea is to check the code there and copy it so we dont have to do it from scratch. Here is the repo and the route: https://github.com/rphlmr/supa-fly-stack/blob/main/app/routes/join.tsx
2. In supa-fly-stack the confirmation email is set to true. In the case of registering with email/password supabase doesn't seem to support automatically sending confirmation email(please correct me if I am wrong), we we need to send it ourselves and build the link for confirmation based on the requirements of supabase. We already use nodemailer which is setup with our smtp server so that can be used to send the email. Check function `sendEmail`

DonKoko:
Nice to see you making an attempt. If you have questions and want to discuss please feel free to join our discord: https://discord.gg/eBcvAEWz

PR #379, opened 2023-09-24
Feature :  Signup with Email and password 

Hitesh-sisara:
added signup with email and password & added verify email page

/claim #370


app/routes/_auth+/verify-email.tsx
DonKoko, line 1: We have to double check some stuff here. I had an existing session already and when I started my app it broke with an endless loop. We have to make sure that users who already have a session and are logged in, are not getting locked out of the platform.

app/routes/_auth+/join.tsx
```diff
@@ -1,56 +1,216 @@
...
+    );
+  }
+
+  const { email, password } = result.data;
+
+  const existingUser = await getUserByEmail(email);
+
+  if (existingUser) {
+    return json(
+      { errors: { email: "user-already-exist", password: null } },
```
DonKoko: We need a proper error message here.

app/routes/_auth+/join.tsx
```diff
@@ -1,56 +1,216 @@
...
+          <PasswordInput
+            label="Password"
+            placeholder="**********"
+            required
+            data-test-id="password"
+            name={zo.fields.password()}
+            autoComplete="new-password"
+            disabled={disabled}
+            inputClassName="w-full"
+            error={zo.errors.password()?.message || data?.errors?.email}
```
DonKoko: The `or` case should not be here. It should be in the email field as its now rendering the email error in the password field:
Screenshot 2023-09-26 at 17 21 21

app/routes/_auth+/join.tsx
```diff
@@ -1,56 +1,216 @@
...
+      { errors: { email: "user-already-exist", password: null } },
+      { status: 400 }
+    );
+  }
+
+  // Sign up with the provided email and password
+  const signUpResult = await signUpWithEmailPass(email, password);
+
+  // Handle the results of the sign up
+  if (signUpResult.status === "error") {
```
DonKoko, line 92: This logic is overall okey but It should be handled inside the `signUpWithEmailPass` function. We can simply catch and raise an ShelfException from there or even handle stuff like redirects. We don't need to separate the error handling from that function.
The general rule we follow that you can see around is that if `signUpWithEmailPass` has an error its handled inside it. If its successful we can return data or simply `success: true` and handle the success redirect directly within the action, so its easier to overview. Basically how for example our `requireAuthSession` function works.

app/routes/_auth+/join.tsx
```diff
@@ -1,56 +1,216 @@
...
+    return json(
+      { errors: { email: "unable-to-create-account", password: null } },
```
DonKoko, line 94: Here we need a proper error message.

app/routes/_auth+/login.tsx
DonKoko, line 1: I tried to register. 
1. I didnt receive any confirmation email.
2. When I tried to login i get stuck in an endless loop.
```
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 2.210 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 53.376 ms
GET /?_data=routes%2F_index 204 - - 227.360 ms
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 5.422 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 67.374 ms
GET /?_data=routes%2F_index 204 - - 222.601 ms
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 2.500 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 58.197 ms
GET /?_data=routes%2F_index 204 - - 191.071 ms
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 2.084 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 53.200 ms
GET /?_data=routes%2F_index 204 - - 163.506 ms
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 3.963 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 63.339 ms
GET /?_data=routes%2F_index 204 - - 222.290 ms
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 2.571 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 53.964 ms
GET /?_data=routes%2F_index 204 - - 194.617 ms
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 4.458 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 58.769 ms
GET /?_data=routes%2F_index 204 - - 166.870 ms
GET /login?_data=routes%2F_auth%2B%2Flogin 204 - - 2.286 ms
GET /login?_data=routes%2F_auth%2B%2F_auth 200 - - 74.339 ms
^C

```

app/routes/_auth+/verify-email.tsx
DonKoko, line 1: I dont see any code here that runs that sends a verification email. Where is this triggered? I can see the resent function but nothing that sends the first verification email.

app/routes/_auth+/login.tsx
```diff
@@ -74,13 +75,29 @@ export async function action({ request }: ActionArgs) {
       },
       { status: 400 }
     );
+  } else if (signInResult.status === "Email verification_required") {
```
DonKoko: Here I have the same comment as on the join. Please stick to our style of coding and dont use elseifs. 
Just handle each `error case` separately and continue executing the code below on success.

app/routes/_auth+/onboarding.tsx
```diff
@@ -7,47 +7,63 @@ import Input from "~/components/forms/input";
...
-  })
-  .superRefine(
-    ({ password, confirmPassword, username, firstName, lastName }, ctx) => {
-      if (password !== confirmPassword) {
-        return ctx.addIssue({
-          code: z.ZodIssueCode.custom,
-          message: "Password and confirm password must match",
-          path: ["confirmPassword"],
-        });
+function createOnboardingSchema(userSignedUpWithPassword: boolean) {
```
DonKoko, line 20: I like that.

app/routes/_auth+/verify-email.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [searchParams] = useSearchParams();
+  const email = searchParams.get("email") || "";
+  const [message, setMessage] = React.useState<string>();
+
+  async function handleResendEmail() {
+    const formData = new FormData();
+    formData.append("email", email);
+
+    try {
+      const response = await fetch("/resend-email-confirmation", {
```
DonKoko, line 30: Okey this works, but Remix has a better way to do this using fetcher.
I think its better to move this function to one of the utils files or a custom hook and use fetcher, becasue then you dont need to handle errors like this, you can just use the patterns that fetcher allows us to use to show errors/success and so on, wihtout having to do those if..else statements and extra state. That is one of the great things about remix, it does a lot of the heavy lifting and eliminates the need for us to use custom states and follow patterns like this. Basically the `fetcher` var will have all the data to handle this.

app/routes/_auth+/verify-email.tsx
```diff
@@ -0,0 +1,95 @@
...
+        const data = await response.json();
+        setMessage(data.error || "Something went wrong. Please try again.");
+      }
+    } catch (error) {
+      setMessage("Something went wrong. Please try again.");
+    }
+  }
+
+  const messageColor =
+    message === "Email resent successfully. Please check your inbox."
```
DonKoko, line 47: This will also be easier if you use the fetcher from remix.

app/routes/_index.tsx
```diff
@@ -2,17 +2,34 @@ import type { LoaderArgs, LoaderFunction } from "@remix-run/node";
...
+  }
+
+  const { data, error } = await getAuthResponseByAccessToken(
+    authSession.accessToken
+  );
 
-  if (user) {
-    return redirect(user.onboarded ? "/assets" : "onboarding");
+  if (error || !data.user.confirmed_at) {
+    return redirect("/verify-email");
```
DonKoko: I dont think this will work. I think its needs to be without the `/`. I couldnt test it as I couldnt get to that point.

app/routes/_index.tsx
```diff
@@ -2,17 +2,34 @@ import type { LoaderArgs, LoaderFunction } from "@remix-run/node";
...
 
-  const user = authSession
-    ? await getUserByEmail(authSession?.email)
-    : undefined;
+  // If there's no authSession, user isn't logged in. Redirect to login.
+  if (!authSession) {
+    return redirect("login");
+  }
+
+  const { data, error } = await getAuthResponseByAccessToken(
```
DonKoko: I am a bit confused on why we need to do and handle this. Can we not handle this all on the onboarding page directly. Also I am a bit confused why in this case we need to find the auth using token, rather than just get it from the user. Can we not get this information from the user?

app/modules/user/service.server.ts
```diff
@@ -135,31 +130,6 @@ export async function tryCreateUser({
   return user;
 }
 
-export async function createUserAccount(
```
DonKoko, line 138: Not sure why this was removed.

DonKoko (changes requested):
Added some comments that need to be addressed. I personally couldnt do a successfull tests. I didn't receive a confirmation email and hit some endless loops. Also I updated your branch with main to not let it fall far behind, so make sure to pull.

app/routes/_auth+/verify-email.tsx
Hitesh-Sisara: i think, we can use useFetcher hook in combined with form submissions however here it's only arbitrary fetch request. still will check if i can implement it n this case

Hitesh-sisara:
> I am trying to test it but it really doesnt work for me. I create a new user with a new email and i never receive the confirmation email. Does it work for you? Do I need to change anything in supabase config?

you need to turn on email confirmation in supabase

Hitesh-sisara:
> I am trying to test it but it really doesnt work for me
270870424-e6bdae9a-95e9-4458-bbb3-1dcf6d9f64a1
. I create a new user with a new email and i never receive the confirmation email. Does it work for you? Do I need to change anything in supabase config?

DonKoko:
Confirmation is already turned off. AS far as I remember when we register users with email/password we need to build the confirmation url and send the email ourselves. I also explained it in the original issue: https://github.com/Shelf-nu/shelf.nu/issues/370

I just thought u had it figured out in some way because you asked me to test it.

Hitesh-sisara:
> Confirmation is already turned off. AS far as I remember when we register users with email/password we need to build the confirmation url and send the email ourselves. I also explained it in the original issue: https://github.com/Shelf-nu/shelf.nu/issues/370
> 
> 
> 
> I just thought u had it figured out in some way because you asked me to test it.

I mean I figured out why emails are not getting sent, because in previous signup method another supabase signup functions was used  that wouldn't send emails but in my pr i changed them So now it should work, supabase also gives way to generate urls but if this work works we should keep it simple

DonKoko:
@mahendraHegde ooh i didnt understand that. I will test it again now. Thanks man.

DonKoko (dismissed):
PR looks good. I tested it locally and everything works as expected.

DonKoko:
@Hitesh-sisara here is what I have done. 

1. I pulled your branch and fixed the test. The function you removed was used only for tests because it handle what is usually a 2 step process when an actual user registers. I renamed the function to make it easier to understand.
2. I merged with main, seems like it was quite out of date. After doing this I fixed all broken stuff due to v2 upgrade
3. I pushed to your branch to update it so this PR is up to date
4. I pushed to a branch within our repo. From this branch I created a PR to our staging environment so we can test it: https://github.com/Shelf-nu/shelf.nu/pull/446 Once this is fully tested on staging and everything is good we can merge this PR.

DonKoko (approved):

PR #383, opened 2023-09-25
fixed camera permission issue for qr-scan

Hitesh-sisara:
fixed - now site will not ask for camera permission on every scan

### DIFF
--- a/app/components/scan-qr/qrcode-scanner.tsx
+++ b/app/components/scan-qr/qrcode-scanner.tsx
@@ -48,13 +48,13 @@ const QRScanner: React.FC<QRScannerProps> = ({ onClose }) => {
 
               if (match) {
                 stopMediaStream();
-                // const qrId = match[4]; // Get the last segment of the URL as the QR id
+                const qrId = match[4]; // Get the last segment of the URL as the QR id
 
                 setScanCompleted(true); // Set the scanCompleted state to true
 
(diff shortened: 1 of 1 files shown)


app/components/scan-qr/qrcode-scanner.tsx
```diff
@@ -48,13 +48,13 @@ const QRScanner: React.FC<QRScannerProps> = ({ onClose }) => {
...
-                // const qrId = match[4]; // Get the last segment of the URL as the QR id
+                const qrId = match[4]; // Get the last segment of the URL as the QR id
 
                 setScanCompleted(true); // Set the scanCompleted state to true
 
-                window.location.href = scannedData;
+                // window.location.href = scannedData;
 
-                // navigate(`/qr/${qrId}`); (using this way will not close camera access)
+                navigate(`/qr/${qrId}`); (using this way will not close camera access)
```
DonKoko: There is an error here. You left part of the comment

DonKoko (changes requested):

app/components/scan-qr/qrcode-scanner.tsx
Hitesh-Sisara: corrected

DonKoko:
I tested this but it didnt really work. Closing this PR for now, will explore other ways to do it.

PR #384, opened 2023-09-25
fixed camera permission issue for qr-scan

Hitesh-sisara:
now it will not ask  camera permission on every scan


DonKoko (approved):

DonKoko:
@jurrejansen merging this with test site so it can be tested. I am not convinced this will work but can you please test it.

PR #406, opened 2023-09-29
chore: upgrade remix to 2xx

DonKoko:
Upgrading to remix v2.
This comes with some small changes:
- Upgrading to TS 5.x. This means that typescript is more strict so it actually helped us solve a lot of issues that were unknown for us.
- Upgrading remix-utils to @next which is a pre-release. Issue with that is that there are some broken dependencies so we need to `npm i` with legacy peer dependencies. Hopefully that gets fixed soon but we need to keep an eye.


DonKoko:
I unfortunately cannot merge this because of the issues with remix-utils with remix v2. 
The issue is that tests wont run because with the way remix-utils currently works we need to install with `--legacy-peer-deps`. 
https://github.com/sergiodxa/remix-utils/issues/228

@mahendraHegde do you maybe have some ideas how we can go around this?

Issue #369, opened 2023-09-19
[Feature request]: Change data scoping to organization instead of user


DonKoko:
### Is your feature request related to a problem? Please describe?

Currently all data queried in all views is based on `userId`. That is how we know that the assets, categories, etc. that we are querying belong to the current user.
With the introductions of organizations/workspaces this will need to change as we would want everything to be scoped based on organization rather than the user itself.

So I have already taken some steps to achieve this. I have made a functionality that each user has a default organization with type `PERSONAL`. So we have to start with that.

### Describe the solution you'd like

We need to adjust all models and queries to use `organizationId` as well as `userId`. This involves getting data, updating and so on.
Some thoughts:

1. We need a 'global' way in the app to know the current organization. One thing to keep in mind is that for now we are using the PERSONAL organization as default, but in the future once we implement organizations/workspaces, the user will be able to change it. I see 2 options how to approach this:
  - Export the organization in the `_layout` loader and then create a hook using `useMatches` to be able to access it. Big disadvantage is that this works only in the react components. If we need the `organizationId` in a loader or action, we would need to query the db based on the `userId` and get it like that. That's not ideal.
  - Add `organizationId` in the session, so we can access it the same way we do now for the `userId` as part of `requireAuthSession`
I personally think the second option is better, would love to hear your opinion.

2. We need to adjust all our existing db queries in all models to query based on the `organizationId` of the current organization.
3. We need to make migration that adds an `organization` and `organizationId` relation to all models. 
4. We need to create a script which we can run against our DB to update all existing entries and add an `organizationId` of the `PERSONAL` organization

I think this is what we have to handle for now. I am sure I am missing something so please add it if needed.

DonKoko:
@mahendraHegde as the expert on DBs I would love to hear your opinion on this. Can you check out what i wrote and we can then discuss if you have some suggestions and how we can move forward with this task. Thanks

mahendraHegde:
having `organizationId` inside the session seems correct approach.
as per the rest of the steps it seems you have covered everything, we would need to look into the code in depth to be sure that we dint miss any edge cases, we can discuss on this further over discord

DonKoko:
We need to update the loaders and actions as well as the db schema for the following models: 

- [ ] Image
- [ ] Asset
- [ ] Category
- [ ] Tag
- [ ] Note
- [ ] Qr
- [ ] Location
- [ ] CustomField

PR #417, opened 2023-10-02
Add Org support to Asset CRUD

mahendraHegde:
for #369

### DIFF
--- a/app/routes/qr+/$qrId_.contact-owner.tsx
+++ b/app/routes/qr+/$qrId_.contact-owner.tsx
@@ -8,6 +8,7 @@ import { SuccessIcon } from "~/components/icons";
 import { Button } from "~/components/shared/button";
 import { usePosition } from "~/hooks";
 import { getAsset } from "~/modules/asset";
+import { requireAuthSession } from "~/modules/auth";
 import { getQr } from "~/modules/qr";
 import { createReport, sendReportEmails } from "~/modules/report-found";
 import { getUserByID } from "~/modules/user";
@@ -23,6 +24,7 @@ export const NewReportSchema = z.object({
 
 export const action = async ({ request, params }: ActionArgs) => {
   assertIsPost(request);
+  const { organizationId } = await requireAuthSession(request);
   const qrId = params.qrId as string;
   const qr = await getQr(qrId);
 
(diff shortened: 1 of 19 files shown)


app/routes/qr+/$qrId_.contact-owner.tsx
```diff
@@ -23,6 +24,7 @@ export const NewReportSchema = z.object({
 
 export const action = async ({ request, params }: ActionArgs) => {
   assertIsPost(request);
+  const { organizationId } = await requireAuthSession(request);
```
DonKoko: This won't work correctly in this case. This action is performed when you scan an asset that does NOT belong to you. So if we use your `organizationId` it won't work. Thats why in this case we find it based on the `ownerId`. 
I think in this case just querying the asset by its own ID is sufficient, because it doesn't matter to which organization it depends, we just need the specific asset that was scanned.

DonKoko (changes requested):
I just noticed one thing that needs to be adjusted because its not going to work as you made it. I explained the case a bit better so its clear.

app/routes/qr+/$qrId_.contact-owner.tsx
mahendraHegde: thanks for pointing, i have fixed it.

app/routes/qr+/$qrId_.contact-owner.tsx
```diff
@@ -34,7 +34,7 @@ export const action = async ({ request, params }: ActionArgs) => {
 
   const owner = await getUserByID(qr.userId);
   if (!owner) return new Response("Something went wrong", { status: 500 });
-  const asset = await getAsset({ organizationId, id: qr.assetId });
+  const asset = await getAsset({ userId, id: qr.assetId });
```
DonKoko, line 37: Hey man. This is still not correct, i think i might have not explained it correctly. You need to find the asset based on the QR ID because this flow triggers only if the asset doesn't belong to the user performing the action. The 
So you need to search for the asset based on its `qrId` which you can get via the params.
```ts
const qrId = getRequiredParam(params, "qrId");
```

DonKoko (changes requested):

app/routes/qr+/$qrId_.contact-owner.tsx
DonKoko: I took a few minutes to make the changes because I wanted to merge. Thanks for PR, lets continue with the other models.

Issue #423, opened 2023-10-03
[Feature request]:Setup PWA (Remix-PWA)


carlosvirreira:
### Is your feature request related to a problem? Please describe?

I want to use shelf as a PWA on my iPhone

### Describe the solution you'd like

Give me PWA feel for Shelf

DonKoko:
Love the idea. There is a package that can help with that: https://github.com/remix-pwa/remix-pwa
Comes with lots of features and its seems very easy to use with existing apps.

DonKoko:
@kremedev are you still attempting that? I saw you closed your PR. If not please cancel your attempt so someone else can take care of it.

5war00p:
@DonKoko  what do you think on having pure native implenetation for PWA?

carlosvirreira:
Maybe you can tell us why we should consider this @5war00p ?

5war00p:
Because it's fun and  easier by doing little bit of research and its straight forward in my opinion. Also it provides more flexbile for doing things like handle own indexedDB for caching API requests, CDNs and static assets.

You can refer one of my PR where I did caching for static assets for a PWA: https://github.com/heyxyz/hey/pull/3876

carlosvirreira:
We got a warning about this > 

"Service worker, once installed, hard to change if you introduce a bug in its code"

What do you think about that statement @5war00p 

Did you encounter this before?

cc @DonKoko

5war00p:
You can maintain the version for the service-worker script and claims the new worker

DonKoko:
hey @5war00p . First of all welcome to the shelf community and thanks for showing interest in contributing. 
I will be quite happy with our own PWA implementation. Of course with pwa we can go very deep but as a start thats not what we want. As long as our users can take advantage of the basic features of pwa and are able to easily add it to their home screen, we will be happy.
If you want to go ahead, please make an attempt and feel free to open a PR.

5war00p:
@DonKoko Added PWA config without remix-pwa. I haven't added any service worker, the current config only makes the web app installable on desktops and mobiles.

carlosvirreira:
Sounds promising @5war00p 

Curious about the review.

PR #426, opened 2023-10-03
feat: PWA

kremedev:
fixes: #423
/claim #423


app/routes/manifest[.]webmanifest.ts
```diff
@@ -0,0 +1,75 @@
+import { json } from "@remix-run/node";
+
+export const loader = async () =>
+  json(
+    {
+      short_name: "Shelf",
+      name: "Shelf | Asset Management Made Simple",
+      start_url: "/",
+      display: "standalone",
+      background_color: "#d3d7dd",
```
Comment (DonKoko) [inline app/routes/manifest[.]webmanifest.ts:10]: These are not our colors. Please adjust it to the correct colors. You can ask @jurrejansen if you need input about this.

app/routes/manifest[.]webmanifest.ts
```diff
@@ -0,0 +1,75 @@
...
+      theme_color: "#c34138",
+      shortcuts: [
+        {
+          name: "Homepage",
+          url: "/",
+          icons: [
+            {
+              src: "/icons/android-icon-96x96.png",
```
Comment (DonKoko) [inline app/routes/manifest[.]webmanifest.ts:18]: Those are just default remix icons. All of them need to be replaced by shelf icons. Please ask @jurrejansen for help if needed. Maybe you can even use something like this to save time and make it easier: https://tools.crawlink.com/tools/pwa-icon-generator/

package.json
```diff
@@ -3,11 +3,11 @@
   "private": true,
   "sideEffects": false,
   "scripts": {
-    "build": "remix build",
+    "build": "run-s build:*",
```
DonKoko, line 6: This seems like something that will affect a lot of things. Our docker file uses the build command, our testing workflows use it, its used for pre-tests and so on. How will that change affect this?

package.json
```diff
@@ -83,7 +87,14 @@
     "tailwindcss-animate": "^1.0.6",
     "tiny-invariant": "^1.3.1",
     "ua-parser-js": "^1.0.35",
-    "zod": "^3.21.4"
+    "zod": "^3.21.4",
+    "dotenv": "^16.3.1",
+    "@remix-pwa/cache": "^2.0.10",
```
DonKoko, line 92: Those are a lot of new packages. Do we need all the remix-pwa packages?

package.json
```diff
@@ -83,7 +87,14 @@
     "tailwindcss-animate": "^1.0.6",
     "tiny-invariant": "^1.3.1",
     "ua-parser-js": "^1.0.35",
-    "zod": "^3.21.4"
+    "zod": "^3.21.4",
+    "dotenv": "^16.3.1",
```
DonKoko, line 91: Not sure why we need dotenv in dependencies. We already have it in the devDependencies.

DonKoko (changes requested):

PR #442, opened 2023-10-08
feat: pwa configured

5war00p:
/claim #423


public/manifest.json
```diff
@@ -0,0 +1,19 @@
...
+  "name": "shelf-webapp",
+  "lang": "en-US",
+  "short_name": "shelf",
+  "description": "Modern Open Source Asset Management System",
+  "start_url": "/",
+  "display": "standalone",
+  "scope": "/",
+  "theme_color": "#EF6820",
+  "background_color": "#FFFFFF",
+  "icons": [
```
DonKoko, line 11: We need to add more icons as this is not really good like this. For example when I added it on my iphone to the home screen, because of the wrong icon format we have a white border on ios for example. I used this to generate the assets: https://progressier.com/pwa-icons-and-ios-splash-screen-generator

Here is the output: 
[splash_screens.zip](https://github.com/Shelf-nu/shelf.nu/files/12868498/splash_screens.zip)

DonKoko (changes requested):
Overall a good start, but something that we are missing and care a lot about is the popup that suggests users to add it to home screen. Without it nobody will.

public/manifest.json
5war00p: Cool, adding these splash screens. Also, we need to setup different size icons, as of now I added only x1024.

5war00p: Added splash screens, also let me have the different size icons, so that we can add them too.

DonKoko: Oooh I thought I sent you the icons as well in there.  I will provide them in a bit.

app/root.tsx
```diff
@@ -45,6 +45,42 @@ export const links: LinksFunction = () => [
   { rel: "stylesheet", href: styles },
   { rel: "stylesheet", href: fontsStylesheetUrl },
   { rel: "stylesheet", href: globalStylesheetUrl },
+  { rel: "manifest", href: "/manifest.json" },
```
DonKoko, line 49: Can you make a function inside the utils folder that returns all of this and just invoke it here. The file gets a bit too big to overview if all this content is here.

public/manifest.json
DonKoko: Okey according to the site i checked, it should already give the icon. I can see there is an icon file that is 512x512. Can you try replacing the one you had added with it? Then i can test it again on my local.

public/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "shelf-webapp",
+  "lang": "en-US",
+  "short_name": "shelf",
+  "description": "Modern Open Source Asset Management System",
+  "start_url": "/",
+  "display": "standalone",
+  "scope": "/",
+  "theme_color": "#EF6820",
```
DonKoko: When I tested the PWA i can see the bar on top of my iphone is orange. I suspect this is the theme color. Can you change it to the same color as our app's background color so we see how it looks like?

5war00p: Yup will change that

5war00p:
@DonKoko Refactored links and changed theme to white. And I find x512 image anywhere. Could you send that to me?

5war00p:
My bad, didn't checked zip. I'm checking in code. Anyway I updated the icon to x512 now.

DonKoko (commented):
Icon works good now. Can we just add this: https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/Making_PWAs_installable#triggering_the_install_prompt

DonKoko (approved):

PR #448, opened 2023-10-09
add orgID to QR codes

mahendraHegde:
for #369


DonKoko (approved):
This looks good. I will approve it to merge it with the base branch and do some more testing on local once its merged as I have to test a few different scenarios.

PR #453, opened 2023-10-11
add invites feature

mahendraHegde:


.env.example
```diff
@@ -11,4 +11,6 @@ SMTP_HOST="mail.shelf.nu"
 SMTP_USER="no-reply@shelf.nu"
 
 MAPTILER_TOKEN="maptiler-token"
-MICROSOFT_CLARITY_ID="microsoft-clarity-id"
\ No newline at end of file
+MICROSOFT_CLARITY_ID="microsoft-clarity-id"
+
+INVITE_EXPIRY_TTL_DAYS=5
```
DonKoko: I would rather we store that in `constants.ts`. I don't think we really need it as an env variable. LMK if I am missing something.

app/database/schema.prisma
```diff
@@ -264,12 +265,14 @@ model Role {
...
 
-  users User[]
+  users   User[]
+  invites Invite[]
 }
 
 enum Roles {
   USER
   ADMIN
+  TEAM_MEMBER
```
DonKoko: Okey so this I think we need to change. i used this type for roles on application level, which will always be just admin and user. I would rather we create a different type called OrganizationRoles or something like this. Also we should consider storing it in a table instead of a type, as in the future we might add functionality that allows the organization owner to create more roles and adjust their permissions.

DonKoko (commented):
Mostly everything looks good to me. I just added small feedback. Lets discuss the questions later.

.env.example
```diff
@@ -11,4 +11,6 @@ SMTP_HOST="mail.shelf.nu"
 SMTP_USER="no-reply@shelf.nu"
 
 MAPTILER_TOKEN="maptiler-token"
-MICROSOFT_CLARITY_ID="microsoft-clarity-id"
\ No newline at end of file
+MICROSOFT_CLARITY_ID="microsoft-clarity-id"
+
+INVITE_TOKEN_SECRET="secret-test-invite"
```
DonKoko, line 16: We need to add some explanation for this on the docs.

DonKoko (approved):
Looks good. i will merge it and test everything now. Only comment i have is for the env variable for the token. WE need to add it to the docs.

Issue #425, opened 2023-10-03
[Feature request]: Onboarding


jurrejansen:
### Contact Details

jurre@shelf.nu

### Is your feature request related to a problem? Please describe?

New users are not getting an introduction after their first login to their Shelf environment. We should give them a small introduction to help them on their way and already gather some important feedback on how the particular user is planning on using Shelf.

### Describe the solution you'd like

A proper onboarding flow. @carlosvirreira @DonKoko any feedback on the current flow? 

A technical thing I am not sure about. How are we going to store the feedback we're asking the user in one of the steps?

DESIGNS

https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=149-118790&mode=design

DonKoko:
Okey this is great, however we have a few things to discuss and handle:
1. Does this happen after our current onboarding process? We still need to collect some user data like we do now. I would expect this process comes after it
2. How do you want to handle the survey and where do we store the results?
3. We need to figure out the UX in those cases:
  - When you skip the process
  - When you complete the process
  - When you want to restart the process

carlosvirreira:
1. Absolutely, this is post our current onboarding. It is more like a pseudo 'user success/get started' tutorial.
2. If we could use something like Formbricks for this or another OS solution. Super helpful. (We are using it on marketing site and works well)
3. UX:
I see two parts of this flow: 
- the info slides
- the checklist process

For info slides > Press skip, skips that slide. No need to allow for a 'restart' of the slides in my opinion.

The checklist process > From what I know, this is a pre-requirement to 'show' the dashboard. So, I think we might as well push for the 'info slides' part of this until @jurrejansen has a clear UI for the dashboard. 

what do you think @DonKoko

jurrejansen:
Hi @hunar1312 I've updated the UX scenario and added some more annotations that should explain the complete process. You can either leave comments in Figma or ask questions in the Discord thread if needed.

hunar1312:
Hey @jurrejansen, Thank you.

PR #454, opened 2023-10-13
Onboarding welcome

hunar1312:
as described in #425


app/root.tsx
```diff
@@ -63,6 +64,14 @@ export const loader = async ({ request }: LoaderFunctionArgs) =>
 
 export const shouldRevalidate = () => false;
 
+if (typeof window !== "undefined") {
```
DonKoko: We have to look into this. By including it in the route we will be always loading that. We want to have it only on the route where we use formbricks. Also I am not sure why we dont do that inside the component. Its just a front-end thing, right? We could make a client only component and initialize formbriks in a useEffect without being worried if we have `window` available.

app/root.tsx
```diff
@@ -63,6 +64,14 @@ export const loader = async ({ request }: LoaderFunctionArgs) =>
...
+  formbricks.init({
+    environmentId: window.env.FORMBRICKS_ENV_ID,
```
DonKoko: OKey we have to think how to manage this, because if there is no formbricks var, we need to just skip those 2 steps of the slider. We need to consider this for the open source part. Because now someone who uses self hosted version will be basically fucked because of this.

Also doing this: `window.env.FORMBRICKS_ENV_ID` is not correct. You should just import `FORMBRICKS_ENV_ID` from `env.ts` and handle getting the var from .env from there.

app/root.tsx
```diff
@@ -63,6 +64,14 @@ export const loader = async ({ request }: LoaderFunctionArgs) =>
...
+    apiHost: "https://app.formbricks.com",
+    debug: true, // remove when in production
```
DonKoko: should be:
`debug: NODE_ENV === "development"`
NODE_ENV should be also imported from env.ts

Also I noticed that in the types for browser and server NODE_ENV is missing. Can you please add this:

```
      NODE_ENV: "development" | "production" | "test";
```

DonKoko (changes requested):
A general comment related to what we discussed on discord:

1. Keep the welcome folder
2. Rename the layout file to _layout
3. Move onboarding route inside welcome folder

app/components/welcome/carousel.tsx
```diff
@@ -0,0 +1,129 @@
...
+  },
+];
+
+export default function WelcomeCarousel() {
+  return <ClientOnly fallback={null}>{() => <Carousel />}</ClientOnly>;
+}
+
+function Carousel() {
+  const [reachedLastSlide, setReachedLastSlide] = useState(false);
+  useEffect(() => {
```
DonKoko: This will run on every re-render. Shouldn't it just run the first time or when we are at step 2?

app/root.tsx
```diff
@@ -1,4 +1,5 @@
 import type { PropsWithChildren } from "react";
+// import formbricks from "@formbricks/js";
```
DonKoko: remove this please.

app/root.tsx
```diff
@@ -66,6 +67,14 @@ export const loader = async ({ request }: LoaderFunctionArgs) =>
 
 export const shouldRevalidate = () => false;
 
+// if (typeof window !== "undefined") {
+//   formbricks.init({
+//     environmentId: window.env.FORMBRICKS_ENV_ID,
+//     apiHost: "https://app.formbricks.com",
```
DonKoko: Remove this please

app/components/welcome/carousel.tsx
```diff
@@ -0,0 +1,129 @@
...
+    if (FORMBRICKS_ENV_ID) {
```
DonKoko: Also what will happen if there is no ID? We will render empty slides? I would say the slides for the survey also need to be conditionally rendered only if we have the formbrick env variable.

DonKoko (changes requested):
Overall good. I made some comments.
Also please add to the docs a part about the formbricks env variable and how it works. Our open source users need to be able to find this information in the docs.

app/routes/_welcome+/welcome.tsx
DonKoko, line 1: You need to add a loader and protect the route. This is supposed to be just for logged in users.

DonKoko:
@hunar1312 Some more feedback:

- [ ] Restore the background of the form. Right now its completely removed and doesnt look good. It should be within the same card/box as the slides, not just floating into nothing
Screenshot 2023-10-18 at 13 36 47

- [ ] Add 2x size of slides images. On retina display for example it looks pixelated
Screenshot 2023-10-18 at 13 37 05

DonKoko (approved):

DonKoko (approved):

PR #455, opened 2023-10-13
created a default value field for assets

### LINKED ISSUE #436: [Feature request]: Create a default value field for assets
### Contact Details

_No response_

We want to show on the dashboard the total value of assets of the user database (or workspace database). 

Screenshot 2023-10-06 at 12 04 16

We cannot rely on a custom field for this. So we would like to do it with a standard field on shelf.

On asset create, have a value field - so our users fill it up and we can calculate it. 

_No response_

### Additional context

_No response_

hunar1312:
resolves #436


app/database/migrations/20230428075403_add_qr_to_db/migration.sql
```diff
@@ -19,6 +19,3 @@ ALTER TABLE "QR" ADD CONSTRAINT "QR_itemId_fkey" FOREIGN KEY ("itemId") REFERENC
 
 -- AddForeignKey
 ALTER TABLE "QR" ADD CONSTRAINT "QR_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
-
```
DonKoko: We removed this for testing purposes when we were checking the issue you had. There should be no changes in this file. Please revert them

app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -50,6 +51,14 @@ import { deleteAssets } from "~/utils/storage.server";
 
 export async function loader({ request, params }: LoaderFunctionArgs) {
   const { userId } = await requireAuthSession(request);
+  const organization = await getOrganizationByUserId({
```
DonKoko: I dont think you need to do that. You can just adjust the query for asset by including organization and selecting only the currency field.
Something like this. 
```
include: {
  organization: {
    select: {
      currency: true
     },
  }
}
```

and then u can do `asset.organization.currency`

app/routes/_layout+/assets.new.tsx
```diff
@@ -160,12 +169,13 @@ export default function NewAssetPage() {
...
   const [searchParams] = useSearchParams();
   const qrId = searchParams.get("qrId");
+  const { currency } = useLoaderData<typeof loader>();
 
   return (
     <>
       <Header title={title} />
       <div>
-        <AssetForm qrId={qrId} />
+        <AssetForm qrId={qrId} currency={currency} />
```
DonKoko: Not sure why we are passing the currency here. It comes from the loader, just get it inside the asset form itself.

DonKoko (changes requested):
Added some comments. Mostly small changes.

DonKoko (changes requested):
So far so good, however we are missing some stuff.

- [x] displaying value on asset page
- [ ] Setting default currency

The first part you can do now already, but the second part is a bit tricky, as it has to be an option on workspace, but we are still working on the workspaces/organizations, so I think we have to wait for it to be merged before we can merge this as if we dont wait it will create db and code conflicts and its just not worth it. Lets resolve the first one and then mark this as blocked by #369

hunar1312:
https://github.com/Shelf-nu/shelf.nu/pull/455/files#diff-6abae98ca8769c62228e2e3bed3a8388223caed03c9c1aab378bbb4b46e95d15 the first part is already done @DonKoko and we already discussed that we won't move froward till the organization's page is refactored. So yeah we gotta put hold on this one for now.

DonKoko:
@hunar1312 yeah i also did some more testing and I realized we are really blocked by the organization refactoring so we have to wait before we can merge it.

DonKoko:
@hunar1312 updated the PR by merging it with main. There are some changes that will need to be done and you will get some TS errors so better solve them first. Some of the types are moved to a different file as well as now everything is scoped with organizationId instead of userId.

DonKoko:
@hunar1312 I have updated the branch with main. Here is what is still pending:

- [ ] code needs to be updated to use organizationId instead of userId for scoping of everything. I believe TS should already give you errors on those places where orgId is missing because its required.
- [ ] We have a view for editing workspaces. We should add a field where they can change workspace currency.

DonKoko (approved):

PR #469, opened 2023-10-19
feat: Implement "uncategorized" for assets

### LINKED ISSUE #450: [Feature request]: Make category a required field for creating an asset
### Contact Details

_No response_

For dashboard and reporting it is important to have insightful outputs. 

Categories being one of them. 

Ideally when a user creates an asset, it has a default 'uncategorised' category that one can change. 

Alternatively, it can show as a required field, if the user inputs or changes to nothing new it groups it under 'uncategorized'

On asset create, have a dropdown category called uncategorised or allow me to create without category and automatically add the uncategorised category.

_No response_

### Additional context

_No response_

vineetjk:
## Completed tasks in this feature

fixes #450 

- [x] Default category as `Uncategorised` while creating new asset
image

- [x] Created `Uncategorised` badge in asset index page
image

- [x] Created `Uncategorised` badge in single asset page
image

- [x] Added `Uncategorised` to categories filter
image 
image

- After applying Uncategorised` filter in index page
image

### DIFF
--- a/app/components/assets/form.tsx
+++ b/app/components/assets/form.tsx
@@ -138,7 +138,7 @@ export const AssetForm = ({
         className="border-b-0 pb-[10px]"
         required={zodFieldIsRequired(FormSchema.shape.category)}
       >
-        <CategorySelect defaultValue={category || undefined} />
+        <CategorySelect defaultValue={category || "uncategorized"} />
       </FormRow>
 
       <FormRow
--- a/app/components/category/category-select.tsx
+++ b/app/components/category/category-select.tsx
@@ -62,6 +62,11 @@ export const CategorySelect = ({ defaultValue }: { defaultValue?: string }) => {
                 </div>
 
                 <div className="border-b border-b-gray-300 py-2 ">
+                  <SelectItem value={"uncategorized"} key={"uncategorized"}>
+                    <Badge color={"#808080"} noBg withDot={false}>
+                      Uncategorized
+                    </Badge>
+                  </SelectItem>
                   {refinedCategories.map((c) => (
                     <SelectItem value={c.id} key={c.id}>
                       <Badge color={c.color} noBg withDot={false}>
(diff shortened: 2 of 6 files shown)


DonKoko (changes requested):
Okey so overall good, however a few issues

1.  If you try to "clear" (set to uncategorized) the category of an item that already has one assigned it will fail. This case needs to be handled. Basically if the category is `uncategorized` when editing an asset, we have to disconnect the category connected to that asset.

2. Seems like you have some different formatting options on your system or something so lots of stuff got their spacing and formatting adjusted which should not happen. Please solve this. If you dont have it, you can install prettier and that should format it properly as we have `.prettierrc.json` setup.

vineetjk:
> Okey so overall good, however a few issues
> 
> 1. If you try to "clear" (set to uncategorized) the category of an item that already has one assigned it will fail. This case needs to be handled. Basically if the category is `uncategorized` when editing an asset, we have to disconnect the category connected to that asset.
> 2. Seems like you have some different formatting options on your system or something so lots of stuff got their spacing and formatting adjusted which should not happen. Please solve this. If you dont have it, you can install prettier and that should format it properly as we have `.prettierrc.json` setup.

Issues are resolved.

DonKoko (approved):

DonKoko:
hey @vineetjk . I had to unfortunately revert those changes because we had some crashes in live.
So I checked it and I noticed that we have different places where we update asset, we dont only do it in the updateAsset form. You can for example check the file `refresh-main-image`.

So we need to adjust how the update function works. WE will have cases when we are updating and we are not passing a `categoryId` but we dont want the category to be disconnected.

Maybe what you can try is to use `null`.
So if `categoryId` is:
- `undefined` - we dont do anything
- `null` - we disconnect. 

Then we will have to simply make sure that the value of `categoryId` when updating an asset and setting the category to `Uncategorized` is passed as `null`.

Please also search around the codebase and see for other places we are updating asset, as we have to also consider it there. I took a quick look and there is a few places where we use `updateAsset` so they all have to be tested.

PR #476, opened 2023-10-23
Fix uncategorized bug

DonKoko:


DonKoko:
@vineetjk just FYI, this fixes the bug with the uncategorized value when updating asset.

PR #502, opened 2023-10-27
fix: import assets to current organization

mahendraHegde:


app/routes/_layout+/assets.import.tsx
```diff
@@ -40,14 +40,9 @@ export const action = async ({ request }: ActionFunctionArgs) => {
   };
 
   try {
-    const { user } = await assertUserCanImportAssets({ userId });
```
DonKoko, line 43: So here is the thing. Based on the logic of how our premium works, we still need to assert this, but only when the currentOrganization is a of type personal.
So a user needs a certain paid plan to import to personal.
When its a team workspace, we can assume that they can just do it because to have team you need a team subscription. With time we will iron this out but I think for now this logic should be good.

DonKoko (changes requested):
We need to adjust 1 thing because of how our premium works.

app/routes/_layout+/assets.import.tsx
mahendraHegde: my bad!! i fixed it now

Issue #558, opened 2023-11-17
[Feature request]: Link values in custom fields should be actual link in the front-end


DonKoko:
### Is your feature request related to a problem? Please describe?

Our users can set text custom fields on their assets. When the value of a text field is a link it is not clickable on the front-end.

### Describe the solution you'd like

 One thing that we want to do is check weather the value of the string is a link(we can use regex) and if it is, render it using our `<Button/>` component and append to it a ref param. Something like ```to={`${customFieldValue}?ref=shelf-webapp`}```

carlosvirreira:
Thank you to all attempters! 

Let's see which one is picked!

PR #565, opened 2023-11-20
feat(assets): display custom field as link when value is a link

tarunrajput:
/claim #558

closes: #558

Custom field values representing links are currently rendered as unclickable text, limiting user interaction. This PR
render such links using <Button/> component and append to it a ref param (`?ref=shelf-webapp`)


tarunrajput:
@DonKoko, mind taking a look?

DonKoko (commented):
Code looks okey but i need to do some testing first and I can do that tomorrow. Thanks for the PR

Issue #461, opened 2023-10-17
[Feature request]: Administrator requires a custodian to read and sign a document when assigning long term custody


jurrejansen:
### Contact Details

jurre@shelf.nu

### Is your feature request related to a problem? Please describe?

As an IT manager, whenever I check out an asset to an employee. I want to have the option to enhance the 'formality' of the checkout. Recording the date/time/IP address and signature of an employee that received an asset from me would do this. In case this employee leaves the company. We can then rightfully request the return of all the assets that he / she got access and signed off for.

### Describe the solution you'd like

Administrator can attach a pre-configured PDF template document when assigning custody. Optionally the administrator can decide if the custodian needs to leave a signature electronically before custody of the asset is officially assigned. We are going to have both non-registered team members and registered team members (users) that actually have an account / email. Both can be custodians.

NON-REGISTERED TEAM MEMBERS
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=534-92886&mode=design

REGISTERED TEAM MEMBERS (USERS)
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=626-29627&mode=design

jurrejansen:
@carlosvirreira @DonKoko Any comments here? Ignore the non-registered team members designs as we discussed to not implement this option when assigning custody to non-registered team members.

carlosvirreira:
If we have to ignore 1 flow then strikethrough it. 

~~before the text~~

carlosvirreira:
Note for the future bounty attempter. You might see there some styles that are new layout wise. I want to ask you to please ignore them and focus on the new dashboard setting tab and the fields inside of it only.

This bounty is part of a bigger epic (collect signatures for custody assignments).

Any questions about this bounty can be discussed on our Discord: https://discord.gg/nGdu8zHG

feliciien:
Hello when i run npm run setup  i have this error message Error: P1001: Can't reach database server at `db.nxphroyicgnenkxwolrg.supabase.co`:`5432`

Please make sure your database server is running at `db.nxphroyicgnenkxwolrg.supabase.co`:`5432`.
ERROR: "setup:db" exited with 1.

or I already add the URL database here https://supabase.com/dashboard/project/nxphroyicgnenkxwolrg/settings/database

[...]

PR #577, opened 2023-11-29
Feat/template

### LINKED ISSUE #462: [Feature request]: Administrators can configure PDF template documents used for assigning Custody and Bookings
### Contact Details

jurre@shelf.nu

As an IT manager, whenever I check out an asset to an employee. I want to have the option to enhance the 'formality' of the checkout. Recording the date/time/IP address and signature of an employee that received an asset from me would do this. In case this employee leaves the company. We can then rightfully request the return of all the assets that he / she got access and signed off for.

Administrators can navigate to Settings > Templates to manage PDF templates that can be used for assigning Custody and Bookings. A PDF template has the following fields.

- Type* | Indicates for what the template will be used (Custody, Bookings, ...)
- Name* 
- Signature | A toggle to specify if a signature is required
- Description
- Upload PDF* | Here the PDF document can be uploaded that will be shown to custodian for reading (and signing)

DESIGNS
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=544-176481&mode=design

_No response_

### Additional context

_No response_

rajdip-b:
Fixes #462

Features added:
- **Templates** tab under **Settings**
- Add template
- Update template
- Set a template as default
- Activate / De-activate a template

Although, there are some issues I wasn't able to fix. Listing them here:
- ~form doesnt submit when  signature required  is set to true~
- ~template pdfs are not being update upon submission of a new pdf~
- ~file picking in the dropzone manually works as expected, but fails when files are dropped~

/claim #462


rajdip-b:
@DonKoko @carlosvirreira

app/components/workspace/nrm-actions-dropdown.tsx
```diff
@@ -49,7 +49,7 @@ export function TeamMembersActionsDropdown({
         align="end"
         className="order w-[180px] rounded-md bg-white p-[6px] text-right "
       >
-        <DropdownMenuItem className="text-gray-700hover:text-gray-700 p-4 hover:bg-slate-100">
+        <DropdownMenuItem className="p-4 text-gray-700 hover:bg-slate-100 hover:text-gray-700">
```
DonKoko, line 52: Nice find. Thanks for fixing that.

app/database/migrations/20231129121343_/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Template" ADD COLUMN     "fileUrl" TEXT;
```
DonKoko: We should remove this migration and adjust the next one. If this was deployed to live, then we would indeed need the next migration to solve it but as we are still on test we can simply remove that one and adjust the next one. You can then run `npm run reset:db`. That will clear your DB and run all migrations again.

app/database/migrations/20231129154756_/migration.sql
```diff
@@ -0,0 +1,10 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `fileUrl` on the `Template` table. All the data in the column will be lost.
+
+*/
+-- AlterTable
+ALTER TABLE "Template" DROP COLUMN "fileUrl",
```
DonKoko: Remove this. See comment on other migration file.

app/database/migrations/20231125054123_/migration.sql
DonKoko, line 1: The folder name is not good. You should give names to all migrations.

app/database/migrations/20231129121343_/migration.sql
DonKoko, line 1: Migration name missing

app/database/migrations/20231129154756_/migration.sql
DonKoko, line 1: Migration name missing.

app/routes/_layout+/settings.template.add-template.tsx
DonKoko, line 1: Our convention for file names is to user `new` for the routes for create.

DonKoko (commented):
Okey so I checked some files and pulled it locally. Ran migrations and so on. When I started testing, I faced some issues. I want to create a new template but when I click "save" nothing happens. Either some validations are failing and errors are not being shown or something is happening with the posting of the form.

app/components/templates/form.tsx
```diff
@@ -0,0 +1,258 @@
...
+        inputType="textarea"
+        name={zo.fields.pdfSize()}
+        disabled={disabled}
+        className="w-full border-b-0 pb-0"
+        value={size}
+        placeholder="Store the booking arrangement for 2023"
+        required={zodFieldIsRequired(NewTemplateFormSchema.shape.pdfSize)}
+      /> */}
+      <FormRow rowLabel="Upload PDF">
+        <FileDropzone
```
DonKoko: I think we might have to drop that. Its not working as expected and it will be so much easier if we just use a normal file input, like we do in most places. I know this was in the design but thats an oversight on my side that I didn't let the designer know it should be changed.

app/components/templates/form.tsx
```diff
@@ -0,0 +1,258 @@
...
+import { FileDropzone } from "../shared/file-dropzone";
+import iconsMap from "../shared/icons-map";
+import { Spinner } from "../shared/spinner";
+
+export const NewTemplateFormSchema = z.object({
+  name: z.string().min(2, "Name is required"),
+  type: z.nativeEnum(TemplateType),
+  description: z.string().optional(),
+  signatureRequired: z.boolean().optional(),
+  pdf: z.any().optional(),
```
DonKoko: PDF should be required.

app/routes/_layout+/settings.template.add-template.tsx
```diff
@@ -0,0 +1,114 @@
...
+  breadcrumb: () => <span>{title}</span>,
+};
+
+export async function action({ request }: LoaderFunctionArgs) {
+  const authSession = await requireAuthSession(request);
+  assertIsPost(request);
+  await assertUserCanCreateMoreTemplates({ userId: authSession.userId });
+
+  const formData = await request.formData();
+  const clonedData = request.clone();
```
DonKoko: I think you need to clone before creating the formData. I tested this again and i get the following error:
```
TypeError: This stream has already been locked for exclusive reading by another reader
    at new ReadableStreamDefaultReader (/Users/donkoko/devs/shelf/app/node_modules/web-streams-polyfill/src/lib/readable-stream/default-reader.ts:108:13)
    at AcquireReadableStreamDefaultReader (/Users/donkoko/devs/shelf/app/node_modules/web-streams-polyfill/src/lib/readable-stream/default-reader.ts:32:10)
    at ReadableStreamDefaultTee (/Users/donkoko/devs/shelf/app/node_modules/web-streams-polyfill/src/lib/readable-stream/tee.ts:59:18)
    at ReadableStreamTee (/Users/donkoko/devs/shelf/app/node_modules/web-streams-polyfill/src/lib/readable-stream/tee.ts:51:10)
    at ReadableStream.tee (/Users/donkoko/devs/shelf/app/node_modules/web-streams-polyfill/src/lib/readable-stream.ts:292:22)
    at clone (/Users/donkoko/devs/shelf/app/node_modules/@remix-run/web-fetch/src/body.js:290:29)
    at new Request (/Users/donkoko/devs/shelf/app/node_modules/@remix-run/web-fetch/src/request.js:98:6)
    at Request.clone (/Users/donkoko/devs/shelf/app/node_modules/@remix-run/web-fetch/src/request.js:246:10)
    at action37 (/Users/donkoko/devs/shelf/app/app/routes/_layout+/settings.template.add-template.tsx:55:30)
    at Object.callRouteActionRR (/Users/donkoko/devs/shelf/app/node_modules/@remix-run/server-runtime/dist/data.js:35:16)
    ````

File: app/routes/_layout+/settings.template.add-template.tsx
Comment (DonKoko) [reply app/routes/_layout+/settings.template.add-template.tsx:1]: I still see this. the file name is `settings.template.add-template.tsx` and it should be `settings.template.new.tsx`

File: app/database/migrations/20231204172150_add_template/migration.sql [hunk 7]
```diff
@@ -0,0 +1,27 @@
...
+    "isActive" BOOLEAN NOT NULL DEFAULT true,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+    "userId" TEXT,
+
+    CONSTRAINT "Template_pkey" PRIMARY KEY ("id")
+);
+
+-- AddForeignKey
+ALTER TABLE "Template" ADD CONSTRAINT "Template_userId_fkey" FOREIGN KEY ("userId") REFERENCES "User"("id") ON DELETE CASCADE ON UPDATE CASCADE;
```
DonKoko, line 28: Just one thing missing. Supabase has something called RLS. Its a security feature that is disabled on tables by default. Every time you create a new model you need to enable it. Please add this to the end of your migration file:

```
-- Enable RLS
ALTER TABLE "Template" ENABLE row level security;
```

app/routes/_layout+/settings.template.add-template.tsx
rajdip-b: on it!

app/utils/crop-image.ts
```diff
@@ -1,18 +1,10 @@
 import type { ResizeOptions } from "sharp";
 
-export const cropImage = async (
-  data: AsyncIterable<Uint8Array>,
-  options?: ResizeOptions
-) => {
-  const chunks = [];
```
DonKoko, line 7: I can see you placed this in a separate function, but I am not sure why. Its not being re-used more than once from what I see and calling it outside the `cropImage` function is just an extra step the developer needs to do and can cause mistakes. Its fine to have it in a different function if we will re-use it, but then invoke that function inside `cropImage`, rather than having to do it separately every time we invoke `cropImage`

app/modules/tier/service.server.ts
```diff
@@ -88,6 +88,25 @@ export const assertUserCanCreateMoreCustomFields = async ({
...
 };
 
+export const assertUserCanCreateMoreTemplates = async ({
+  userId,
+}: {
+  userId: User["id"];
+}) => {
+  /** Get the tier limit and check if they can export */
+  const tierLimit = await getUserTierLimit(userId);
+  const canCreateMore = canCreateMoreCustomFields({
```
DonKoko: Its good you made this function to assert the template creation, but inside it you are calling `canCreateMoreCustomFields` which is not correct. I believe you want to call the function for templates.

app/modules/template/template.server.ts
DonKoko, line 1: FIle should be called `service.server.ts` inside the template folder.

app/modules/template/template.server.ts
```diff
@@ -0,0 +1,215 @@
...
+      connect: {
+        id: userId,
+      },
+    },
+  };
+
+  const template = await db.template.create({ data });
+
+  // Check whether there are templates of same type for the user
+  const result = await db.template.findMany({
```
DonKoko: I would do this check before doing the create. Then you dont need to update it later, you can just set it on default based on this. Also you should just use `count` instead of `findMany`. Its more performant and its gives you all the info you need. Also I think the name `result` for this var is very confusing.

app/modules/template/template.server.ts
```diff
@@ -0,0 +1,215 @@
...
+  });
+  // If there is only one template, make it default
+  if (result.length === 1) {
+    await db.template.update({
+      where: { id: result[0].id },
+      data: { isDefault: true },
+    });
+  }
+
+  template.isDefault = result.length > 0 ? false : true;
```
DonKoko: If you follow the comment from above, this wont be needed.

app/components/templates/template-actions-dropdown.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const map = new Map<string, TTemplate>();
+    items.forEach((item) => {
+      if (item.isDefault) map.set(item.type, item);
+    });
+    return map;
+  });
+
+  return (
+    <>
+      <DropdownMenu modal={false}>
```
DonKoko, line 34: Something is off with the design here:
Screenshot 2023-12-06 at 16 28 24

1. Make Default looks wrong
2. Icon of deactivate looks kinda bad. Are those the final icons that Jurre delivered?

app/components/templates/form.tsx
```diff
@@ -0,0 +1,255 @@
...
+            required={!isEdit}
+            disabled={disabled}
+            accept="application/pdf,.pdf"
+            name={"pdf"}
+            type="file"
+            size={5_000_000}
+            onChange={handleFileChange}
+            label={""}
+            hideLabel
+            error={fileError}
```
DonKoko: Something wrong here with this. When I submit the form without adding a PDF, it fails, so that means the validation works but I don't see an error message.

In the meantime file size validation for example works, however there is a small issue. Even though an error is shown the PDF is still shows as added to the form on the dom. That should not be the case:
Screenshot 2023-12-06 at 16 35 11

app/components/templates/template-actions-dropdown.tsx
```diff
@@ -0,0 +1,192 @@
...
+          disabled={true}
+          icon="star"
+          variant="tertiary"
+          className="border-0"
+          width="full"
+        >
+          Make default
+        </Button>
+      ) : typeDefault && typeDefault.id !== template.id ? (
+        <AlertDialog>
```
DonKoko, line 119: This is not styled correctly. It looks like a delete alert modal, and it shouldn't.
Screenshot 2023-12-06 at 16 37 09

Please refer to design: https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=544-217286&mode=design&t=ou3VY2BWfFjDjVAj-0

app/routes/_layout+/settings.template.index.tsx
```diff
@@ -0,0 +1,281 @@
...
+    title: "Templates",
+  });
+};
+
+export async function action({ request }: ActionFunctionArgs) {
+  assertIsPost(request);
+  const authSession = await requireAuthSession(request);
+  const userId = authSession.userId;
+  const url = new URL(request.url);
+  const id = url.searchParams.get("templateId");
```
DonKoko: Okey so this is a bit of a situation we have here. What you did technically works, but it doesn't fit to the standards we follow. 
We follow Web standards and we have certain conventions on how we manage data so you will need to slightly rework this. I will add more comments to address the specific parts.

app/routes/_layout+/settings.template.index.tsx
```diff
@@ -0,0 +1,281 @@
...
+export const ErrorBoundary = () => <ErrorBoundryComponent />;
+
+export default function TemplatePage() {
+  const { items, canCreateMoreTemplates, tier, totalItems } =
+    useLoaderData<typeof loader>();
+
+  const hasItems = totalItems > 0;
+
+  let upgradeMessage =
+    "You are currently able to create a max of 10 templates. If you want to create more than 10 Team templates, please get in touch with sales";
```
DonKoko: Just curious, did those numbers come from the design or Jurre or its something you felt makes sense?

app/components/templates/template-actions-dropdown.tsx
```diff
@@ -0,0 +1,192 @@
...
+  template,
+}: {
+  typeDefault?: TTemplate;
+  template: TTemplate;
+}) => {
+  const fetcher = useFetcher();
+  const disabled = isFormProcessing(fetcher.state);
+  const submit = useSubmit();
+
+  const handleMakeDefault = () => {
```
DonKoko: Okey so this might technically work, but its not the correct way to do it in Remix.
You need to use a Form and submit that, not manually do it. One of the best things about remix is that it gives you the tools to follow html and web standards and you dont have to do this manual hacky js submissions. Let me give you an example:

```
<Form method="post" > // you dont need to pass the action as its within the same route
  <input type="hidden" value={template.isActive} name="isActive" />
  <input type="hidden" value={template.id} name="id" />
  .... /// add hidden fields for all your other data that you need to send.
  <Button type="submit" name="intent" value="makeDefault">Make default</Button>
</Form>
```

This is an example that doesnt include all the fields but you get the idea. Our convention is to add the "intent" keyword to manage intent. You can see this in many examples in shelf: 
- example of form: https://github.com/Shelf-nu/shelf.nu/blob/main/app/components/workspace/delete-member.tsx#L86-L102
- example of how to handle the cases for intent in the action: https://github.com/Shelf-nu/shelf.nu/blob/main/app/routes/_layout%2B/assets.import.tsx#L50-L63
- Example on how to get formData sent via Form to an action: https://github.com/Shelf-nu/shelf.nu/blob/main/app/routes/_layout%2B/assets.%24assetId.give-custody.tsx#L57

app/components/templates/template-actions-dropdown.tsx
```diff
@@ -0,0 +1,192 @@
...
+              className="justify-start text-gray-700 hover:text-gray-700"
+              width="full"
+            >
+              Edit
+            </Button>
+          </DropdownMenuItem>
+          <DropdownMenuItem className="px-4 py-3">
+            {/* <Form method="post"> */}
+            {/* <input type="submit" value={"submit"} /> */}
+            <Button
```
DonKoko: Here we have the same case as with the "make default". Please adjust it to follow Web standards and send data via a form with POST.

app/components/templates/template-actions-dropdown.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+const MakeDefaultButton = ({
+  typeDefault,
+  template,
+}: {
+  typeDefault?: TTemplate;
+  template: TTemplate;
+}) => {
+  const fetcher = useFetcher();
+  const disabled = isFormProcessing(fetcher.state);
```
DonKoko: I dont think this will do anything as you are not using the fetcher to submit. You are submitting using `useSubmit`'s function so you should be looking at navigation for the state. This can be removed as you should handle it all with a form.

DonKoko (commented):
I wrote you down a lot of feedback. For now you have to change mostly stuff related to coding style and how we do stuff at shelf. Lets do this and then we continue.
Also I checked the modal for assigning custody and I dont see the template choice showing. We need to also work on that and on the email part. There is still quite some logic to be handled.

app/utils/crop-image.ts
rajdip-b: My intention here was to use crop only when it's required. When I started working, every operation related to file was pivoted around images. So most of the places had logical errors when I also added the `.pdf` support. In this case, before I started to code, the `cropImage` function was coded into the place where we upload images as far as I recall. The problem with that was, we don't want to resize PDFs (or do we?). And hence I moved this to a separate function and call it only when the upload file is an image. 

P.S. I'm not entirely sure about what the last line meant.

app/components/templates/template-actions-dropdown.tsx
rajdip-b: 1. Fixing that (dunno why it shows like that)
2. No, it was sort of a placeholder, and I absolutely forgot about asking for the icon. Can you send the svg code in here? I don't have SVG copy access in figma

app/components/templates/form.tsx
rajdip-b: 1. the validation is taking place in the `action` function of the creat new template component. I am however sending an error. But I have no idea about how I should display it. I feel this can be a lot easier if we do the validation in the form itself, but that too isn't happening.

2. Was unsure about this, I'll fix that

app/routes/_layout+/settings.template.index.tsx
rajdip-b: Yeah, actually I put up this question when I initially started with, but I think I didn't get any response or so. And hence I thought about moving with a dummy value for now. What should the value be btw?

rajdip-b: UPDATE: @carlosvirreira told me to set it to 3. I have updated the code accordingly

app/components/templates/form.tsx
rajdip-b: P.S. went through [this](https://github.com/colinhacks/zod/issues/387) article. seems like there is a problem with zod file input. A little bit of help would be really appreciated

app/database/migrations/20231204172150_add_template/migration.sql
DonKoko: @rajdip-b the way you did it doesnt work. It breaks the migrations. Please do it as I showed it. Add it as a separate query below the AddForeignKey query.

rajdip-b: sure!

app/utils/crop-image.ts
DonKoko: OKey your point is valid but from what I see you moved just the part that creates the buffer. Its fine that `getFileArrayBuffer` is in separate function, but we shouldn't have to manually call it every time before we call `cropImage` so you should invoke it inside `cropImage` but then use it by itself when u need a buffer for a PDF for example.

rajdip-b: ohh! got your point now. sure, i'll make the updates!

rajdip-b: image
image

This works right?

app/components/templates/template-actions-dropdown.tsx
DonKoko:
```ts
const ToggleIcon = (props: SVGProps<SVGSVGElement>) => (
  <svg
    xmlns="http://www.w3.org/2000/svg"
    width={24}
    height={24}
    fill="none"
    {...props}
  >
    <path
      stroke="#000"
      strokeLinecap="round"
      strokeLinejoin="round"
      strokeWidth={2}
      d="M2 12a6 6 0 0 1 6-6h8a6 6 0 0 1 0 12H8a6 6 0 0 1-6-6Z"
    />
    <path
      stroke="#000"
      strokeLinecap="round"
      strokeLinejoin="round"
      strokeWidth={2}
      d="M8 14.5a2.5 2.5 0 1 0 0-5 2.5 2.5 0 0 0 0 5Z"
    />
  </svg>
)
```

rajdip-b: Thanks for the SVG, but @carlosvirreira already sent that to me yesterday :D

app/components/templates/form.tsx
DonKoko: Okey so before we never validated files using zod. I am not sure why, maybe due to laziness maybe some other reason, but in this case that we have required files, I think its nice to create a function that can help us validate zod file fields. You can then manage the error on the front-end like with any other zod field.
So please take a look at this: https://github.com/colinhacks/zod/issues/387#issuecomment-1191390673

It provides a full schema that validates type, size, and that its present. There are also different solutions in this file but i think this one is the one i like the most.

app/routes/_layout+/settings.template.index.tsx
DonKoko: @rajdip-b i think its also important to create a migration that adds the value to the tier limits model.

rajdip-b: image
I am already using this in the `schema.prisma` file. Would a migration be absolutely necessary?

DonKoko:
@rajdip-b your updated PR is really nice. You resolved almost everything however I manually had to unresolve some feedback as I cant test it. I cant test it because the migration file is broken. LMK when you solve it so I can finish testing.

app/components/templates/template-actions-dropdown.tsx
DonKoko: Okey so idk what carlos sent you but I also noticed something else and I wanted to explain to you.
Screenshot 2023-12-07 at 13 55 48

If you check our other svgs in the library you will notice most of them on the color fields have the value `currentColor` instead of a color. THat means they will inherit the color of the parent. You need to adjust the SVG's you added to work like this because if not, then we get the issue like in the screenshot.

app/routes/_layout+/settings.template.index.tsx
```diff
@@ -0,0 +1,286 @@
...
+        <span className="text-text-sm font-medium text-gray-900">
+          {item.name}
+        </span>
+        <span className="text-text-sm font-light text-gray-600">
+          {item.type}
+        </span>
+      </div>
+    </Td>
+    <Td>
+      {item.isDefault && (
```
DonKoko, line 274: Lets add also a badge for disabled. Text should be "Inactive" and lets Make it gray. You can find some example of the gray color we use for badges in the code.
Screenshot 2023-12-07 at 14 07 25

DonKoko (changes requested):

DonKoko:
hey @rajdip-b . I have looked and the last changes you did are good. A few things:

1. How does it work with the files. I didnt create a bucket called templates but the upload "seems" to work. Does it actually work? I don't think so but I don't get an error. This should be handled what happens if the bucket doesnt exist. We should throw an error saying to contact admin.
2. Adding the templates and so on is nice, but we need to also build the functionality for using them. I know that @jurrejansen created a few different issues for this, so maybe he can point you to the other related issues.

rajdip-b:
> hey @rajdip-b . I have looked and the last changes you did are good. A few things:
> 
>     1. How does it work with the files. I didnt create a bucket called templates but the upload "seems" to work. Does it actually work? I don't think so but I don't get an error. This should be handled what happens if the bucket doesnt exist. We should throw an error saying to contact admin.
> 
>     2. Adding the templates and so on is nice, but we need to also build the functionality for using them. I know that @jurrejansen created a few different issues for this, so maybe he can point you to the other related issues.

1. I didn't see this coming TBH. I'll add this check asap!
2. I can start working on this if you link the issues

carlosvirreira:
Hi @rajdip-b - these two issues are 100% related to the issue you are working on now.

https://github.com/Shelf-nu/shelf.nu/issues/462
https://github.com/Shelf-nu/shelf.nu/issues/461

rajdip-b:
@carlosvirreira I think someone is already working on #461. Maybe merging them will do the trick once both are finished? I am not sure though!

DonKoko (approved):
This PR is good for now. I will merge it with `signing-templates` branch so it can be picked up for the rest of the feature to be built.

rajdip-b:
> This PR is good for now. I will merge it with `signing-templates` branch so it can be picked up for the rest of the feature to be built.

Sounds good!

PR #583, opened 2023-12-04
add scheduled and non scheduled emails for booking; make booking name required

mahendraHegde:


app/utils/env.ts
```diff
@@ -13,6 +13,7 @@ declare global {
       ENABLE_PREMIUM_FEATURES: string;
       FORMBRICKS_ENV_ID: string;
       MAINTENANCE_MODE: string;
+      DATABASE_URL: string;
```
DonKoko, line 16: Do we need to add it to the Window? That could cause some security issues as well right?

DonKoko (approved):

app/utils/env.ts
mahendraHegde: it should not as its just for typings, but let me remove

PR #600, opened 2023-12-12
fix: importing when no teammembers are present as well as handling …

DonKoko:
…non-printable charaters for csv columns

A user had an issue with importing. After investigating we found out there were 2 issues with importing:

1. If no `custodian` column exists for asset, it would fail as it would try to create `teamMembers` based on broken data
2. When creating the `data` array for creating assets from import, we didn't trim the key of each element and there were some Non-printable characters inside of it.


DonKoko (commented):
image

PR #602, opened 2023-12-13
feat: add more granular roles and permissions

### LINKED ISSUE #588: [Feature request]: Inviting a user and assigning them a 'Self-service' user role with limited permissions
### Contact Details

jurre@shelf.nu

After booking functionality is implemented some Shelf users / workspaces will require a user role that is only permitted to create a reservation. After that an administrator user can take care of doing the actual Check-out and Check-in to complete the booking process.

When inviting a user to a workspace the administrator or account owner can choose the user role from a dropdown. I'm thinking the 'Self-service' user role needs to be a default system role.

A small description of the 'Self-service' user role. 

"Users with the Self-service user role can reserve assets."

Asset overview: The Self-service user can see the assets but is unable to create / update / delete. They are also not allowed to update locations, or give and take custody over assets.

This means the Self-service user is able to see all assets in the index except for assets that are marked as 'Unavailable for bookings'

Creating booking reservations for themselves: Self-service users are able to generate their own reservations but are unable to perform a check-out or initiate a check-out from the beginning. **An administrator will need to handle the check-out and check-in process.**

Cannot manage settings: The Self service user cannot change workspace settings nor invite or update other account users. They can only change settings related to their own account. User details etc.

Their UI will probably look similar to this:

Bookings
Index
Asset

Wwith the eye on the future, permission will have to be structured kind of modular in a sense. Meaning the permissions that are attached to a user role should be categorized so at a later stage custom user roles can be made and certain permissions can be toggled on/off per user role. Hope that makes sense. Please don't hesitate to let me know if anything is unclear whoever picks up this issue. Thanks!

DESIGNS

     ↳ Inviting new users to a workspace
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=2059-21594&mode=design

Now we're introducing a 2nd role it would be good to introduce the Edit action for users so administrators can change the role of a user

     ↳ Edit the role of a user
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=510-271842&mode=design

_No response_

### Additional context

_No response_

mahendraHegde:
closes #588

### DIFF
--- a/app/utils/permissions/types.ts
+++ b/app/utils/permissions/types.ts
@@ -1,0 +1,30 @@
+import type { OrganizationRoles } from "@prisma/client";
+
+export enum PermissionAction {
+  create = "create",
+  read = "read",
+  update = "update",
+  delete = "delete",
+  checkout = "checkout",
+  checking = "checkin",
+  export = "export",
+  import = "import",
+}
+export enum PermissionEntity {
+  asset = "asset",
+  booking = "booking",
+  tag = "tag",
+  category = "category",
+  location = "location",
+  customField = "customField",
+  workspace = "workspace",
+  teamMember = "teamMember",
+}
+
+export interface PermissionCheckProps {
+  organizationId: string;
+  roles?: OrganizationRoles[];
+  userId: string;
+  action: PermissionAction;
+  entity: PermissionEntity;
+}
--- a/app/database/migrations/20231213225815_add_self_service_org_role/migration.sql
+++ b/app/database/migrations/20231213225815_add_self_service_org_role/migration.sql
@@ -1,0 +1,2 @@
+-- AlterEnum
+ALTER TYPE "OrganizationRoles" ADD VALUE 'SELF_SERVICE';
--- a/app/database/schema.prisma
+++ b/app/database/schema.prisma
@@ -378,6 +378,7 @@ enum OrganizationRoles {
   ADMIN
   TEAM_MEMBER
   OWNER
+  SELF_SERVICE
 }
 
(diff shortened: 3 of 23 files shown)


app/utils/permissions/types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export enum PermissionAction {
+  create = "create",
+  read = "read",
+  update = "update",
+  delete = "delete",
+  checkout = "checkout",
+  checking = "checkin",
+}
+export enum PermissionEntity {
```
DonKoko, line 13: I think we should add the rest as well:
- tags
- categories
- locations
- custom fields
- workspaces
- team members

I think thats it. Idk if I am missing something

DonKoko (commented):
Looks good to me. Just added 1 comment/question. I would need to test it and get some more examples how to use it.
Also I am wondering about inviting users to a workspace. Should we adjust that also to handle the different roles?

app/utils/permissions/types.ts
mahendraHegde: yes, its really a basic draft, just to show the idea, i have to add all these.

mahendraHegde:
> Looks good to me. Just added 1 comment/question. I would need to test it and get some more examples how to use it. Also I am wondering about inviting users to a workspace. Should we adjust that also to handle the different roles?

we already handle it in BE, same role used to invite will be assigned, FE changes are required

DonKoko (approved):

PR #609, opened 2023-12-15
booking backend changes

mahendraHegde:
- fix(bookings): show assets from current booking in the asset select dialog
- feat(bookings): update asset status on booking status change
- feat(bookings): use scheduler to auto update booking status to overdue

### DIFF
--- a/app/database/schema.prisma
+++ b/app/database/schema.prisma
@@ -117,7 +117,6 @@ view AssetSearchView {
 enum AssetStatus {
   AVAILABLE
   IN_CUSTODY
-  BOOKED
   CHECKED_OUT
 }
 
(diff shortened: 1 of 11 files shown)


app/database/schema.prisma
```diff
@@ -117,7 +117,6 @@ view AssetSearchView {
 enum AssetStatus {
   AVAILABLE
   IN_CUSTODY
-  BOOKED
   CHECKED_OUT
```
DonKoko, line 120: I just realized we need 1 more status. When an asset has `availableToBook` set to false, we need to change its status from AVAILABLE to UNAVAILABLE

DonKoko (approved):

PR #625, opened 2023-12-22
backend fixes to bookings

mahendraHegde:
- fix(bookings): use localised date in booking emails
- fix(assets): return available to book assets only when booking search is made
- fix(bookings): send checkin reminder immediately if checkin reminder deadline has passed already
-   fix(bookings): update asset status for assets added after booking started

### DIFF
--- a/app/modules/asset/service.server.ts
+++ b/app/modules/asset/service.server.ts
@@ -212,6 +212,9 @@ export async function getAssets({
       message: "booking dates are needed to hide unavailable assets",
     });
   }
+  if (bookingFrom && bookingTo && where.asset) {
+    where.asset.availableToBook = true;
+  }
 
   if (tagsIds && tagsIds.length > 0 && where.asset) {
     where.asset.tags = {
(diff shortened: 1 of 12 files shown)


app/modules/asset/service.server.ts
```diff
@@ -212,6 +212,9 @@ export async function getAssets({
       message: "booking dates are needed to hide unavailable assets",
     });
   }
+  if (bookingFrom && bookingTo && where.asset) {
```
DonKoko, line 215: @mahendraHegde can we put a comment here what is happening. This function handles lots of cases and it will help a lot to other developers as well.

DonKoko (approved):

DonKoko:
Testing: 

- [ ] fix(bookings): use localised date in booking emails
- [ ] fix(assets): return available to book assets only when booking search is made
- [ ] fix(bookings): send checkin reminder immediately if checkin reminder deadline has passed already
- [ ] fix(bookings): update asset status for assets added after booking started

PR #629, opened 2023-12-23
Feature: Add template signing while assigning custody

### LINKED ISSUE #461: [Feature request]: Administrator requires a custodian to read and sign a document when assigning long term custody

rajdip-b:
Fixes #461 

Aims to send a PDF template along with an asset custody for the custodian to sign.

/claim #461


carlosvirreira:
We have a 'read more' link on a modal for signature. If we could make it hit this URL it would be great. 

https://www.shelf.nu/blog/efficient-asset-management-with-shelfs-new-signed-custody-feature

PR #647, opened 2024-01-08
enhanced email templates

### LINKED ISSUE #642: [Feature request]: Enhanced email templates using react
Currently we send emails with main content only, We need enhanced email templates for all use cases to send fulfilled info to users.

hunar1312:
resolves #642


DonKoko (changes requested):
I checked the PR and its mostly good, however I am still having the issue in mac mail that fonts sizes are not correct:
Screenshot 2024-01-10 at 16 15 32

app/modules/invite/service.server.ts
```diff
@@ -154,10 +156,19 @@ export async function createInvite({
   const token = jwt.sign({ id: invite.id }, INVITE_TOKEN_SECRET, {
     expiresIn: `${INVITE_EXPIRY_TTL_DAYS}d`,
   }); //keep only needed data in token to maintain the size
+
   await sendEmail({
     to: inviteeEmail,
     subject: `You have been invited to ${invite.organization.name}`,
-    text: inviteEmailText({ invite, token }),
+    text: "",
```
DonKoko: We should not remove that.  We still need the text email as some people have html emails disabled.

DonKoko (changes requested):
1 small change required.

DonKoko (approved):

PR #680, opened 2024-01-23
Bookings emails

DonKoko:


app/emails/bookings-updates-template.tsx
```diff
@@ -0,0 +1,116 @@
...
+          <p style={{ ...styles.p }}>
+            <span style={{ color: "#101828", fontWeight: "600" }}>
+              Custodian:
+            </span>{" "}
+            {`${booking.custodianUser?.firstName} ${booking.custodianUser?.lastName}` ||
+              booking.custodianTeamMember?.name}
+          </p>
+          <p style={{ ...styles.p }}>
+            <span style={{ color: "#101828", fontWeight: "600" }}>From:</span>{" "}
+            {booking.from.toLocaleString()}
```
DonKoko: That is not the correct way to format the date as it will not always use the correct style and formatting. Please check `baseBookingEmailContent` for an example how we format dates.

app/emails/bookings-updates-template.tsx
```diff
@@ -0,0 +1,116 @@
...
+          >
+            contact the owner
+          </Link>{" "}
+          of the workspace.
+        </Text>
+        <Text
+          style={{ marginBottom: "32px", fontSize: "14px", color: "#344054" }}
+        >
+          {" "}
+          © 2023 Shelf.nu, Meander 901, 6825 MH Arnhem
```
DonKoko: We need to change this address. Its not the correct one.

app/emails/bookings-updates-template.tsx
```diff
@@ -0,0 +1,116 @@
...
+        </Text>
+      </Container>
+    </Html>
+  );
+}
+
+export const bookingUpdatesTemplateString = ({
```
DonKoko, line 135: I am not sure whats the purpose here. We are just rendering the other component and passing the same data.

DonKoko (commented):
Some more comments:
- [ ] Logo should be centered
- [ ] There should be a little bit more space between the logo and h1
- [ ] Overdue booking is missing the design
- [ ] We need to remove some info from the view:
Screenshot 2024-01-23 at 18 31 52

app/emails/bookings-updates-template.tsx
hunar1312: This one is from the designs, Can you provide me the correct one?

app/emails/bookings-updates-template.tsx
hunar1312: We cannot use a TSX component in TS file. Importing it in service module and converting into renderable email html string (which won't work), exporting the converted HTML string.

DonKoko:
Screenshot 2024-02-05 at 17 18 12
This is kinda cracked. Text needs fixing.

Issue #681, opened 2024-01-24
Bookings remaining feedback


DonKoko:
This is the remaining feedback from the bookings that was on the PR. Moved it to its own issue.

- [x] When configuring a booking and clicking save, additional Draft bookings get created. Sometimes 1, sometimes 2 additional ones that shouldn't have been created
https://github.com/Shelf-nu/shelf.nu/assets/76112272/1927176b-6175-46c7-9e46-9516e8c0fd01

Sorry to suggest this late in testing but...

- [ ] Can we send user back to booking index after performing actions like Reserve, Check-out, Check-in, Delete and Archive. I think in 90% of the cases they'll prefer this over staying on the individual booking page. We'll just have to make sure the notification works consistently informing them the action was successful. In my experience the notifications consistently  
 
- [ ] We need a ‘Check-in reminder’ mail when a booking gets checked out after the End date / time. Currently the mail is quite weird when this happens. This example is when check-out happened 44 minutes after End date / time. “Your booking was due for check-in 44 minutes ago”. Not a real case but it can happen.

![Screenshot 2024-01-23 at 15 07 57](https://github.com/Shelf-nu/shelf.nu/assets/76112272/1f0e480a-64d5-43cd-a9f9-dc0ca86826e5)

- [ ]  If we remove the left and right padding for the elements containing the fields for Name, Booking period and Custodian we can save some horizontal screen real estate

![Screenshot 2024-01-23 at 15 08 42](https://github.com/Shelf-nu/shelf.nu/assets/76112272/bff06415-a3e7-438a-ab29-a291151b51cd)

- [ ] when configuring booking we're using the 24-hour and DD/MM/YYYY time / date format. In the booking index we're using the 12-hour and MM/DD/YY time / date format. Maybe not for now, as we have not yet implemented the settings for this. There was talk about potentially syncing it with device settings.

- [ ] If we want to follow designs the 'checked out' warning label should say 'Unavailable' instead. Also the 'checked out' label is only showing on assets in the Manage assets modal and not on the assets in the index within the booking. The logic of when the 'checked out' AKA 'Unavailable' label should show is described in the annotations under design.

https://github.com/Shelf-nu/shelf.nu/assets/76112272/c98c9800-ec9d-4c43-8acb-b21ecab0de8c

_Originally posted by @jurrejansen in https://github.com/Shelf-nu/shelf.nu/issues/555#issuecomment-1906124042_

carlosvirreira:
[...]

PR #713, opened 2024-02-02
UI updates

hunar1312:
resolves UI feedback mentioned in #681


DonKoko (approved):
@hunar1312 I made some changes to fix some spacing consistencies. Overall its good, but i think something in this PR broke the user dropdown:

PR #726, opened 2024-02-06
improvement: webhook automation for free trial

DonKoko:
Currently when a user wants free trial, we just give it via the stripe dashboard. 
However because we don't have hooks to manage free trials, we don't automatically update the DB, which can be a mess as it has to be managed manually now.
The goal of this PR is to automate this, so the sales team can give free subscriptions easily and not worry about cancelling them.

To do:
- [x] Handle a trial subscription being added to a user
- [x] Send email when trial subscription is about to expire
- [x] Handle trial subscription expiring - de-activating it in our db
- [x] Display free trial information on dashboard

### DIFF
--- a/app/components/subscription/current-plan-details.tsx
+++ b/app/components/subscription/current-plan-details.tsx
@@ -1,5 +1,6 @@
-import { useLoaderData } from "@remix-run/react";
+import { Link, useLoaderData } from "@remix-run/react";
 import type { loader } from "~/routes/_layout+/settings.subscription";
+import { Button } from "../shared";
 
 export const CurrentPlanDetails = () => {
   const { activeProduct, expiration, subscription, isTrialSubscription } =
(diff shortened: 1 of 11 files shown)


DonKoko:
@carlosvirreira @jurrejansen This is currently deploying. Can you please help me test it on testapp? https://github.com/Shelf-nu/shelf.nu/actions/runs/7802039358

You should test the 4 points above. Maybe I missed something, let me know if thats the case.

How to test? 

1. Create a new clean user
2. Give it a subscription in stripe (test mode) and select a trial period(any length would work)
3. Once you are ready to test the expiration, go to the customer in stripe and use the "Current time simulation" to test the pre-expiry email and to make sure subscription gets disabled in the app.

DonKoko:
> Even if a user is on a Trial Team Plan - it is not allowed to create a team workspace (You cannot create a workspace on a Free subscription. Please consider upgrading to team plan)
> Request: You’re currently using the Team version of Shelf on a free trial. Your free trial expires on 3/8/2024 at 3:51:31 PM. If we could make the word 'free trial' lead to this page: https://www.shelf.nu/knowledge-base/free-trial | It would be great.

@carlosvirreira those are implemented and deploying now: https://github.com/Shelf-nu/shelf.nu/actions/runs/7815182072

PR #728, opened 2024-02-07
create customer on onboarding

DonKoko:


carlosvirreira:
- Even if a user is on a Trial Team Plan - it is not allowed to create a team workspace (You cannot create a workspace on a Free subscription. Please consider upgrading to team plan)
- Request: You’re currently using the Team version of Shelf on a free trial. Your free trial expires on 3/8/2024 at 3:51:31 PM. If we could make the word 'free trial' lead to this page: https://www.shelf.nu/knowledge-base/free-trial  | It would be great.

PR #858, opened 2024-03-19
Adding Environment Variables for Seed Script

jugrajsingh:
Adding a functionality that allows for overriding the email username and password in the seed script


DonKoko (changes requested):
This is good, however it requires also an update of the docs. Can you please update the information about the env variables in docs/getting-started so it reflects those new ones?

jugrajsingh:
> This is good, however it requires also an update of the docs. Can you please update the information about the env variables in docs/getting-started so it reflects those new ones?

Updated the Getting Started Doc with the new Environment Variables

app/database/seed.server.ts
```diff
@@ -90,9 +90,11 @@ const supabaseAdmin = createClient(SUPABASE_URL, SUPABASE_SERVICE_ROLE, {
 
 const prisma = new PrismaClient();
 
-const email = "hello@supabase.com";
+const email = process.env.ADMIN_EMAIL || "hello@supabase.com";
```
DonKoko: Something else I noticed here. This works but its not fitting with the way we usually do it in shelf. Please take a look at the file `env.ts` and how we manage and export/import env variables. This needs to be done for all 3 new variables.

DonKoko (changes requested):
Something I noticed after further review that needs to be adjusted to fit the style of dev of shelf

DonKoko:
hey @jugrajsingh. I requested some changes a few days back. Can you let me know if its possible for you to do them so I know how to proceed with this PR.

app/database/seed.server.ts
jugrajsingh: Will work on it in few hours

jugrajsingh: @DonKoko I think the new Changes should follow the practice.

DonKoko: Thanks @jugrajsingh. Thats perfect. I am merging the PR.

PR #840, opened 2024-03-12
rphlmr/fixing error handling

rphlmr:
Forks and continue https://github.com/Shelf-nu/shelf.nu/pull/541

# Patch note (partial)

- `loader` and `action` now use `parseData` helper to validate the request formData/SearchParams/Params and enforce typing + error handling.
- `loader` and `action` now use `json(data(...))` and `json(error(...))`
    - Throwable code have to be surrounded by a `try`/`catch` block to properly handle service errors or unhandled errors.
    - Don't throw anything else than `ShelfError` (like `redirect`) or it will not work as expected and it will be captured in the `catch` bloc.
    - In `catch` bloc: 
        - `loader` should throws `json(error(...))` (or return if it is a resource route)
        - `action` should return `json(error(...))` to access `actionData.error` in `useActionData` or `fetcher.data`.
- `tryCreateUser` has been removed because wrong usage would result in complete user account deletion.
- `requireOrganisationId` renamed to `getSelectedOrganisation` to better reflect what it does
- `getSelectedOrganisation` no longer throws `redirect` or anything related to the HTTP layer (`json`).
   - This fix wrong error handling in loader/action or other utils that use it.
   - Default `organizationId` is the `PERSONAL` org.
- `error` helper (`utils/http.server.ts`) send a notification of the error it catch.
    It should help catch unhandled action error responses.
- `_layout+` no longer have to handle null `user` since this case is no longer possible. 
- add a `SMTP_FROM` env to use in `nodemailer` because secure SMTP servers reject unknown sender emails
- `settings/team/invite-user` now properly handles its `loader` error and falls back to `settings/team` `ErrorBoundary`

Progress:

### Routes:

- [x] ./api+/$organizationId.qr-codes[.zip].ts
- [x] ./api+/admin.export-org-assets.$organizationId.$fileName[.csv].tsx
- [x] ./api+/admin.import-org-assets.$organizationId.tsx
- [x] ./api+/asset.refresh-main-image.ts
- [x] ./api+/client-notification.ts
- [x] ./api+/image.$imageId.ts
- [x] ./api+/model-filters.ts
- [x] ./api+/oss-friends.ts
- [x] ./api+/public-stats.ts
- [x] ./api+/sse.notification.ts
- [x] ./api+/stripe-webhook.ts
- [x] ./api+/user.change-current-organization.ts
- [x] ./api+/user.prefs.dismiss-support-banner.ts
- [x] ./api+/user.prefs.minimized-sidebar.ts
[...]

PR #859, opened 2024-03-19
Adding Environment Variables to Override From Address

jugrajsingh:
Adding Environment Variables to Override From Address
This is required for Custom SMTP Implementation


DonKoko:
hey @jugrajsingh. Thanks for the PR. We already have another big PR open that is solving this. It will be merged either today or tomorrow, so I will close this PR to prevent conflicts. The fix is in #840

Issue #845, opened 2024-03-16
[Bug]: Documentation missing for running database migration in Docker


DASPRiD:
### What happened?

Steps to reproduce:

1. Install via documented Docker method
2. Try to log in
3. Sentry reports that tables are missing, so apparently the schema never got applied

### What is the expected behaviour?

Either the documentation should state on how to apply the database schema, or the start.sh script should run the migrations.

I tried to run `npm run setup:db` in the container, but that fails because there are no schema files in the container.

### Version

Self-hosted

### What browsers are you seeing the problem on?

Firefox

### Relevant log output

```shell
PrismaClientKnownRequestError: 
Invalid `prisma.organization.findFirstOrThrow()` invocation:
The table `public.Organization` does not exist in the current database.
```

anatolinicolae:
I think I've seen the same behaviour on my side. I'll get back with a fix for this 👍

casef007:
Any new information on that? I am facing the same issue.

anatolinicolae:
#866 is merged. @DASPRiD & @casef007 give the latest version a try, it should automatically take care of migrations 👍

DonKoko:
Thanks @anatolinicolae. I am closing this issue for now. 
@DASPRiD @casef007 let us know if there is further feedback about this.

capriciousduck:
Can I not run my own Postgres database and use it with my shelf container? Is supabase strictly necessary?

DonKoko:
hey @capriciousduck . Supabase is not only used for DB but also for auth. At this moment it is required to use supabase with shelf. If you choose to, you can always self host supabase via their docker image.

PR #866, opened 2024-03-20
fix: Docker image migrations

anatolinicolae:
## Issue
Base startup script doesn't include DB migrations, therefore app may be unable to start or will error when database is not synchronized.

Related issue: #845

## In this PR
- [x] App database is now packaged in the image for proper migrations
- [x] Docker entrypoint is decoupled from normal start.sh
- [x] Docker entrypoint now generates the schema and runs migrations before app startup
- [x] Dockerfile underwent a little cleanup with the unification of WORKDIR directive in the base step

## Update steps
Recreating your previous container with a freshly-pulled image should suffice (make sure to `pull` first).

## New startup behaviour

```
+ npm run setup:db

> setup:db
> prisma generate && prisma migrate deploy

Prisma schema loaded from app/database/schema.prisma

✔ Generated Prisma Client (v5.11.0) to ./node_modules/@prisma/client in 141ms

Start using Prisma Client in Node.js (See: https://pris.ly/d/client)

import { PrismaClient } from '@prisma/client'
const prisma = new PrismaClient()

or start using Prisma Client at the edge (See: https://pris.ly/d/accelerate)

import { PrismaClient } from '@prisma/client/edge'
const prisma = new PrismaClient()


See other ways of importing Prisma Client: http://pris.ly/d/importing-client

┌─────────────────────────────────────────────────────────────┐
│  Deploying your app to serverless or edge functions?        │
│  Try Prisma Accelerate for connection pooling and caching.  │
│  https://pris.ly/cli/accelerate                             │
└─────────────────────────────────────────────────────────────┘

Prisma schema loaded from app/database/schema.prisma
Datasource "db": PostgreSQL database "postgres", schema "public" at "aws-0-eu-central-1.pooler.supabase.com:5432"

91 migrations found in prisma/migrations

Applying migration `20240124105759_add_aed_currency`

The following migration(s) have been applied:

migrations/
  └─ 20240124105759_add_aed_currency/
    └─ migration.sql
      
All migrations have been successfully applied.
+ npm run start

> start
> NODE_ENV=production node build/index.js

🚀 Server started on port 8080
Scheduler and workers registration completed
```


DonKoko (changes requested):
Okey so I pulled this and manually merged it with our dev branch, so I can test our CI/CD on fly.io.
Unfortunately this change, doesn't seem to work for fly. If I deploy with this new dockerfile our app completely breaks. 
When you try to login you get the following error:
Screenshot 2024-03-21 at 13 24 47

I am not sure but I think its something with that new docker file overriding some types or something, because the type `OrganizationRoles` exists in the DB. I double checked that. 
Also running the app locally(no docker) and connecting to the staging DB, also works without issue. So this tells me its not an actual issue with the DB.

Maybe fly.io has some limitations related to what we are trying to do? Is it possible to have different docker files to be used when deploying to fly and when using for self-hosting?

anatolinicolae:
@DonKoko great catch! Possibly all fixed now...

Went ahead and further split Dockerfiles for fly.io and image builds so they're independent now. Reworked the Dockerfiles themselves and the steps, which I included for fly.io as well—give that a try and if it still breaks, I'll roll that back.

Upgraded to `--include` and `--omit` as  `--production` seems to be deprecated.

Tried deploying the container on my side and everything seems to be working as expected. Can you give another try on staging?

package.json
```diff
@@ -8,6 +8,7 @@
     "db:diff": "prisma migrate diff --from-url=${DATABASE_URL} --to-migrations=./app/database/migrations --shadow-database-url=${SHADOW_DATABASE_URL} --script > rollback.sql",
     "db:prepare-migration": "prisma generate && prisma migrate dev --create-only --skip-seed",
     "db:generate-type": "prisma generate",
+    "db:deploy": "prisma migrate deploy",
```
DonKoko, line 11: Cant use use `db:deploy-migration` for this?

DonKoko (changes requested):
Okey so I deployed this to our staging app and it works fine. No issues faced.
1. The file named `Dockerfile.image`. The part after the dot, is that following some requirement from docker or it can be anything? If it can be anything i rather do it more descriptive. Something like `Dockerfile.selfhost`

2. I think we need to add a bit of explanation of this to the docs, so people understand whats the purpose of each docker file and know where to touch if they are making adjustments/changes.

package.json
anatolinicolae: I would like to avoid regenerating the client at runtime, keeping that to the build process only. Running `db:deploy-migration` would introduce some drift between what the image is supposed to look like and it's current state, basically "we ship a thing, you run another."

However, totally doable if you prefer using that 👍

DonKoko: No thats fine. Just needed to understand what is going on.

DonKoko (approved):

null-dev:
Hey @anatolinicolae, thanks for packaging this a `Docker` container!

BTW, [pruning the non-dev dependencies](https://github.com/Shelf-nu/shelf.nu/blob/f0b1bfc303cb73568f0e16a0d0ea04b0523a16ae/Dockerfile#L30) breaks [seeding the DB](https://github.com/Shelf-nu/shelf.nu/blob/main/docs/docker.md#instructions) (`docker exec -it shelf npm run setup:seed`), prisma [relies on `tsx`](https://github.com/Shelf-nu/shelf.nu/blob/f0b1bfc303cb73568f0e16a0d0ea04b0523a16ae/package.json#L162) to seed the DB, and it's a [dev dependency](https://github.com/Shelf-nu/shelf.nu/blob/f0b1bfc303cb73568f0e16a0d0ea04b0523a16ae/package.json#L152).

anatolinicolae:
Good catch @null-dev! Mind opening a separate issue for that? I'll prepare a PR later this week to get it sorted 👍

PR #886, opened 2024-03-27
Fixing sentry errors

DonKoko:
Fixing some errors from sentry.

### DIFF
--- a/app/modules/user/service.server.ts
+++ b/app/modules/user/service.server.ts
@@ -217,6 +217,8 @@ export async function createUser(
           organizationIds.push(organizationId);
         }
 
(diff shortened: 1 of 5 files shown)


app/modules/user/service.server.ts
```diff
@@ -217,6 +217,8 @@ export async function createUser(
           organizationIds.push(organizationId);
         }
 
+        // @TODO this is weird and organizationIds is not used
```
DonKoko, line 220: Hey @mahendraHegde. We are having some bugs with the invitations and some other registration cases. I found this code in `createUser` that i found kinda confusing. Could you check the comment and try to remember what was the reason behind this? I want to improve it to iron out some bugs.

PR #892, opened 2024-04-03
testing feature that allows users to add a specific booking to their …

DonKoko:
…calendar


jurrejansen:
- [ ] Suggestion: only show calendar button when booking reaches Reserved state
- [ ] After downloading an updated calendar file with new info (e.g. End date/time) and clicking the file to add to calendar, it doesn't update the old entry

https://github.com/Shelf-nu/shelf.nu/assets/76112272/be15485d-5670-419e-ace7-5158073138a8

- [ ] Suggestion: think it's better to add this function in the Actions dropdown
Screenshot 2024-04-04 at 13 02 11

PR #902, opened 2024-04-08
Improvement to subscription display

DonKoko:


jurrejansen:
- [ ] After being sent to Stripe and activating trial user gets sent back to Settings > Subscriptions. The page is not refreshed and still says user is on Free. It would cause less confusion if the page would already refreshed and says user is now on Team subscription
- [ ] In the scenario I described it would be nice if the user is sent to Settings > Workspaces after activating the trial in Stripe environment. Also the 'Create workspace' modal should be triggered so they instantly get the chance to create their Team workspace.
- [ ] In the scenario I had put some a couple of multistep modals explaining the trial behaviour before activating.
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=3588-7557&mode=design

DonKoko:
> After being sent to Stripe and activating trial user gets sent back to Settings > Subscriptions. The page is not refreshed and still says user is on Free. It would cause less confusion if the page would already refreshed and says user is now on Team subscription

Seems like a bug that happened to you. That shouldn't happen. The experience should be the exact same as updating to a team subscription(for now). I just did a test and saw this:
Screenshot 2024-04-12 at 10 41 07


> In the scenario I described it would be nice if the user is sent to Settings > Workspaces after activating the trial in Stripe environment. Also the 'Create workspace' modal should be triggered so they instantly get the chance to create their Team workspace.

I think the UX should be the same as updating to team and indeed I agree we should improve it. I would say the modal we have should have the CTA "Create your team workspace" that directly sends you to workspace/new.

>  In the scenario I had put some a couple of multi-step modals explaining the trial behavior before activating.

Yeah I don't really like that tbh. It seems like extra steps that I don't see the value of. Would require like 4 extra clicks. If we are concerned whether they have the correct info, we can always send them an email when they start a trial, with some info. That way they also have a trace of it.

carlosvirreira:
I have to agree with @DonKoko on the steps. 

> I would say the modal we have should have the CTA "Create your team workspace" that directly sends you to workspace/new.

Very much agree

> Would much rather optimize for speed of enabling the trial as they are already having an account on shelf. We can always send additional info per email. (What you got into, what you can expect, etc).

carlosvirreira:
> @carlosvirreira I made some more updated. I made sure that you can get a free trial just once, because stripe doesnt control that so we have to check it ourselves. I am deploying it now to the testapp: https://github.com/Shelf-nu/shelf.nu/actions/runs/8739097019
> 
> Can you please give it a test and lmk if its good and what improvements we can do. I also updated the CTA so it shows "Create team workspace" if u got a team plan or a team trial.

Requesting testing backup here @jurrejansen

carlosvirreira:
Scenario 1: I request for a trial - and immediately cancel the plan. What happens? 
Behavior: I am allowed to cancel it. I get this "You’re currently using the Team version of Shelf on a free trial. Your plan has been canceled and will be active until 5/7/2024 at 10:25:49 AM. You can renew it at any time by going to the customer portal." I can resume it as my subscription is "Paused". Everything went well here.

Scenario 2: I request for a trial - I reach the end date. I end up cancelling my plan.
Behavior: I can add my payment method information but cannot find a way to create a new subscription for my workspace. I basically have to contact support (Would not spend time fixing this path) but do want to mention it.


I'd say we can release this.

PR #908, opened 2024-04-14
Treat asset description text as pre-formatted

null-dev:
Display asset description text with `whitespace-pre-wrap` on asset page. This displays newlines in the asset description while still maintaining text wrapping.

Feel free to close if you have a reason why this shouldn't be the case though!

Old behavior:
Screenshot_20240413_221546

New behavior:
Screenshot_20240413_221635


DonKoko:
hey @null-dev. Thanks for your PR and contribution. 
That's an awesome update. I honestly didn't even know about this css property.

PR #915, opened 2024-04-17
Signing templates - rajdip

rajdip-b:


DonKoko:
hey @rajdip-b . I updated you branch and fixed all the conflicts. I added a bunch of @TODOs and answered your questions in the code. Still quite some stuff to updated but they are all basically the same. I tried to add @TODO everywhere but it might be that I missed something. Here are the general points:

- update all service functions to properly catch errors. See any service.server file for example
- You are not using parseData to get form data
- You are not using getParams to get params

carlosvirreira:
Excited for this release!

PR #931, opened 2024-04-26
feat: Fix icons used in menu items for dropdown menu's

### LINKED ISSUE #925: [Improvement request]: Fix icons used in menu items for dropdown menu's
### Contact Details

jurre@shelf.nu

We're not consistently using a container for the icons used in menu items which results in the icons looking off. It's not something that is very high impact but it doesn't look very professional.

EXAMPLE:

Screenshot 2024-04-23 at 17 38 54
Screenshot 2024-04-23 at 17 41 56

Identify all places within the app where these dropdown menus are used and giving the icons a consistent container. Maybe there is way to create a global component for it so from now on it can stay consistent.

_No response_

### Additional context

_No response_

Bhavyajain21:
/fixes #925 
/claim #925 

image


DonKoko (changes requested):
Some changes required based on our discussion on discord.

app/components/assets/actions-dropdown.tsx
```diff
@@ -86,7 +87,8 @@ const ConditionalActionsDropdown = () => {
                 [data-radix-popper-content-wrapper] {
                   transform: none !important;
                   will-change: auto !important;
-              }
+              }import iconsMap from '../shared/icons-map';
```
DonKoko: Not sure what happened here. This shouldn't be there.

app/components/icons/WrapIcon.tsx
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import IconHug from "./IconHug";
+import type { Icon } from "../shared/icons-map";
+import iconsMap from "../shared/icons-map";
+
+export interface WrapIconProps {
+  icon?: Icon;
+  enableWrap?: boolean;
+}
+const WrapIcon = React.forwardRef<HTMLElement, WrapIconProps>(
```
DonKoko: Rename the component to simply `Icon`

app/components/icons/WrapIcon.tsx
```diff
@@ -0,0 +1,16 @@
...
+  function WrapIcon({ icon, enableWrap }: WrapIconProps) {
```
DonKoko: `enableWrap` should be `true` by default so we dont have to pass it every time, only when we want it to be false as that will be the more rare case.

app/components/assets/actions-dropdown.tsx
Bhavyajain21: Resolved, I think it got added by mistake!

app/components/icons/WrapIcon.tsx
Bhavyajain21: Replaced Icon as IconType, as we were seeing some errors while import the component with the same name as that of Type

app/components/icons/WrapIcon.tsx
Bhavyajain21: Added a disableWrap prop, which needs to be passed when wrapping is not required

Bhavyajain21:
Made the changes accordingly!

DonKoko (approved):

PR #932, opened 2024-04-26
Unclaimed codes

DonKoko:
- made it so userId and organizationId are not required for qr codes
- updated some services that needed change due to different fields
- created new services for generating unclaimed codes
- created admin dashboard index for qrs
- created endpoint to generate and print unclaimed codes
- created a util that takes an array of Prisma.Qr and generates a zipBlob with svgs of those codes
- changed batch from a field within Qr to its own model with Relation to Qr
- updated qr codes index for admin to have better data
- added filter based on batch name/id
- added a form that allows to mark a batch as printed

### DIFF
--- a/app/components/admin/generate-batch-qr.tsx
+++ b/app/components/admin/generate-batch-qr.tsx
@@ -1,0 +1,64 @@
+import { useState } from "react";
+import { useNavigation } from "@remix-run/react";
+import { isFormProcessing } from "~/utils/form";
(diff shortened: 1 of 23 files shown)


DonKoko:
@jurrejansen issue is fixed. Deploying now: https://github.com/Shelf-nu/shelf.nu/actions/runs/8893042370

I also improved the pagination component cuz i found another bug and improved the code to make it smaller and smoother.

PR #937, opened 2024-04-27
feat: Calender View

### LINKED ISSUE #564: [Feature request]: Calendar view of booking index
### Contact Details

jurre@shelf.nu

This is something requested by a user. The booking index is sorted by starting date / time, however to get the best overview a calendar view would be best.

Screenshot 2024-03-21 at 17 35 51

A simplified implementation of a calendar view showing booking entries. Starting out with a monthly overview exclusively.

WIP SCENARIO
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=4434-4519&mode=design

_No response_

### Additional context

_No response_

Bhavyajain21:
/fixes #564 
/claim #564 

Current WIP:

Screenshot 2024-05-01 at 12 51 01 PM


app/root.tsx
```diff
@@ -84,6 +84,7 @@ function Document({ children, title }: PropsWithChildren<{ title?: string }>) {
 
         <Meta />
         {title ? <title>{title}</title> : null}
+        <style data-fullcalendar />
```
DonKoko: Not sure what this style tag is? Remix has its own way of adding styles, can we not use that?

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,47 @@
...
+    title: "Calendar", // Change to an appropriate title
+  };
+
+  return json({ header });
+};
+const Calendar = () => {
+  const { header } = useLoaderData<typeof loader>();
+  return (
+    <>
+      <header className={tw("-mx-4 bg-white")}>
```
DonKoko: Just use <Header/> component and pass it title and subtitle. That will render it in our normal styles.

app/styles/global.css
```diff
@@ -243,3 +243,11 @@ dialog {
 .word-break {
   word-break: break-word;
 }
+.fc-toolbar {
```
DonKoko: This is fine, but I would rather we make a css file specifically for the calendar and only load it on the calendar route. Every byte we save is valuable.

vite.config.ts
```diff
@@ -21,6 +21,9 @@ export default defineConfig({
         "./app/routes/**/*",
       ],
     },
+    hmr: {
```
DonKoko: Not sure what this is. Can you give me some explanation

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,162 @@
...
+        initialView="dayGridMonth"
+        timeZone="local"
+        events={calendarEvents}
+        datesSet={handleMonthChange}
+      />
+    </>
+  );
+};
+
+const calendar = () => <Calendar />;
```
DonKoko: Not sure why you do it like this. You can just directly export default Calendar

DonKoko (changes requested):
Hey @Bhavyajain21 . I have checked your code as well as pulled the branch to look at the designs. So here is some general feedback about the code that I also wrote you on discord:

1. **Loader** - please take a look at any of our other loaders and read the docs about handling responses and errors. You need to follow our standard for writing code in loaders. Moreover I see you have some filters and some dates, those look good but will require some changes when we make the DB queries and query the data.
2. **Component** - i saw the `useEffect` thing with getting the bookings. You don't need to do any of that in remix as it has automatic revalidation. So basically what you need to do is move this code to the loader and from the loader directly return the events data, so the loader will directly return `CalendarEvent[]`.
3. **Handle month change** - that is mostly good but you shouldn't use `navigate`. Please look at the hook called `useSearchParams()`. It gets the params and also allows you to update them. The way remix works, is that when you use `setSearchParams`, it automatically causes a navigation and will "reload" the page which means it triggers the loader and returns to you updated data. That's also why u don't ever need to `useEffect` and rarely you need local state go manage the data.

Moreover looking at the design here are some feedback points that need to be solved:

- [ ] page is still missing the header. The designs are a bit different than our current headers as they are not updated globally, but please just use the <Header/> component and that will give you the effect you need.
- [ ] Page is still missing meta title. Please check other pages for example
- [ ] Calendar event styling - currently they all have the same color. In the design its a bit different. @jurrejansen can you give us an explanation about the logic behind the colors and provide what colors we should use for BG and Border
- [ ] The text for each event on the calendar should have the following structure: `starting time | booking name | custodian full name`
- [ ] I still see the buttons for switching months different than the design. Are there some limitations around this? Let me know if I can help with that somehow.

app/routes/_layout+/calendar.tsx
Bhavyajain21: yes, we can do that as well. Made the changes accordingly!

app/root.tsx
Bhavyajain21: This can be removed. Thanks for pointing out!

app/routes/_layout+/calendar.tsx
Bhavyajain21: Made the changes, used the already existing Header component.

vite.config.ts
Bhavyajain21: When an error occurs, Vite typically displays an overlay in the browser to inform you of the error. This was added to remove the error coming up on the browser while local development. This can be removed now!

app/styles/global.css
Bhavyajain21: True that! Have created a separate file for calendar styles.

Bhavyajain21:
Hey @DonKoko , I have updated the code based on the suggestions provided:

**_for loaders:_** To align with the code practices, have updated the code with the revised version of the loader with appropriate error handling and response structures. The code still uses dummy data for now, it can be replaced by getting the data through the queries.

**_Component:_** Thanks for pointing that out! I noticed that the use of useEffect to fetch bookings wasn't necessary in Remix due to its automatic revalidation. To improve efficiency and simplify the component, I moved the data-fetching logic to the loader, which now directly returns the CalendarEvent[] data.

**_Handle month change:_** Have replaced it with useSearchParams, this is a better way when we want to manipulate query parameters to update the state, trigger revalidation, or allow users to interact with a specific part of the URL.

Bhavyajain21:
For the Design Changes:

image

This is how it looks currently. Have made the changes accordingly based on the above comments made.

Have a slight different header for now, please let me know if that is what the expectation is, or should we remove the home icon?

Also, would like to know if we have to give different colors based on the events. Please provide the feedback on the design.

carlosvirreira:
I just want to share some thoughts about:

**Calendar event styling - currently they all have the same color. In the design its a bit different. @jurrejansen can you give us an explanation about the logic behind the colors and provide what colors we should use for BG and Border**

I think that launching with a flat color logic (all events have the same color) as being the best option. Just to avoid complexity and ship to then iterate/get user feedback.

The original idea is (which is still good - but it could cause a slowdown):

```Booking are represented by the colored badges, commonly used in calendar views. The color of the badge indicates the current status of booking. In this particular designs you can see a **Completed**, **Ongoing** and **Reserved** booking. Other states can be **Overdue**, **Archived** and **Cancelled**.```


Screenshot 2024-05-01 at 09 42 58

**Review by DonKoko — CHANGES_REQUESTED** [review]
Something i noticed about your css file. You should instead of styling with default styles use @apply and tailwind classes, like we do it in other places. That way colors and spacing and everything will be consistent.

**Comment by Bhavyajain21** [comment]
> Something i noticed about your css file. You should instead of styling with default styles use @apply and tailwind classes, like we do it in other places. That way colors and spacing and everything will be consistent.

Made the changes

File: app/styles/layout/calendar.css [hunk 6]
```diff
@@ -0,0 +1,72 @@
...
+  @apply ml-1 !important;
+}
+
+.fc-next-button {
+  @apply ml-0 rounded-tl-none rounded-bl-none border border-gray-400 bg-white text-gray-600 !important;
+}
+
+.fc-today-button {
+  @apply ml-0 rounded-none border border-gray-400 bg-white text-black pt-[0.5em] !important;
+  border-left-width: 0 !important;
```
DonKoko: Instead of doing this you can use border-x or border-y to only add on the axis you want

app/styles/layout/calendar.css
```diff
@@ -0,0 +1,72 @@
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
DonKoko: I don't think you need to import that. We don't import it in other files. That is managed by remix vite plug-in

DonKoko (changes requested):
Some small things I noticed.

app/styles/layout/calendar.css
Bhavyajain21: Made the changes!

app/styles/layout/calendar.css
Bhavyajain21: Resolved!

Bhavyajain21:
https://github.com/Shelf-nu/shelf.nu/assets/64700692/9f32e4cb-64f1-45bc-9fd3-ff0962204b7a



Demo Video With Loader

Bhavyajain21:
https://github.com/Shelf-nu/shelf.nu/assets/64700692/f04e20e5-c823-4f0b-bfa2-37bf2abefe1a



Adding Calendar events with spinning loader.

CC: @carlosvirreira @DonKoko

carlosvirreira:
Looks very nice - can't wait to test it!

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,180 @@
...
+        } else {
+          ripple.current.classList.add("hidden");
+        }
+      }
+    },
+    [ripple]
+  );
+
+  useEffect(() => {
+    updateTitle();
```
DonKoko: This is not needed. You can get the default title from the loader, for the first load.

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <Button
+                variant="secondary"
+                className="border-r px-3 py-2 text-sm font-semibold text-[#344054]"
+                onClick={() => handleNavigation("today")}
+              >
+                Today
+              </Button>
+              <Button
+                variant="secondary"
+                className="p-[0.75em] text-[#667085]"
```
DonKoko: This should be `text-gray-500`

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <Button
+                variant="secondary"
+                className="border-r p-[0.75em] text-[#667085]"
+                onClick={() => handleNavigation("prev")}
+              >
+                <ChevronLeftIcon />
+              </Button>
+              <Button
+                variant="secondary"
+                className="border-r px-3 py-2 text-sm font-semibold text-[#344054]"
```
DonKoko: This should be text-gray-700. In figma you can actually see the values.

DonKoko (changes requested):
- [ ] Colors on the events still dont look good to me. They don't look like the design
- [ ] Something is wrong with the sizing of the date boxes. For some reason when there are events it adds lots of white space below. 
Screenshot 2024-05-07 at 14 44 10
- [x] Clicking on a booking should lead to booking page
- [ ] Text color of draft events looks kinda off
- [ ] Styling of current day is not like in the design
- [ ] We still need to make sure that self-service can only see bookings they created themselves

These are some stuff I noticed. For the rest it looks fine.

DonKoko:
One more thing to add that I noticed when deployed to testapp. This is how it looks when you have just 1 day events.
Screenshot 2024-05-07 at 15 10 54
For sure all the empty space has to be solved like i pointed out in the review, but also, I am not sure how we should display the events themselves. I like how apple calendar does it:
Screenshot 2024-05-07 at 15 12 43

However then we have to figure out how the interaction works/looks like.

carlosvirreira:
Additional Feedback (Nikolay's was spot on):
- [ ] Colors and statuses seem to not be matching
- [ ] One day events seem to be bold - perhaps this is expected as they are happening today - but I would like to get it looked at
- [ ] Calendar menu item let us move it please above bookings - they are related conceptually to it makes sense to put them close to each other
- [ ] Mobile does not look broken - for sure something to take a look at in the future (comments/feedback) I think if clicking on a booking takes you there then it is already valuable as a quick link

jurrejansen:
Feedback that hasn't been mentioned:

- [ ] Calendar component is missing a white background color
- [ ] Would be good that if you hover over an event that stretches out over multiple rows the event gets the hover state on both rows

https://github.com/Shelf-nu/shelf.nu/assets/76112272/20956567-cc02-4416-90ae-ce1a10f34e7c

- [ ] I would prefer that single day booking events appear the same as events that stretch over multiple days
Screenshot 2024-05-07 at 14 51 02

- [ ] I would like to suggest to remove the date from the events itself to leave space for information that is not clear by just looking at the calendar overview
Screenshot 2024-05-07 at 14 52 46

- [ ] Color of event badges should correspond with the status of the asset. I've put the badges in under the annotation in scenario. Colors can be fetched there but let me know if more input is needed
Screenshot 2024-05-07 at 15 10 59

app/routes/_layout+/calendar.tsx
Bhavyajain21: Fixed!

app/routes/_layout+/calendar.tsx
Bhavyajain21: fixed

app/routes/_layout+/calendar.tsx
Bhavyajain21: fixed

app/hooks/use-main-menu-items.tsx
```diff
@@ -29,6 +29,11 @@ export function useMainMenuItems() {
       to: "locations",
       label: "Locations",
     },
+    {
+      icon: <Icon icon="calendar" />,
+      to: "calendar",
+      label: "Calendar",
```
DonKoko, line 35: We should use different icons for the calendar and bookings. Can you make it look like the icons on the figma file

DonKoko (changes requested):
A lot of points from the previous reviews are still pending.

app/routes/_layout+/calendar.tsx
```diff
@@ -110,6 +110,55 @@ const Calendar = () => {
...
+  type BookingStatus =
+    | "DRAFT"
+    | "ARCHIVED"
+    | "CANCELLED"
+    | "RESERVED"
+    | "ONGOING"
+    | "OVERDUE"
+    | "COMPLETE";
+
+  const statusClassesOnHover: Record<BookingStatus, string> = {
```
DonKoko: You should make this work with tailwind classNames rather than with hardcoded colors. On the hover events you can add and remove classes. If we do it this way once we change the colors of the tailwind theme we would have to update it here and I dont wnat to do that. it should be managed in 1 central place. In that case you also dont need 2 objects, you can have 1 object with classes.

app/routes/_layout+/calendar.tsx
```diff
@@ -110,6 +110,55 @@ const Calendar = () => {
...
+    COMPLETE: "#ECFDF3",
+  };
+
+  const handleEventMouseEnter = (info: any) => {
+    const statusClass: BookingStatus = info.event._def.extendedProps.status;
+    const className = "bookingId-" + info.event._def.extendedProps.id;
+    const elements = document.getElementsByClassName(className);
+    for (let i = 0; i < elements.length; i++) {
+      const element = elements[i] as HTMLElement;
+      element.style.backgroundColor = statusClassesOnHover[statusClass];
```
DonKoko: Based on the comment above, here you can instead add/remove a class name.

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,229 @@
...
+  breadcrumb: () => <Link to="/calendar">Calendar</Link>,
+};
+
+// Loader Function to Return Bookings Data
+export const loader = async ({ request, context }: LoaderFunctionArgs) => {
+  const authSession = context.getSession();
+  const { userId } = authSession;
+
+  try {
+    // @TODO here we have to handle self-service, and make sure they can only see bookings that belong to them
```
DonKoko: PLease remove this comment. This is solved already.

DonKoko (changes requested):
It works good, but there are some small issues with the code that need to be updated.

app/routes/_layout+/calendar.tsx
Bhavyajain21: Done

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,220 @@
...
+    DRAFT: "!bg-gray-100",
+    ARCHIVED: "!bg-gray-100",
+    CANCELLED: "!bg-gray-100",
+    RESERVED: "!bg-blue-100",
+    ONGOING: "!bg-purple-100",
+    OVERDUE: "!bg-warning-100",
+    COMPLETE: "!bg-success-100",
+  };
+
+  const handleEventMouseEnter = (info: any) => {
```
DonKoko: Do not use type `any`. Pretty sure you can get the type of info from FullCalendar types.

app/routes/_layout+/calendar.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+// Calendar Component
+const Calendar = () => {
+  const { title } = useLoaderData<typeof loader>();
+  const [_error, setError] = useState<string | null>(null);
+  const [calendarTitle, setCalendarTitle] = useState(title);
+  const calendarRef = useRef<FullCalendar>(null);
+  const ripple = useRef<HTMLDivElement>(null);
+
+  const handleNavigation = (navigateTo: any) => {
```
DonKoko: Do not use any. You know exactly what the 3 possible options are.

carlosvirreira:
@Bhavyajain21 we want to make enhancements to calendar view. Do you think you can help us release? 

We want to have all these features on our calendar view.
Screenshot 2024-10-15 at 17 51 23

Our current calendar view.
Screenshot 2024-10-15 at 17 51 43

Please let me know if we can discuss and assign to you (bounty ofc).

PR #975, opened 2024-05-15
fixing issue that made the formData already be consumed

DonKoko:


DonKoko:
@jurrejansen was a bug I introduced when we did the big change in how the bookings work. Has been solved.

PR #996, opened 2024-05-25
Feat/revamped asset page

### LINKED ISSUE #972: [Improvement request]: Revamped asset page
### Contact Details

jurre@shelf.nu

The current asset page is outdated, has too little space to accommodate UI elements related to currently existing and/or upcoming features and too focused on the Notes / Activity log.

A new layout that uses tabs in order to categorize the different the information shown on the asset page.
Figma preview link: https://www.figma.com/design/iKu7Cn98DOCHcpBHgPCPCt/Shelf-%7C-UX-Scenarios-(Copy)?m=dev&node-id=107%3A42253&t=K9D3hlXCJuvsk9Z5-1

You can see the preview above. This should be the basic design we have to start with.

Some important things to keep in mind:

- The new design uses tabs. You need to create a route for each tab and we need to adjust the loaders and how data is managed. You can see exactly how to create the different routes and tabs in the settings area of shelf.
- Some specific elements such as the QR area or the Map, might look different than what we have right now. Please ignore them and just use this as a reference for the structure of the page. You should not for example redesign the map component and all its data. It should look as it does now, just positioned differently as based on the new design.
- The activity tab should include the current notes that are on the asset page. No design change, just placed within the activity tab.

Bhavyajain21:
/claim #972 
/fixes #972 

Work done so far:

image


app/components/assets/notes/index.tsx
```diff
@@ -8,7 +12,8 @@ import type { NoteWithDate } from "./note";
 import { Note } from "./note";
 
 export const Notes = () => {
-  const { asset } = useLoaderData<typeof loader>();
+  const { asset } = useOutletContext<any>();
```
DonKoko: This is a fine approach but its better to use `useRouteLoaderData`.
Its better for typing and its more clear where the data comes from when looking at the file.

app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -248,338 +267,40 @@ export const links: LinksFunction = () => [
...
-          ) : (
-            <>
-              <TextualDivider text="Notes" className="mb-8 lg:hidden" />
-              <Notes />
-            </>
-          )}
-        </div>
+      <HorizontalTabs items={items} />
+      <div>
+        <Outlet context={data} />
```
DonKoko: If you change the subroutes to useRouteLoaderData this is not needed anymore I think.

app/components/layout/header/index.tsx
```diff
@@ -23,6 +25,12 @@ export default function Header({
   hidePageDescription?: boolean;
   hideBreadcrumbs?: boolean;
   classNames?: string;
+  asset?: {
```
DonKoko: Okey, so this approach is okey but it feels a bit hardcoded. What I would rather do is adjust the header component that the title is either a string or a react node. So we can just pass some react code. Kinda similar like it works with children in react.

app/components/assets/asset-qr.tsx
```diff
@@ -0,0 +1,173 @@
...
+    };
+  };
+  title: string;
+}
+
+const QrLabel = React.forwardRef<HTMLDivElement, QrLabelProps>((props, ref) => {
+  const { data, title } = props;
+  return (
+    <div
+      className="flex h-auto w-4/5 flex-col justify-center gap-3 rounded-md border-[5px] border-[#E3E4E8] bg-white px-8 py-[17px]"
```
DonKoko: The `w-4/5` as far as I remember, was specifically done because the QR panel had a fixed width so it was calculated based on that. Please make the column of the qr have the same width as the QR sidebar had before.

DonKoko (changes requested):
I added some comments and pushed a small commit with some tiny css stuff I noticed in the code. Some more general changes:
- [ ] There are some eslint warnings. Please fix those
- [ ] The asset actions are broken now because you are missing the outlet for the modals
- [ ] Something is wrong with how you are loading data in the notes.  Notice the undefined
Screenshot 2024-05-27 at 09 21 20

app/components/assets/notes/index.tsx
Bhavyajain21: Fixed, using useRouterLoaderData

app/routes/_layout+/assets.$assetId.tsx
Bhavyajain21: Yes, fixed it!

app/components/assets/asset-qr.tsx
Bhavyajain21: Changes made according to the older sidebar width!

app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -84,6 +77,24 @@ export async function loader({ context, request, params }: LoaderFunctionArgs) {
 
     const asset = await getAsset({ organizationId, id });
 
+    const searchParams = getCurrentSearchParams(request);
```
DonKoko, line 81: We only use this on `assetId.overview`. Lets move it to the loader there so we keep things a bit cleaner.

app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -122,6 +133,12 @@ export async function loader({ context, request, params }: LoaderFunctionArgs) {
       title: asset.title,
     };
 
+    const qrObj = {
```
DonKoko, line 137: Same as above

app/routes/_layout+/assets.$assetId.overview.tsx
```diff
@@ -0,0 +1,371 @@
...
+                  src="/static/images/default_pfp.jpg"
+                  alt="custodian"
+                  className="size-10 rounded"
+                />
+                <div>
+                  <p className="">
+                    In custody of{" "}
+                    <span className="font-semibold">
+                      {/* {asset.custody?.custodian.name} */}
+                      Bhavya jain
```
DonKoko: not sure about this.

app/routes/_layout+/assets.$assetId.overview.tsx
```diff
@@ -0,0 +1,371 @@
...
+    };
+  };
+}
+
+export default function AssetOverview() {
+  const data = useRouteLoaderData<AssetType>("routes/_layout+/assets.$assetId");
+  const { asset, locale, qrObj } = data ?? {};
+
+  // const { asset, locale, qrObj } = useOutletContext<AssetType>();
+  // console.log(asset);
```
DonKoko: Dont forget to clean this up.

app/components/layout/header/index.tsx
Bhavyajain21: Changes Made

DonKoko (approved):

Issue #942, opened 2024-04-30
[Feature request]: Generating a PDF with booking overview


jurrejansen:
### Contact Details

jurre@shelf.nu

### Is your feature request related to a problem? Please describe?

"Option to print out the generated list of assets as a PDF for record keeping and for a physical reference on set."

### Describe the solution you'd like

Add a menu item under Actions called 'Generate PDF...' which will download a PDF with an overview of the booking. There's not really a reason to not let users download the PDF when the booking is in Draft but I feel we should only allow it when booking reaches Reserved.

- Workspace logo (if uploaded and workspace is Plus or Team)
- Generation date
- Workspace name
- Title (with booking name)
- Booking name
- Custodian
- Booking period
- Table with included assets (number, thumbnail, name, category, location, QR, checkbox)

SCENARIO
https://www.figma.com/file/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?type=design&node-id=4918-566890&mode=design

carlosvirreira:
This seems quite ready to assign. Do you have any comments?

DonKoko:
Okey, here are some ideas based on some research I did. 
So as always with remix, we can use a resource route, which is amazing. Here is an example from Sergio: https://sergiodxa.com/tutorials/generate-a-pdf-in-remix-with-resource-routes
Also I like that he uses this react pdf approach. I think this could be so easy to do and we can create a route in the admin dashboard so we can preview it.

Moreover this is a more general article with a lot of different libraries. They all look interesting in case we want to use one of them instead of `react-pdf`. https://dev.to/handdot/generate-a-pdf-in-js-summary-and-comparison-of-libraries-3k0p

carlosvirreira:
Thank you for the research it sounds promising. What can we do to get this moving? The design is ready - is this a bounty-able-feat?

jurrejansen:
@DonKoko processed feedback from user. They asked for including the asset thumbnail. Also I have changed the logo in top left to 1:1. Updated the original message with these things.

MohakBajaj:
Sorry! I'll be withdrawing from this issue. I'm unable to find time to work on it.

rajeshj11:
@jurrejansen please review the demo and PR changes. Please Let me know if any changes are needed.

PR #999, opened 2024-05-28
feat:Generating a PDF with booking overview #942

rajeshj11:
/claim #942 

Added pdf generator for booking review

demo video:
https://www.loom.com/share/7817717d84134914b45d2bc4d841b659

note: right now considering all the bookings have this feature.(even draft)


app/components/booking/generate-booking-pdf.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const isMobileDevice = /Mobi/.test(navigator.userAgent);
+    setIsMobile(isMobileDevice);
+  }, []);
+
+  const handleIframeLoad = () => {
+    setIframeLoaded(true);
+  };
+
+  const handleMobileView = () => {
+    window.location.href = url;
```
DonKoko, line 30: Remix has ways to handle navigation that I would rather use than this. Check out the hook `useNavigation`.

app/modules/qr/service.server.ts
```diff
@@ -436,3 +437,41 @@ export async function claimQrCode({
...
+  size: "small" | "medium" | "large" | "cable";
+}) {
+  try {
+    return await Promise.all(
+      assets?.map(async ({ id: assetId }) => {
+        try {
+          let qr = await getQrByAssetId({ assetId });
+
+          if (!qr) {
+            qr = await createQr({ assetId, userId, organizationId });
```
DonKoko: That should not be needed and will never happen. An asset cannot be without a QR code.

app/components/booking/generate-booking-pdf.tsx
```diff
@@ -0,0 +1,125 @@
...
+  AlertDialogContent,
+  AlertDialogDescription,
+  AlertDialogFooter,
+  AlertDialogHeader,
+  AlertDialogTitle,
+  AlertDialogTrigger,
+} from "~/components/shared/modal";
+import { Spinner } from "../shared/spinner";
+
+export const GenerateBookingPdf = ({ booking }: { booking: any }) => {
```
DonKoko: We dont use `any`. This should be properly typed.

DonKoko (changes requested):
Okey so overall your PR is really nice man, especially for a first PR working with shelf, however I dont like the ejs approach. It introduces another thing and i would rather for the most part use the tools we already have. So my suggestion is to use the approach of Sergio:  https://sergiodxa.com/tutorials/generate-a-pdf-in-remix-with-resource-routes

Here is how I would handle it step by step: 
1. Create a PDFReport component which we render inside the alert dialog(where we are currently showing the iframe).
2. We can then style this component using our existing tools and 
3. Update the resource route that you have made to actually work like in the example above with `renderToStream` from `"@react-pdf/renderer"` and use the same component we are previewing above
4. The download button should then be a link to that resource route which will trigger a download.

Also I think we should create a function that fetches the DB and gets us all the data we need for this, kind of like in the example. So the loader is very clean.

Let me know on discord if you have any questions.

DonKoko:
@rajeshj11 congrats on the PR. This I am way happier with than what we had before with ejs. I have deployed it to our staging env so the rest of the team can test it. I still have to review the files but I am doing that already.

rajeshj11:
Will look into it

Dockerfile
```diff
@@ -7,11 +7,54 @@ ENV NODE_ENV="production"
 ARG DEBIAN_FRONTEND="noninteractive"
 WORKDIR /src
 
-# Install openssl for Prisma
+# Install openssl for Prisma,
+#     apt-get install -y openssl && \
```
DonKoko: Please remove this old lines and update the comment to say that it installs dependencies for Pupp, chrome and prisma

jurrejansen:
- [ ] The time of when the PDF was generated doesn't seem to be correct.
Screenshot 2024-06-03 at 11 53 19

- [ ] The thumbnails images on the PDF should have fixed height and width and the image should be set to Cover. This to avoid images with weird dimensions breaking the table on the PDF

Screenshot 2024-06-03 at 11 55 24

If we're going to use the vertical border lines, let's make sure they're also on the spots I marked in the screenshot. Lets also make sure there's enough padding on the cells. Looks like the first column with the numbers has a fixed width and the width doesn't get bigger when the numbers go into double digits
Screenshot 2024-06-03 at 12 00 37
Screenshot 2024-06-03 at 12 02 03

- [ ] Line up the workspace logo with the rest of the content
Screenshot 2024-06-03 at 12 04 57

- [ ] For longer booking names, we can maybe shorten / truncate the booking name after a certain amount of character in the top header where it also displays generation date and page number

Screenshot 2024-06-03 at 12 09 14
Screenshot 2024-06-03 at 12 13 28

jurrejansen:
Hi @rajeshj11 here is some more stuff

- [ ] Remove the border around workspace logo
- [ ] Using the Download button downloads a ...pdf.txt file a minute after clicking it
- [ ] There's missing a space between the booking name and generation date in the header
- [ ] Change Booking:, Custodian: and Booking period: into Booking, Custodian and Booking period
- [ ] Change font weight of Booking, Custodian and Booking period into Medium
- [ ] Change column labels into Font size: 8px and Font weight: medium
- [ ] Add corner radius of 4 to the tables

Screenshot 2024-06-04 at 10 12 01

rajeshj11:
@jurrejansen
 Using the Download button downloads a ...pdf.txt file a minute after clicking it

1. I did understand the pdf.txt extension file. is it downloading as a txt file?(not able to replicate that. need more inputs on this like screenshots, in which device etc..)
2. Yes It will take a few minutes to download the file. when the user clicks the download button again it will hit the backend and directly download the pdf. (As I could not able to find a way to use the already generated iframe data. I have configured it like this.). please do suggest any approach we can use other than this.

jurrejansen:
HI @rajeshj11 yeah it's weird. Both on Safari and Chrome the download functionality that is part of the PDF previewer is working fine.

However the custom download button that was added in the footer of the modal is not working unfortunately.

- After clicking it takes a minute or so before something happens
- Then it look like stuff is downloading because the browser is responding
- Takes even longer before files start showing up in Finder / Explorer
- Getting a bunch of .txt and .json files

Made some recordings, in the second on you can see stuff starting to download after a while but nothing appears in Finder.

https://github.com/Shelf-nu/shelf.nu/assets/76112272/b64500cb-8dd8-4ed1-8597-0c5f94565c43

https://github.com/Shelf-nu/shelf.nu/assets/76112272/0d92f975-14a0-4084-a753-134b4b912663


[booking-checklist-2024-06-04.pdf.json](https://github.com/user-attachments/files/15550329/booking-checklist-2024-06-04.pdf.json)
[booking-checklist-2024-06-04.pdf.txt](https://github.com/user-attachments/files/15550330/booking-checklist-2024-06-04.pdf.txt)

rajeshj11:
@jurrejansen hey, can you please check now? addressed the download issue.

jurrejansen:
@rajeshj11 Nice one, works. There's still a 5 seconds delay or something though. Maybe we can  implement a loading state of the button.

Screenshot 2024-06-04 at 13 52 29

https://github.com/Shelf-nu/shelf.nu/assets/76112272/b97230aa-d41a-44fb-8a68-b7ce1bb35620

DonKoko (approved):

PR #1010, opened 2024-05-31
Implement NProgress inside the app 

### LINKED ISSUE #1008: [Feature request]: Implement NProgress inside the app
### Contact Details

_No response_

Sometimes its hard to know what is going on if a request is slower and if the app is processing it.

Implement nprogress.
This works very well with remix and should be implemented on a global level, so it works across the whole app. The loading bar should be customized to use the shelf primary color.
Here is a guide on how to implement it: https://sergiodxa.com/tutorials/use-nprogress-in-a-remix-app

_No response_

### Additional context

_No response_

Bhavyajain21:
/fixes #1008
/claim #1008


app/root.tsx
```diff
@@ -102,6 +110,29 @@ function Document({ children, title }: PropsWithChildren<{ title?: string }>) {
 }
 
 function App() {
+  let transition = useNavigation();
```
DonKoko: man this is good, but can you please put it in a custom hook, so we keep the Root component clean. Thanks

app/styles/nprogress.css
```diff
@@ -0,0 +1,13 @@
...
+  background: linear-gradient(
+    to right,
+    #ef6820,
+    #ef6820
+  );
+}
+
+#nprogress .peg {
+  box-shadow:
+    0 0 10px #ef6820,
```
DonKoko: Please use apply and the color classes from tailwind. This needs to be dynamic as if its not, when we decide to change the main color, it will be a nightmare, and we already have enterprise clients that want that. You can just use @apply from tailwind to give it the correct color. It should be `primary-700`

DonKoko (changes requested):
2 small changes required.

app/root.tsx
Bhavyajain21: Done!

app/styles/nprogress.css
Bhavyajain21: done!

DonKoko (approved):

PR #1035, opened 2024-06-08
Feat/scroll to error

### LINKED ISSUE #1018: [Feature request]: On asset edit - show the required field validation errors above the page
### Contact Details

_No response_

I tried to change the name of an asset 5 times. I thought the app was broken. Happens that I had to scroll all the way down to see the error. 

If you (platform) know of the error and show me SAVE buttons. Please warn me about my action not being possible to complete due to custom fields.

Show me the error with a toast or some other way please

_No response_

### Additional context

_No response_

Bhavyajain21:
/claim #1018 
/fixes #1018


app/utils/scroll-to-error.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (errorElements.length > 0) {
+      let firstError = errorElements[0];
+      const elementToScrollTo = firstError.previousElementSibling || firstError;
+      elementToScrollTo.scrollIntoView({ behavior: "smooth" });
+    }
+  };
+
+  scrollToFirstError();
+
+  const observer = new IntersectionObserver(
```
DonKoko: I am having trouble to understand how this works. I see the observer and I understand that part but how does it interact with `scrollToError`. Seems like scroll will always be called on all entries, not just one the ones that are not visible.

Also please update this code by adding comments how it works, it needs to be more clear to developers who have to use and interact with this.

DonKoko (changes requested):
Overall good, but I have some comments:

- [x] I don't like that we have to pass the `onSubmit` to the form. What I would rather do is create our own Form component that just re-exports the Form from remix by passing this handler. This way we do have to change the imports everywhere, but that's just a 1 time task. We don't have to think about it every time, we use the Form component as it will be automatically managed.
- [ ] The scrolling currently scrolls directly to the field. There should be an offset so it field doesn't stick directly to the top of the browser. See screenshot. This shouldn't happen
Screenshot 2024-06-10 at 12 50 44

- [ ] currently the scrolling happens even if the field is already on the dom. The idea was that we use the `intersectionObserver` so we only add the event handler for inputs that are not visible in the viewport.

- [ ] In general, using `document.querySelectorAll` in react is not recommended. This is not the highest priority but I would like to give it a try. So I think that if we have our own Form component we can use the form ref to access the children of the form and search them, instead of having to access the dom using document.selector. If that doesn't work you have to do some extra checks to make sure that document is available as I think this will cause hydration errors as we have no document on server so the onSubmit value will be different on client and server

app/utils/scroll-to-error.ts
Bhavyajain21: Hey @DonKoko , I have updated this method to scroll only to fields if it isn't visible in the viewport. Have added few comments as well.

app/components/CustomForm.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+interface CustomFormProps
+  extends Omit<React.ComponentPropsWithoutRef<"form">, "method" | "onSubmit"> {
+  method?: HTMLFormMethod;
+  encType?: HTMLFormEncType;
+  onSubmit?: React.FormEventHandler<HTMLFormElement>;
+  replace?: boolean;
+}
+
+export const CustomForm = React.forwardRef<HTMLFormElement, CustomFormProps>(
```
DonKoko: We should call it just `Form`. Logically its a custom form but for us in shelf, its our main form. And I know we can do a named import with "as" but that adds overhead. With vscode you can just click cmd+. and it will auto import. I dont want to have to manually adjust the imports every time I need to use the form.

app/components/CustomForm.tsx
DonKoko, line 1: File name is not following our convention. Should be `custom-form.tsx`

app/components/CustomForm.tsx
```diff
@@ -0,0 +1,27 @@
...
+  (props, ref) => <RemixForm ref={ref} {...props} onSubmit={scrollToError} />
```
DonKoko: Here we have an issue. If the form already has an onSubmit function? I think it needs to be something like:
```ts
onSubmit={(v) => {
  onSubmit(v);
  scrollToError()
}}
```

So if the developer using the form is already passing an onSubmit function it runs.

app/components/CustomForm.tsx
```diff
@@ -0,0 +1,27 @@
+import React from "react";
+import {
+  Form as RemixForm,
+  FormProps as RemixFormProps,
+} from "@remix-run/react";
+import { scrollToError } from "~/utils/scroll-to-error";
+
+type HTMLFormMethod = "post" | "get" | undefined;
```
DonKoko: I dont think you need those custom types. You can just use the RemixFormProps instead. I did some testing and some changes(except component name) and came to this:
```ts
import React from "react";
import {
  Form as RemixForm,
  type FormProps as RemixFormProps,
} from "@remix-run/react";
import { scrollToError } from "~/utils/scroll-to-error";

export const CustomForm = React.forwardRef<HTMLFormElement, RemixFormProps>(
  (props, ref) => (
    <RemixForm
      ref={ref}
      {...props}
      onSubmit={(v) => {
        scrollToError();
        props?.onSubmit?.(v);
      }}
    />
  )
);

CustomForm.displayName = "CustomForm";
```

DonKoko (commented):
Added some comments and made some changes myself that I pushed. There is still some stuff pending like the form name.

app/components/CustomForm.tsx
Bhavyajain21: Fixed!

app/components/CustomForm.tsx
Bhavyajain21: Fixed!

app/components/CustomForm.tsx
Bhavyajain21: Fixed by you!!

app/components/CustomForm.tsx
Bhavyajain21: Resolved

Issue #894, opened 2024-04-03
[Enhancement request]: Change date picker on custom field


carlosvirreira:
### Is your feature request related to a problem? Please describe?

I sometimes spend too much time clicking around with the current date modal for custom fields (date) to select the purchase date of my assets. 

Some of my assets are 10 years old. Lots of clicking around.

### Describe the solution you'd like

Can we use a date picker similar to the one we have in Bookings? Without the time field. Just date. Or just let me input a date in text.

DonKoko:
@Bhavyajain21 I would like you to work on this when you are done with your current issue. Lets discuss it.

DonKoko:
@Bhavyajain21 yes, it needs to be just a normal html date input, the same as with bookings but without the time. That means we dont need the popover or anything like this as that comes native in the browser.

/bounty 10

DonKoko:
@carlosvirreira this has been adjusted and released to live. We replaced the date picker with had with a native browser date picker which works smoother on all devices. We also included the button to clear it for the case when its a not required field and the user might want to do that. Would be great if you can inform the users that asked for it.

carlosvirreira:
Doing right now and will be part of the small improvement notes on an upcoming Newsletter.

PR #1046, opened 2024-06-10
Feat/fix datepicker

Bhavyajain21:
/claim #894


DonKoko (changes requested):
A few things:
- [ ] You should use the components we already have. Please use FormRow and Input from our own components. Then you dont have to handle the errors or styles manually. Also those components are pretty flexible so i think you should be able to achieve everything.
- [ ] You need the clear field button we had before. Imagine a user puts a date they want to remove/clear because the field is not required. This should be supported.

DonKoko (approved):

PR #1049, opened 2024-06-11
1028 bug user are displayed only with first name and lower capital letters

DonKoko:
Resolve multiple issues related to custodian name:
- created function that resolves custodian name from teamMember (resolveTeamMemberName)
- updated dynamic select to use renderItem also for displaying the trigger button + implemented truncate
- updated multiple queries that select user from teamMember to also include first and last name
- updated the relevant types realted to selecting teamMember
- updated asset._index to better display the custodian name using resolveTeamMemberName()
- updated asset.overview.check-out dynamic select, to render the items including email using resolveTeamMemberName
- updated kit.check-out dynamic select, to render the items including email using resolveTeamMemberName
- updated asset and kit page to better display name of custodian


DonKoko:
@rockingrohit9639 Pushed some more changes, did the update also on DynamicDropdown. What we need to do is make sure the model filters work with this:

1. If the custodian doesnt have a user, it should search in the `custodian.name`
2. If the custodian has a user attached to it, it should search in `user.firstName` & `user.lastName` & `user.email`

PR #1059, opened 2024-06-13
Sort lists

DonKoko:
Added list sorting to asset index. Lets test it and if its good, we can implement it on other pages as well.

### DIFF
--- a/app/components/forms/select.tsx
+++ b/app/components/forms/select.tsx
@@ -17,7 +17,8 @@ const SelectTrigger = React.forwardRef<
     <SelectPrimitive.Trigger
       ref={ref}
       className={tw(
-        "select-trigger flex w-full items-center justify-between rounded border border-gray-300 bg-white px-3 py-2 text-[16px] text-gray-500 placeholder:text-gray-500 focus:border-primary-300 focus:outline-none focus:ring-2 focus:ring-primary-25 focus:ring-offset-2  disabled:opacity-50 ",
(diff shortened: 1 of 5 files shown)


DonKoko:
@carlosvirreira @jurrejansen i quickly implemented that for assets index. Its deployed to staging env. Can you take a look and if there are no bugs, we can release it and then later add it to different places.

carlosvirreira:
cool will test - Jurre is busy with production work for walkthrough (some shots) so I can better test this after my call

carlosvirreira:
Comments (Updating bit by bit):

1) I think we need a button. Specially on the case shown below it is quite confusing. 
Screenshot 2024-06-14 at 10 58 18

For example: The default filtering could be desired but you need to swap down the dropdown selector to get it executed instead of just pressing 'apply'

Screenshot 2024-06-14 at 10 58 46


https://github.com/Shelf-nu/shelf.nu/assets/12449255/fad44342-5054-496a-8890-e0a10dece6f1

DonKoko:
> For example: The default filtering could be desired but you need to swap down the dropdown selector to get it executed instead of just pressing 'apply'

Not true. Default state is always loaded, without you having to touch the filter. 
If you want to show the default state in the button, that's a different thing. When you land on the page with no filters, you are filtered on Date created + desc. That's how it is now. This hasn't been changed. If you want to display that, its fine.
let me know if with that change you still want the button to apply sorting. I can add it
I have 2 options:
image
image

Keep in mind that this long thing will always be visible, even if you are in default state and no filters are actually active.

carlosvirreira:
I very much like the one that it states clearly what the current filtering is!

That removes all my other concerns. 
Screenshot 2024-06-14 at 12 25 10

DonKoko:
@carlosvirreira awesome, sounds good. I am merging this so our users can test it a bit. Then we can implement it in other views and places. Its quite reusable so it should be straight forward.

Issue #1068, opened 2024-06-14
[Feature request]: Custom primary Asset ID field


Teraskull:
### Contact Details

anton.grouchtchak@gmail.com

### Is your feature request related to a problem? Please describe?

Our inventory is already labeled with our own QR codes and unique 6 digit asset IDs.

Would be great to have a basic `Asset ID` field that only allows unique values, and use it as a primary/secondary ID.

For example, allow accessing the asset using that custom asset ID like this: `app.shelf.nu/assets/123456/overview`.

### Describe the solution you'd like

Add a basic `Asset ID` field to the asset and allow accessing the asset with only that asset ID.

Here's a quick mockup of the field:
![image](https://github.com/Shelf-nu/shelf.nu/assets/24798198/78861ebe-d7a1-483b-aef1-6f7e1d2432ff)

### Describe alternatives you've considered

I have been testing https://github.com/hay-kot/homebox, but unfortunately, it's no longer being supported and developed.

The assets there can be accessed by `/item/<Asset UUID>` - `homebox.fly.dev/item/66fed3c4-675c-4de3-b320-b5a2484fcc8a`

But they can also be accessed by `/assets/<Asset ID>` - `homebox.fly.dev/assets/000160`, which then redirects to the primary  `/item/<Asset UUID>`.

The only downside here is that the `Asset ID` field is not unique, so it will return a list of all assets with the same asset ID if there are duplicate entries.

### Additional context

This will be useful for scanning existing QR Codes with the built-in scanner, and open the asset page if the asset ID exists.

carlosvirreira:
I've heard of this need from other users too.

"I think a separate ID for internal use and providing the user with the option to use their own would be beneficials.
 
For my use case, when I’m assisting users, I will request they give me the Asset tag of the Device. IE (KTOS06945) this is a lot easier for the user to read over the phone then a long string of random numbers and letters with no consistency.  "

rasithahewage:
Hi , Can we create our own id as Asset ID not SAM-?

carlosvirreira:
@rasithahewage this is not possible!

rasithahewage:
Thnks for the reply and Is there solution to add a our own D for it _

PR #1078, opened 2024-06-19
Feature: 1068 Add a custom asset ID field

Teraskull:
Hi! This is my attempt at being useful.

First of all, thank you for this amazing software. The best UI among all existing inventory management systems.

__Full disclosure__: I am a backend dev and have little experience in frontend. This is the first time I'm touching Remix and React.

I do not expect this PR to be merged, this is only to show the general UX of the custom Asset ID feature.

This is a rough naive working PoC for #1068 that adds an optional and unique `propertyId` field to the `Asset` table.

This PR misses a lot of implementation details, types, and does not account for a lot of places where the code should be updated.

### Notes:

* I named the field `propertyId` to avoid confusion with Asset ID. It can be renamed to whatever fits best.
* The position of the asset text field also does not matter.
* If you duplicate an asset that has a Property ID, the duplicate will not have a Property ID set.
* If no property ID is set, the "Property ID" filed in the asset details will not be shown.
* Updated the `searchVector` to search by `propertyId`.
* Added a Property column to the `/assets` page. It's completely optional because it takes up a lot of space.
* I added a `/property/$propertyId` page that redirects to the corresponding asset page if it exists. Otherwise show an "Asset not found" page. The route name is also an example, can be whatever.
* This redirect feature is pretty important since we can make our own QR Code scanner that redirects to an asset page, using an existing ID from our asset stickers.

### Known Issues:

* Creating/editing/duplicating assets with a `propertyId` is handled. I have no idea what other processes are affected.
* The field is unique, and if you try creating an asset with a duplicate `propertyId` it will throw a hardcoded `maybeUniqueConstraintViolation` error that says "Asset __name__ is already taken".

image

### Screenshots of the feature:

On the list of assets:

image

Creating/editing an asset:

image

On the asset page:

image

Overall, this does not look like a big feature to implement, I'm just unfamiliar with the codebase and the general flow.

Thanks!

### DIFF
--- a/app/components/assets/form.tsx
+++ b/app/components/assets/form.tsx
@@ -56,6 +56,7 @@ export const NewAssetFormSchema = z.object({
     .string()
     .optional()
     .transform((val) => (val ? +val : null)),
+  propertyId: z.string().transform((val) => val.trim()),
   addAnother: z
     .string()
     .optional()
@@ -74,6 +75,7 @@ interface Props {
   valuation?: Asset["valuation"];
   qrId?: Qr["id"] | null;
   tags?: Tag[];
+  propertyId?: Asset["propertyId"];
 }
 
 export const AssetForm = ({
@@ -87,6 +89,7 @@ export const AssetForm = ({
   valuation,
   qrId,
   tags,
+  propertyId,
 }: Props) => {
   const navigation = useNavigation();
 
@@ -369,6 +372,28 @@ export const AssetForm = ({
               {currency}
             </span>
           </div>
+        </FormRow>
+
+        <FormRow
+          rowLabel={"Property ID"}
+          subHeading={
+            <p>
+              Specify the property ID of your asset.
+            </p>
+          }
+          className="border-b-0 pb-[10px]"
+          required={false}
+        >
+          <Input
+            label="Property ID"
+            hideLabel
+            name={zo.fields.propertyId()}
+            disabled={disabled}
+            error={zo.errors.propertyId()?.message}
+            className="w-full"
+            defaultValue={propertyId || ""}
+            required={zodFieldIsRequired(FormSchema.shape.propertyId)}
+          />
         </FormRow>
 
         <AssetCustomFields zo={zo} schema={FormSchema} />
--- a/app/modules/asset/service.server.ts
+++ b/app/modules/asset/service.server.ts
@@ -92,6 +92,33 @@ export async function getAsset<T extends Prisma.AssetInclude | undefined>({
   }
 }
 
+export async function getAssetByPropertyId<T extends Prisma.AssetInclude | undefined>({
+  propertyId,
+  organizationId,
+  include,
+}: Pick<Asset, "propertyId"> & {
+  organizationId: Asset["organizationId"];
+  include?: T;
+}): Promise<AssetWithInclude<T>> {
+  try {
+    const asset = await db.asset.findFirstOrThrow({
+      where: { propertyId, organizationId },
+      include: { ...include },
+    });
+
+    return asset as AssetWithInclude<T>;
+  } catch (cause) {
+    throw new ShelfError({
+      cause,
+      title: "Asset not found",
+      message:
+        "The asset you are trying to access does not exist or you do not have permission to access it.",
+      additionalData: { propertyId, organizationId },
+      label,
+    });
+  }
+}
+
 /** This is used by both  getAssetsFromView & getAssets
  * Those are the statuses that are considered unavailable for booking assets
  */
@@ -614,9 +641,10 @@ export async function createAsset({
   customFieldsValues,
   organizationId,
   valuation,
+  propertyId,
 }: Pick<
   Asset,
-  "description" | "title" | "categoryId" | "userId" | "valuation"
+  "description" | "title" | "categoryId" | "userId" | "valuation" | "propertyId"
 > & {
   qrId?: Qr["id"];
   locationId?: Location["id"];
@@ -670,6 +698,7 @@ export async function createAsset({
       qrCodes,
       valuation,
       organization,
+      propertyId,
     };
 
     /** If a categoryId is passed, link the category to the asset. */
@@ -768,6 +797,7 @@ export async function updateAsset({
   currentLocationId,
   userId,
   valuation,
+  propertyId,
   customFieldsValues: customFieldsValuesFromForm,
 }: UpdateAssetPayload) {
   try {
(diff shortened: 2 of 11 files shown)


carlosvirreira:
Amazing @Teraskull we are super grateful that you took the time to submit a POC we have a scheduled discussion for it tomorrow on our Discord to provide some more info from our end. 

You rock! Thank you!

Teraskull:
Thanks a lot! I appreciate it @carlosvirreira 

Would be super great if the QR Code block on the asset page gets hidden if an alternative ID is set, since the original ID will not be used for stickers anyways.

app/components/assets/form.tsx
```diff
@@ -371,6 +374,28 @@ export const AssetForm = ({
           </div>
         </FormRow>
 
+        <FormRow
+          rowLabel={"Property ID"}
+          subHeading={
+            <p>
+              Specify the property ID of your asset.
```
DonKoko, line 381: I think we should add some more info about this. Like what is the purpose of this field. I think some of our users might be confused especially if they are not familiar with the terminology.

app/components/assets/form.tsx
```diff
@@ -56,6 +56,7 @@ export const NewAssetFormSchema = z.object({
     .string()
     .optional()
     .transform((val) => (val ? +val : null)),
+  propertyId: z.string().transform((val) => val.trim()),
```
DonKoko, line 59: Doesn't this make it required? This field should be optional.

app/modules/asset/service.server.ts
```diff
@@ -92,6 +92,33 @@ export async function getAsset<T extends Prisma.AssetInclude | undefined>({
   }
 }
 
+export async function getAssetByPropertyId<T extends Prisma.AssetInclude | undefined>({
```
DonKoko, line 95: 🤟🏼

app/routes/_layout+/assets.$assetId.overview.tsx
```diff
@@ -268,6 +268,15 @@ export default function AssetOverview() {
                 </div>
               </li>
 
+              {asset?.propertyId ? (
```
DonKoko, line 271: Indentation is broken. Make sure you use prettier so your formatting is consistent with our codebase. Also when you run the commit message, our post-commit hooks should also fix the formatting so you can also do it this way.

app/routes/_layout+/property.$propertyId.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    return redirect(`/assets/${asset.id}`);
+
+  } catch (cause) {
+    const reason = makeShelfError(cause);
+    throw json(error(reason));
+  }
+}
+
+export default function Property() {
```
DonKoko, line 42: You dont need a default export here. This is basically a "resource" route that will just handle the redirection.

app/routes/_layout+/property.$propertyId.tsx
```diff
@@ -0,0 +1,44 @@
...
+      userId,
+      request,
+      entity: PermissionEntity.asset,
+      action: PermissionAction.read,
+    });
+
+    const asset = await getAssetByPropertyId({
+      propertyId,
+      organizationId,
+      include: {},
```
DonKoko, line 31: No need to pass the empty object.

DonKoko (changes requested):
Thanks for opening the PR and doing such a high quality job at your first attempt. It is really appreciated.
I have done a first run through the code and added some comments of small things I noticed, but overall your PR is great. 
It covers most stuff and keeps consistent with our coding manners so that's amazing.

We haven't had the time yet to discuss this internally but I will try to address the points in your proposal that I feel need addressing:

> I named the field propertyId to avoid confusion with Asset ID. It can be renamed to whatever fits best.

I think its a great name. I can't think of anything better right now. Will let you know if it needs updating

> If you duplicate an asset that has a Property ID, the duplicate will not have a Property ID set.
> If no property ID is set, the "Property ID" filed in the asset details will not be shown.
> Updated the searchVector to search by propertyId.

Great 🤟🏼

> Added a Property column to the /assets page. It's completely optional because it takes up a lot of space.

I personally like it and don't think its a problem, but I think it depends on how commonly used it will be by our users. We are planning to soon work on a feature that allows you to choose which columns to show, so once we have that it will be even less of an issue.

> This redirect feature is pretty important since we can make our own QR Code scanner that redirects to an asset page, using an existing ID from our asset stickers.

I agree the redirect is great. Still I would love to hear a bit more how you are planning to use that. Very curious.

> Creating/editing/duplicating assets with a propertyId is handled. I have no idea what other processes are affected.

I think you handled a lot of the important cases. I will think about what we are missing but off the top of my head, 2 things that are quite important:
- Asset imports
- Backup exports & backup imports

The field should be handled in those cases as well. Its just a string field so it should be pretty straightforward to add.

> The field is unique, and if you try creating an asset with a duplicate propertyId it will throw a hardcoded `maybeUniqueConstraintViolation` error that says "Asset name is already taken".

Yeah I understand what you mean. Its just kind of hardcoded, as its currently meant to handle just a singular field per model called name. We would need to adjust how `maybeUniqueConstraintViolation` to be able to show the correct error.


This is my feedback for now. Lets address the points from above and once we discuss it internally, we will let you know if any further adjustments are needed.

app/routes/_layout+/property.$propertyId.tsx
DonKoko, line 1: Something I realized here when discussing it with Carlos. This will work however, we have to consider the same flow we have when scanning a normal QR. There are quite some edge cases to be handled like when users are logged in or not, whether they have access to that code and so on. We need to handle all the same cases here.

DonKoko:
@carlosvirreira Something we should already prepare is an article on our knowledge base on how the `propertyId` works and how users can use the custom url to redirect to an asset. I think a lot of users would love this feature. Once we have a link for this article, even if its not published yet, we can add it to the sub title of the field label.

carlosvirreira:
> @carlosvirreira Something we should already prepare is an article on our knowledge base on how the `propertyId` works and how users can use the custom url to redirect to an asset. I think a lot of users would love this feature. Once we have a link for this article, even if its not published yet, we can add it to the sub title of the field label.

We absolutely 100% will add to marketing site. I will pre-make the page and send the url here.

carlosvirreira:
> @carlosvirreira Something we should already prepare is an article on our knowledge base on how the `propertyId` works and how users can use the custom url to redirect to an asset. I think a lot of users would love this feature. Once we have a link for this article, even if its not published yet, we can add it to the sub title of the field label.

Here it is : https://www.shelf.nu/knowledge-base/understanding-and-using-custom-property-ids-in-shelf 

Let me know if amendments need be made to the kb article

DonKoko:
Closing due to inactivity. Feel free to re-open the PR when you want to resume this.

PR #1080, opened 2024-06-20
updating sentry to latest version and enabling autoInstrumentRemix

DonKoko:


package.json
```diff
@@ -4,7 +4,7 @@
   "sideEffects": false,
   "type": "module",
   "scripts": {
-    "build": "remix vite:build",
+    "build": "NODE_OPTIONS='--import=./server/instrument.server.mjs' remix vite:build",
```
onurtemizkan: This update should be in your `start` (and optionally `dev`) script instead of `build`.

PR #1112, opened 2024-06-28
refactor: adjust sizing of sidebar menu

DonKoko:
@carlosvirreira @jurrejansen, so as we discussed before I was a bit triggered by our sidebar. I feel like the menu items are too huge and the icons are very big and bulky with very wide stroke. Also our icons have different color than the text so that also looks not so great. Especially when i compare it to some other sidebars that look so much cleaner. 

### Mintlify example: 
Screenshot 2024-06-28 at 15 53 54
Screenshot 2024-06-28 at 15 53 50

I also really like how their close/open works.
Also this part is kinda messy in terms of code and I think it should be in general cleaned up and made easier to maintain and work with.

I would also like to consider the idea of using radix icons in the sidebar, as they just work. We dont need to export and do some shenanigans for each icon for it to work. https://www.radix-ui.com/icons

I have deployed this to testapp(https://github.com/Shelf-nu/shelf.nu/actions/runs/9713220465) so u can see it and lmk what u think. Its very experimental for now.

### DIFF
--- a/app/components/dashboard/checklist.tsx
+++ b/app/components/dashboard/checklist.tsx
@@ -47,7 +47,9 @@ export default function OnboardingChecklist() {
             <div className="flex h-full items-start justify-between gap-1 rounded border p-4">
               <div className="flex items-start">
                 <div className="mr-3 inline-flex items-center justify-center rounded-full border-[5px] border-solid border-primary-50 bg-primary-100 p-1.5 text-primary">
-                  <AssetsIcon />
+                  <i className="size-5">
+                    <AssetsIcon />
+                  </i>
                 </div>
                 <div className="text-[14px]">
                   <div className="mb-3">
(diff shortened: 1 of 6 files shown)


jurrejansen:
@DonKoko is there a demo environment or something like this where I can check how the maximizing / minimizing works with the Mintlify sidebar component?

I agree let's make it more compact in general. The current implementation of what you pushed to testapp looks pretty good. However there are issues when sidebar is minimized and the stroke width of icons and text doesn't feel right. Spent a bit of time on design and basically looks like what you implemented but I would like to be very specific with paddings / spacings / stroke width, etc.

Screenshot 2024-07-01 at 13 32 58
Screenshot 2024-07-01 at 13 33 08

The slight color difference between icon and text is intentional. I personally like it but let's see.

Using the Radix Icons would drastically limit the amount of icons we can use. Would prefer to find a procedure that works for using the Untitled UI ones. 

Screenshot 2024-07-01 at 13 42 02

The sidebar was the only place where they were looking good and contained actually. Besides the edited icon for scanning I provided and the the one for Calendar. Stroke width was too thin. But if we can figure out a procedure to make the icons look good and consistent for the elements down below we're good.

Screenshot 2024-07-01 at 13 45 35

DonKoko:
@jurrejansen thanks man. I agree with most of the stuff you said. Once we finish the stuff I am busy with I will pick this up again.

Issue #825, opened 2024-03-06
[Bug]: All "other" sessions of user should be destroyed when password is changed


DonKoko:
### What happened?

1- login From two browser at a time [ From Chrome browser and From Mozilla
Firefox ]
2- Change password in setting from chrome browser
3- Now Check Mozilla Firefox
4- Your Session Got Updated in place of expiration

### What is the expected behaviour?

All all sessions from that user in supabase need to be destroyed expired/destroyed.

### Version

Hosted (app.shelf.nu)

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

rockingrohit9639:
I can work on this issue @DonKoko

rajeshj11:
@DonKoko can you please assign it to me?

PR #1118, opened 2024-06-29
feat:handling the reset password logout

rajeshj11:
/claim #825 
addressed the feedback changes:

https://github.com/Shelf-nu/shelf.nu/pull/1094


rajeshj11:
@DonKoko @rphlmr Could you please check the changes?

rajeshj11:
v1-approach demo:
https://www.loom.com/share/69f3cefe9cf4427586b288a37eb2c53f

v2-approach-demo:
https://www.loom.com/share/69f3cefe9cf4427586b288a37eb2c53f

DonKoko (commented):
2nd approach I like a lot, but its very slow. My tests on local show me that this request takes between 150ms to 300ms which is too much to do on every request. 
Can we try to directly query the token from the auth schema? We might need to do a custom sql query in prisma to do that.

.env.example
```diff
@@ -50,4 +50,8 @@ GEOCODE_API_KEY="geocode-api-key"
 SENTRY_ORG="sentry-org"
 SENTRY_PROJECT="sentry-project"
 SENTRY_DSN="sentry-dsn"
+REFRESH_APPROACH_VERSION="v1" #possible values v1, v2
```
DonKoko: I want to test this in our staging server. Can you remove this part with the ENV variable and leave option 2 as default

app/modules/auth/service.server.ts
```diff
@@ -299,6 +307,32 @@ export async function getAuthResponseByAccessToken(accessToken: string) {
...
+      LIMIT 1 
+    `;
+  const t1 = performance.now();
+
+  // eslint-disable-next-line no-console
+  console.log(`Call to validateSession took ${t1 - t0} milliseconds.`);
+
+  if (result.length === 0) {
+    //logging for debug
+    Logger.error("Refresh token is invalid or has been revoked");
```
DonKoko: Here you should log a shelf error, so we can see it in our errors logs.

DonKoko (changes requested):
Some small changes and then we can test it on staging server. I still find it quite slow when I do it on local, but production servers are faster so I want to give it a try.

.env.example
rajesh-jonnalagadda: ack

app/modules/auth/service.server.ts
rajesh-jonnalagadda: ack

rajesh-jonnalagadda: if (!isValidSession) {
        throw new ShelfError({
          cause: null,
          message: "Your session has expired. Please log in again.",
          label: "Auth",
        });
      }
      we are already doing it. this will log

.env.example
rajesh-jonnalagadda: if we don't pass the value by default it will consider v2 approach

DonKoko: Still that should be removed from the .env.example file. Also the code that checks which mode is active should be removed: https://github.com/Shelf-nu/shelf.nu/pull/1118/files#diff-65810b1364bf0c64d3666ec80137196e42d96762a19db504aaf2f88141e2d3fbR83

DonKoko (commented):
Okey I tested this in the staging server and the performance is way better than my local tests. It takes between 50-80ms per request which is doable.

@rphlmr do you mind reviewing this last revision of the PR. I want to get your opinion and if you think we are missing something.

app/modules/auth/service.server.ts
rphlmr: You can pass a `new ShelfError(...)` in `Logger.error` if you need it to follow the error reporting pattern.

```ts
Logger.error(new ShelfError({...}))
```

app/routes/_auth+/reset-password.tsx
```diff
@@ -70,10 +71,16 @@ export async function action({ context, request }: ActionFunctionArgs) {
 
         const authSession = await refreshAccessToken(refreshToken);
 
-        await updateAccountPassword(authSession.userId, password);
+        await updateAccountPassword(
+          authSession.userId,
+          password,
+          authSession.accessToken
+        );
+        //on updating the password. it is removing the session_id from the collection. so we need to create a new session
```
rphlmr: Are we sure about that? Even with using `others`?

server/middleware.ts
```diff
@@ -72,11 +76,25 @@ export function refreshSession() {
...
 
-    if (!auth || !isExpiringSoon(auth.expiresAt)) {
+    if (!auth) {
       return next();
     }
-
+    let isValidSession = true;
+    if (!isExpiringSoon(auth.expiresAt)) {
+      isValidSession = await validateSession(auth.refreshToken);
+      if (isValidSession) {
```
rphlmr: ⁠`validateSession` doesn’t handle errors that can be thrown when calling ⁠`db.$queryRaw`.
Shouldn’t this be moved to the try-catch block?

server/middleware.ts
```diff
@@ -72,11 +76,25 @@ export function refreshSession() {
     const session = getSession<SessionData, FlashData>(c);
     const auth = session.get(authSessionKey);
 
-    if (!auth || !isExpiringSoon(auth.expiresAt)) {
+    if (!auth) {
       return next();
     }
-
+    let isValidSession = true;
+    if (!isExpiringSoon(auth.expiresAt)) {
```
rphlmr: @DonKoko Am I wrong to think that the sessions on other user devices will still be valid until it expires soon?

server/middleware.ts
rphlmr: I would go `next` if `isValidSession` && `!isExpiringSoon`.
Else, we can continue and try to `refreshAccesToken`. It should failed for invalid sessions since the refresh token has been revoked on password change.

rphlmr:
@DonKoko Feel free to resolve comments if the suggestions are not what you expect.

app/routes/_auth+/reset-password.tsx
DonKoko: @rajeshj11 I don't see anything in supabase docs about this. Where does this information come from? Based on how `signOut` works with others, the user current session will not be destroyed. 
Also if I look at the reset password code, you are first destroying the sessions and then resetting the password. 
Wouldn't it be more logical to destroy other sessions if the password reset is successful?

server/middleware.ts
DonKoko: Yes, indeed. This should be caught. @rajeshj11

DonKoko: Yeah indeed that makes more sense.
@rajeshj11 I am not even sure whats the purpose of those changes? Aren't those left over from trying the different approach. In main we handle it like this:
```tsx
/**
 * Refresh access token middleware
 *
 */
export function refreshSession() {
  return createMiddleware(async (c, next) => {
    //@ts-expect-error fixed soon
    const session = getSession<SessionData, FlashData>(c);
    const auth = session.get(authSessionKey);

    if (!auth || !isExpiringSoon(auth.expiresAt)) {
      return next();
    }

    try {
      session.set(authSessionKey, await refreshAccessToken(auth.refreshToken));
    } catch (cause) {
      session.flash(
        "errorMessage",
        "You have been logged out. Please log in again."
      );

      session.unset(authSessionKey);
    }

    return next();
  });
}

```

server/middleware.ts
DonKoko: I guess they will be, but consider that we check the session on each request, if any of the other devices tries to access shelf, there will be no session on supabase so they will anyways get automatically logged out of the app. Or am I missing something?

server/middleware.ts
rajesh-jonnalagadda: > I would go `next` if `isValidSession` && `!isExpiringSoon`. Else, we can continue and try to `refreshAccesToken`. It should failed for invalid sessions since the refresh token has been revoked on password change.

The above logic ensures that validateSession is only triggered when the local session does not expire soon. if it is expiring soon it will trigger the direct refreshAccestoken(let's say if the password changed. it will still throw error while trying to refresh token using refresh token)
FYI: @DonKoko

server/middleware.ts
rajesh-jonnalagadda: > I guess they will be, but consider that we check the session on each request, if any of the other devices tries to access shelf, there will be no session on supabase so they will anyways get automatically logged out of the app. Or am I missing something?

that's how the approach works and implemented

server/middleware.ts
rajesh-jonnalagadda: ack

rajesh-jonnalagadda: it was re-written like this

```
     if (!auth) {//if no auth
          return next();
     }
    let isValidSession = true;
    if (!isExpiringSoon(auth.expiresAt)) {// if not expiring soon then only hit the validateSession
      isValidSession = await validateSession(auth.refreshToken);
      if (isValidSession) {// if valid and and not expiring soon we will go to next
        return next();
      }
    }
    try {
      //if not true. then we can throw error
      if (!isValidSession) {// if not a valid session we simply throw an error which will clear the session.
        throw new ShelfError({
          cause: null,
          message: "Your session has expired. Please log in again.",
          label: "Auth",
        });
      }
     //only hit when the toke is expiring soon
      session.set(authSessionKey, await refreshAccessToken(auth.refreshToken));
    } catch (cause) {
      session.flash(
        "errorMessage",
        "You have been logged out. Please log in again."
      );

      session.unset(authSessionKey);
    }

**Comment by DonKoko** [comment]
@rajeshj11 I found a bug that happens on staging. Here is the process:
1. Create new user via OTP
2. Sign in with the new user
3. Follow onboarding
4. Select personal plan and click on create first asset
5. See error:
```
2024-07-09T12:42:39Z app[2874310b695008] ams [info]  <-- POST /onboarding?_data=routes%2F_welcome%2B%2Fonboarding
2024-07-09T12:42:39Z app[2874310b695008] ams [info]Call to validateSession took 65.71088499948382 milliseconds.
2024-07-09T12:42:40Z app[2874310b695008] ams [info]  --> POST /onboarding?_data=routes%2F_welcome%2B%2Fonboarding 204 1s
2024-07-09T12:42:40Z app[e82d950b056798] ams [info]  <-- GET /welcome?_data=routes%2F_welcome%2B%2Fwelcome
2024-07-09T12:42:40Z app[e82d950b056798] ams [info]Call to validateSession took 43.948870999738574 milliseconds.
2024-07-09T12:42:40Z app[e82d950b056798] ams [info]{"level":50,"time":1720528960540,"pid":337,"hostname":"e82d950b056798","msg":"Refresh token is invalid or has been revoked"}
```

**Comment by rajeshj11** [comment]
> @rajeshj11 I found a bug that happens on staging. Here is the process:
> 
> 1. Create new user via OTP
> 2. Sign in with the new user
> 3. Follow onboarding
> 4. Select personal plan and click on create first asset
> 5. See error:
> 
> ```
> 2024-07-09T12:42:39Z app[2874310b695008] ams [info]  <-- POST /onboarding?_data=routes%2F_welcome%2B%2Fonboarding
> 2024-07-09T12:42:39Z app[2874310b695008] ams [info]Call to validateSession took 65.71088499948382 milliseconds.
> 2024-07-09T12:42:40Z app[2874310b695008] ams [info]  --> POST /onboarding?_data=routes%2F_welcome%2B%2Fonboarding 204 1s
> 2024-07-09T12:42:40Z app[e82d950b056798] ams [info]  <-- GET /welcome?_data=routes%2F_welcome%2B%2Fwelcome
> 2024-07-09T12:42:40Z app[e82d950b056798] ams [info]Call to validateSession took 43.948870999738574 milliseconds.
> 2024-07-09T12:42:40Z app[e82d950b056798] ams [info]{"level":50,"time":1720528960540,"pid":337,"hostname":"e82d950b056798","msg":"Refresh token is invalid or has been revoked"}
> ```

let me check

**Comment by rajeshj11** [comment]
> > @rajeshj11 I found a bug that happens on staging. Here is the process:
> > 
> > 1. Create new user via OTP
> > 2. Sign in with the new user
> > 3. Follow onboarding
> > 4. Select personal plan and click on create first asset
> > 5. See error:
> > 
> > ```
> > 2024-07-09T12:42:39Z app[2874310b695008] ams [info]  <-- POST /onboarding?_data=routes%2F_welcome%2B%2Fonboarding
> > 2024-07-09T12:42:39Z app[2874310b695008] ams [info]Call to validateSession took 65.71088499948382 milliseconds.
> > 2024-07-09T12:42:40Z app[2874310b695008] ams [info]  --> POST /onboarding?_data=routes%2F_welcome%2B%2Fonboarding 204 1s
> > 2024-07-09T12:42:40Z app[e82d950b056798] ams [info]  <-- GET /welcome?_data=routes%2F_welcome%2B%2Fwelcome
> > 2024-07-09T12:42:40Z app[e82d950b056798] ams [info]Call to validateSession took 43.948870999738574 milliseconds.
> > 2024-07-09T12:42:40Z app[e82d950b056798] ams [info]{"level":50,"time":1720528960540,"pid":337,"hostname":"e82d950b056798","msg":"Refresh token is invalid or has been revoked"}
> > ```
> 
> let me check

is it happening in local?

File: server/middleware.ts (hunk 4 above)
Comment (rphlmr) [reply server/middleware.ts]: Ok I can understand the confusion here. Maybe the 'isValidSession' code should be moved to the `protect` middleware and `refreshSession` reverted?

Protect purpose is to check that a given route is protected (😅).
Sorry for not having seen this before @DonKoko

Comment (rajesh-jonnalagadda) [reply server/middleware.ts]: got it

File: app/modules/auth/service.server.ts [hunk 6]
```diff
@@ -299,6 +307,37 @@ export async function getAuthResponseByAccessToken(accessToken: string) {
...
+      LIMIT 1 
+    `;
+    const t1 = performance.now();
+
+    // eslint-disable-next-line no-console
+    console.log(`Call to validateSession took ${t1 - t0} milliseconds.`);
+
+    if (result.length === 0) {
+      //logging for debug
+      Logger.error("Refresh token is invalid or has been revoked");
```
DonKoko: This is still not handled correctly. You MUST log shelf errors. 
Example:
```tsx
          Logger.error(
            new ShelfError({
              cause: null,
              message: "Couldn't find custodian for asset",
              additionalData: { asset: a },
              label,
            })
          );
          ```

File: app/modules/auth/service.server.ts [hunk 7]
```diff
@@ -299,6 +307,37 @@ export async function getAuthResponseByAccessToken(accessToken: string) {
...
+    }
+    return result.length > 0;
+  } catch (err) {
+    Logger.error("Something went wrong while valdiating the session");
```
DonKoko: Same as above

app/routes/_auth+/reset-password.tsx
DonKoko: @rajeshj11 I am not sure what happened here.
Seems like you removed the code that destroys the session.
Can you please address the question that both me and Rapha asked?

app/routes/_welcome+/onboarding.tsx
```diff
@@ -133,6 +136,17 @@ export async function action({ context, request }: ActionFunctionArgs) {
      * This is to make sure that we have a stripe customer for the user.
      * We have to do it at this point, as its the first time we have the user's first and last name
      */
+
+    if (user) {
```
DonKoko: I don't understand this? Why wouldn't there be a session already? That should not happen, this is a protected route. If it happens that there is no session it needs to be resolved properly by finding whats the cause of the issue.

app/modules/auth/service.server.ts
```diff
@@ -299,6 +307,37 @@ export async function getAuthResponseByAccessToken(accessToken: string) {
   }
 }
 
+interface Session {
```
DonKoko: What is this type? We already have a type for our session and it has more data inside it. If you want to use it with a Partial, just use the proper session type we have.

DonKoko (changes requested):
Added some comments and some questions

server/middleware.ts
DonKoko: @rphlmr this has been updated. Can you take a look and lmk what you think?

app/modules/auth/service.server.ts
rajesh-jonnalagadda: ack

rphlmr: > What is this type? We already have a type for our session and it has more data inside it. If you want to use it with a Partial, just use the proper session type we have.

I think it is just to type the query result but It can be removed. We just have to now if the query returns an array with a given length. `unknow[]` should be enough

app/routes/_welcome+/onboarding.tsx
rajesh-jonnalagadda: > I don't understand this? Why wouldn't there be a session already? That should not happen, this is a protected route. If it happens that there is no session it needs to be resolved properly by finding whats the cause of the issue.

In Supabase, updating a user's password using the updateUserById function will invalidate the user's current session, logging them out as part of security best practices.

rphlmr: > > I don't understand this? Why wouldn't there be a session already? That should not happen, this is a protected route. If it happens that there is no session it needs to be resolved properly by finding whats the cause of the issue.
> 
> In Supabase, updating a user's password using the updateUserById function will invalidate the user's current session, logging them out as part of security best practices. If you need to update a user's password without invalidating their session, you will have to handle it with caution due to potential security risks. However, here's a general approach you might consider:

Can you share the doc about that? Maybe I miss something (or this is something new)

rphlmr (commented):
Thanks for moving that to the ⁠protect middleware, and again, sorry for not having seen that before.

@DonKoko For the pending comments, feel free to resolve or ping me again as you need.
I share your concerns, but as soon as you are okay, I am okay.

app/routes/_welcome+/onboarding.tsx
DonKoko: I also can't find it in the docs, but actually when I think about it, when they reset their password, we shouldn't automatically log them in. We should actually just send them to the login page so they can login with their new pwd. I think that is better for security and it will completely resolve the issue we are having here.

app/modules/auth/service.server.ts
rajesh-jonnalagadda: > > What is this type? We already have a type for our session and it has more data inside it. If you want to use it with a Partial, just use the proper session type we have.
> 
> I think it is just to type the query result but It can be removed. We just have to now if the query returns an array with a given length. `unknow[]` should be enough
we can do this. since we are using it only here, we can remove that interface and define it like below
 const result = await db.$queryRaw<{id:String, revoked:boolean}[]>`

DonKoko: Thats good. I just didn't want to have this type there with such a generic name like Session as it can get confusing.

app/routes/_welcome+/onboarding.tsx
rajesh-jonnalagadda: > I also can't find it in the docs, but actually when I think about it, when they reset their password, we shouldn't automatically log them in. We should actually just send them to the login page so they can login with their new pwd. I think that is better for security and it will completely resolve the issue we are having here.

I have implemented it without breaking the existing flow. Again it's your call how the flow should be.

app/routes/_auth+/reset-password.tsx
rajesh-jonnalagadda: > Are we sure about that? Even with using `others`?

With signout(token, 'others') it will sign out other sessions belonging to that particular user except the current session.
post updating the password with updateAccountPassword, it will clear the current session due to security reasons(which is a standard practice followed by many companies.)

rphlmr: I understand what you mean, and I agree it is best practice to invalidate all user sessions and recreate a new one in the current session that updates the password.
This behavior could prevent a “session copy attack” (so, changing the password without invalidating the whole session would be useless since the stolen token could be used with the duplicated refresh token to obtain a valid access token).

But, if we really want this behavior, we should sign out all the sessions (so, something other than `others`), including the current one, and then the `newSession` makes sense.

So, it is up to you, @DonKoko: Revoke all the sessions or remove the following code.

I think this is why we didn't understand it at first. Now I see what you are trying to achieve.

DonKoko:
@rajeshj11 @rphlmr 
Okey I am placing it in a separate comment because its becoming a bit hard to follow. Lets do the following and that completely resolves the issue with the session after reset pwd and I actually think its more secure.

1. When a user changes password, we destroy **all** sessions
2. Once they change their password, We dont give a new session to the user we just redirect to login and ask them to login with the new pwd.

I think this will completely resolve the issue. If you update the code I will resolve all comments related to this.

rajeshj11:
> @rajeshj11 @rphlmr Okey I am placing it in a separate comment because its becoming a bit hard to follow. Lets do the following and that completely resolves the issue with the session after reset pwd and I actually think its more secure.
> 
> 1. When a user changes password, we destroy **all** sessions
> 2. Once they change their password, We dont give a new session to the user we just redirect to login and ask them to login with the new pwd.
> 
> I think this will completely resolve the issue. If you update the code I will resolve all comments related to this.

will update the pr and let you know.

DonKoko (approved):

PR #1136, opened 2024-07-05
feat: Add QR codes to Kits

DonKoko:
This PR adds the functionality to kits to have a QR code.

Development checklist: 

- [x] create relationship with kit and qr
- [ ] ~~Create script that adds a QR code for each existing kit in the DB~~
- [x] Create QR code when a kit is created
- [x] Display Kit QR code on kit page
- [x] Implement updated logic on route `/$qrId`, that handles both kits and assets


carlosvirreira:
Feedback:
- Alignment of text on overlay is not ideal on the Assets tab (see image)
Screenshot 2024-07-10 at 10 29 27

I tested a few scenarios on my phone and all worked correctly. Not from a Superadmin user but from a normal user.

Screenshot 2024-07-10 at 10 50 18


Only one thing I noticed is that, whenever you are using the in app scanner. If you do not 'move the camera away' whenever a code is detected it seems to stay on some sort of loop. 

https://github.com/Shelf-nu/shelf.nu/assets/12449255/1f105f0e-5468-428b-9375-fc25eb869ec1

DonKoko:
@carlosvirreira 
alignment is fixed and deployed to staging.

The other issue is not related. I have created a bug for it #1150

PR #1159, opened 2024-07-11
feature: build version validation

DonKoko:


vite.config.ts
```diff
@@ -5,6 +5,13 @@ import devServer, { defaultOptions } from "@hono/vite-dev-server";
 import esbuild from "esbuild";
 import { flatRoutes } from "remix-flat-routes";
 import { cjsInterop } from "vite-plugin-cjs-interop";
+import { init } from "@paralleldrive/cuid2";
+
+const createHash = init({
+  length: 8,
+});
+
+const buildHash = process.env.BUILD_HASH || createHash();
```
rphlmr, line 14: How do you plan to inject BUILD_HASH env?
Do you think it could be generated during some docker build?

vite.config.ts
```diff
@@ -26,6 +33,21 @@ export default defineConfig({
...
+    assetsDir: `file-assets`,
+    rollupOptions: {
+      output: {
+        entryFileNames: `file-assets/${buildHash}/[name]-[hash].js`,
+        chunkFileNames() {
+          return `file-assets/${buildHash}/[name]-[hash].js`;
+        },
+        assetFileNames() {
+          return `file-assets/${buildHash}/[name][extname]`;
+        },
```
rphlmr, line 47: If changing the `assetsDir` is enough to solve the Sentry issue, this could be removed.
AFAIK there is no error handling on the static file middleware that serves these files.

rphlmr (approved):

vite.config.ts
DonKoko: Thats the thing, I haven't thought of it yet, but I guess your suggestion is the correct way. I will add this and ask for a review again.

vite.config.ts
DonKoko: Yeah I know, but we have the hash already so I see no issue with using it here. Do you see any downsides?

rphlmr: No downsides ;) If we don’t remember why we did that, we now have a note about it :p

DonKoko:
@rphlmr i think this is good now. I will merge it so we don't get those errors anymore in production and I will open a different issue about the remaining part we talked about that asks the user to refresh the page.

PR #1168, opened 2024-07-12
feat: create booking via asset/kit page

Bhavyajain21:


app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -209,7 +210,20 @@ export default function AssetDetailsPage() {
           </div>
         }
       >
-        {!isSelfService ? <ActionsDropdown /> : null}
+        {!isSelfService ? (
+          <>
+            <ActionsDropdown />
+            <Button
```
DonKoko: Self service should be allowed to book.

app/routes/_layout+/bookings.new.tsx
```diff
@@ -32,6 +32,8 @@ import { requirePermission } from "~/utils/roles.server";
  * This way all actions are available and its way easier to manage so in a way this works kind of like a resource route.
  */
 export async function loader({ context, request }: LoaderFunctionArgs) {
+  const url = new URL(request.url);
```
DonKoko: Please use the way we already have for getting params on the server: `getCurrentSearchParams`

app/routes/_layout+/bookings.new.tsx
```diff
@@ -154,17 +157,23 @@ export async function action({ context, request }: ActionFunctionArgs) {
...
-      // We force the as Date because we know that the booking.from and booking.to are set and exist at this point.
-      bookingFrom: (booking.from as Date).toISOString(),
-      bookingTo: (booking.to as Date).toISOString(),
-      hideUnavailable: "true",
-      unhideAssetsBookigIds: booking.id,
-    })}`;
-
-    return redirect(manageAssetsUrl);
+    const url = new URL(request.url);
+    const assetIds = url.searchParams.get("assetIds");
```
DonKoko: That is not the correct way to do it in the action. We don't send search params to actions. You need to add it to the formData and parse the array of ids using zod. You can see how we do it in the bulk actions. For an example you can check `tags.bulk-actions.ts`

Also you should pass the `assetIds` to upsert booking. The way it is now the assets won't get added.

app/routes/_layout+/bookings.new.tsx
```diff
@@ -154,17 +157,23 @@ export async function action({ context, request }: ActionFunctionArgs) {
...
+    if (assetIds) {
```
DonKoko: Instead of checking it like this, you should create a clear boolean so the code is better to read.

app/components/booking/form.tsx
```diff
@@ -377,7 +379,9 @@ export function BookingForm({
         </div>
         {isNewBooking ? (
           <div className="text-right">
```
DonKoko, line 376: You need to also add some hidden inputs to send the `assetIds` to the action using the formData.
Here is some pseudo code:
```ts

          {assetIds.map((item, i) => (
            <input
              key={item}
              type="hidden"
              name={`assetIds[${i}]`}
              value={item}
            />
          ))}
          ```

**Review by DonKoko — CHANGES_REQUESTED** [review]
Added comments

File: app/routes/_layout+/assets.$assetId.tsx (hunk 1 above)
Comment (Bhavyajain21) [reply app/routes/_layout+/assets.$assetId.tsx]: Resolved!

File: app/routes/_layout+/bookings.new.tsx (hunk 2 above)
Comment (Bhavyajain21) [reply app/routes/_layout+/bookings.new.tsx]: Changed the way of accessing the query params. Resolved!

File: app/routes/_layout+/bookings.new.tsx (hunk 4 above)
Comment (Bhavyajain21) [reply app/routes/_layout+/bookings.new.tsx]: Created a clear boolean. Resolved!

File: app/routes/_layout+/bookings.new.tsx (hunk 3 above)
Comment (Bhavyajain21) [reply app/routes/_layout+/bookings.new.tsx]: Parsed assetsId array using zod and pass assetIds to upsert. Resolved!

File: app/components/booking/form.tsx (hunk 5 above)
Comment (Bhavyajain21) [reply app/components/booking/form.tsx:376]: Added a hidden input. Resolved!

**Comment by Bhavyajain21** [comment]
Changes made. Please review @DonKoko

File: app/routes/_layout+/bookings.new.tsx [hunk 6]
```diff
@@ -131,6 +140,8 @@ export async function action({ context, request }: ActionFunctionArgs) {
       zone: hints.timeZone,
     }).toJSDate();
 
+    const assetIds = z.array(z.string()).parse(formData.getAll("assetIds"));
```
DonKoko: This is not the correct way to do it. Please look at the example i gave you with bulk actions as well as how we usually parse data. The 'from' and 'to' you see above are exceptions, not the way to do it. you need to use `parseData` and add the `assetIds` to the `NewBookingFormSchema`.  
Please always check our other actions and how we do things so we dont have to do so many reviews for small things.

app/routes/_layout+/bookings.new.tsx
```diff
@@ -204,6 +223,14 @@ export default function NewBooking() {
               : undefined
           }
         />
+        {assetIds.map((item, i) => (
```
DonKoko: This is not the correct place to add this. It should be in the form so it gets sent with the form data. This will never work.

DonKoko (changes requested):

app/routes/_layout+/bookings.new.tsx
Bhavyajain21: Added in the form component

app/routes/_layout+/bookings.new.tsx
Bhavyajain21: Sure thing, have got the assetIds using the parseData and added the assetIds to the NewBookingFormSchema

DonKoko (approved):

DonKoko (approved):
🤟🏼

PR #1176, opened 2024-07-16
chore: update use of disabled prop in Button component

DonKoko:
adjusted basic button component + removed a few uses of ControlledActionButton. Still needs testing and more refactoring


DonKoko:
I have replaced all uses of `ControlledActionButton` with just the clean `Button` component with its updated `disabled` prop. Those changes are quite critical also for the business of shelf. Could you please review? @rockingrohit9639

rockingrohit9639:
Sure @DonKoko, I will review it

rockingrohit9639 (approved):
🤟

PR #1211, opened 2024-07-26
Feat 1065 filter persistance on asset index

rajeshj11:


app/hooks/use-search-param-utils.ts
```diff
@@ -1,35 +1,103 @@
...
+) {
+  keys.forEach((key) => {
+    setSearchParams((prev) => {
+      prev.delete(key);
+      return prev;
+    });
+  });
+}
+
+export function destoryCookieValues(
```
DonKoko: Spelling mistake in the name. Also please annotate your code.

app/utils/cookies.server.ts
```diff
@@ -86,3 +86,36 @@ export async function initializePerPageCookieOnLayout(request: Request) {
   }
   return cookie;
 }
+
+export const createAssetFilterCookie = (orgId: string) =>
+  createCookie(`${orgId}_assetFilter`, {
+    path: "/",
```
DonKoko: This cookie should be sent only on the /assets page

app/utils/cookies.server.ts
```diff
@@ -86,3 +86,36 @@ export async function initializePerPageCookieOnLayout(request: Request) {
...
+    secure: process.env.NODE_ENV === "production",
+    maxAge: 60 * 60 * 24 * 365, // 1 year
+  });
+
+export async function getFiltersFromRequest(
+  request: Request,
+  organizationId: string
+) {
+  const url = new URL(request.url);
+  let filters = url.searchParams?.toString();
```
DonKoko: We already have a function for getting searchparams on the server: `getCurrentSearchParams`. Please use the tools we already have available and use in different places.

app/routes/_layout+/assets._index.tsx
```diff
@@ -159,6 +170,27 @@ export async function loader({ context, request }: LoaderFunctionArgs) {
       plural: "assets",
     };
 
+    const headers = new Headers();
```
DonKoko: We already have a method of creating and setting headers. Please use the approach we use everywhere. 
Example:
```ts
    const headers = [
      setCookie(
        await setSelectedOrganizationIdCookie(
          organizationsIds.find((orgId) => orgId === qr.organizationId) ||
            personalOrganization.id
        )
      ),
    ];
    ```

Comment (DonKoko) [inline app/routes/_layout+/assets._index.tsx]: I think everything from this line to line 192 should be abstracted into a function that returns the headers.

File: app/components/booking/status-filter.tsx [hunk 5]
```diff
@@ -17,11 +18,13 @@ export function StatusFilter({
   const disabled = isFormProcessing(navigation.state);
   const [searchParams, setSearchParams] = useSearchParams();
   const status = searchParams.get("status");
+  const { destoryCookieValues } = useCookieDestory();
 
   function handleValueChange(value: string) {
     setSearchParams((prev) => {
       /** If the value is "ALL", we just remove the param */
       if (value === "ALL") {
+        destoryCookieValues(["status"]);
```
DonKoko: I am a bit concerned about this approach, because that requires us to never forget to add this when having functions that clear filters and so on.
A suggestion, what if we create our own abstraction of `useSearchParams()` that handles this and makes sure that every time a value is removed from the search params, we also remove it from the cookie.

app/hooks/use-search-param-utils.ts
```diff
@@ -1,35 +1,103 @@
...
-import { useSearchParams } from "@remix-run/react";
+import { useLoaderData, useLocation, useSearchParams } from "@remix-run/react";
+import Cookies from "js-cookie";
+import type { loader } from "~/routes/_layout+/assets._index";
+
+type SetSearchParams = (
+  setter: (prev: URLSearchParams) => URLSearchParams
+) => void;
+
+export function useAssetIndexMeta() {
```
DonKoko: Please annotate what each of those new functions you added do.

DonKoko (changes requested):
Requested some changes mostly related to coding style and conventions we use.
I also did some testing on my localhost and it seems to work. Once those changes are implemented I will also ask a third person to review, in case we are missing something.

app/routes/_layout+/assets._index.tsx
rajesh-jonnalagadda:
```
const headers = [
      setCookie(userPrefsCookie),
      ...(filtersCookie? [setCookie(filtersCookie)] : [])
    ];

rajesh-jonnalagadda: > ```
> const headers = [
>       setCookie(userPrefsCookie),
>       ...(filtersCookie? [setCookie(filtersCookie)] : [])
>     ];
> ```

this is what you mean

app/components/booking/status-filter.tsx
rajesh-jonnalagadda: I could not able to figure out a way to do it due to some scenarios.

1. Not able to differentiate if the user has cleared the filters in the asset page or if the user is directly going to the asset page after some navigations. (in both cases search parameters are empty.)

DonKoko: If the user has search params saved in the cookie, they wont be empty as they come with the server response and useSearchParams checks on the client so I am not sure what you mean here.

DonKoko (changes requested):
Okey we still have quite some stuff to handle here. A few of the filters are currently broken because of this change. For example:
- [ ] manually clearing an item from the dropdown filters doesnt work anymore. I assume its because of what i mentioned before, that if we don't manage the `  destroyCookieValues(["s"]);` somehow globally, it will create sitiations like this where we forgot to add it in 1 place and it causes a bug.

So I did something to help out. I created the abstraction of the `setSearchParams` that automatically removed search params values that were removed from the cookie.
However this still needs to be further tested and the import of `useSearchParams` needs to be replaced. Can you please take care of that. 
I also placed one inline comment to address.

rajeshj11:
> Okey we still have quite some stuff to handle here. A few of the filters are currently broken because of this change. For example:
> 
> * [ ]  manually clearing an item from the dropdown filters doesnt work anymore. I assume its because of what i mentioned before, that if we don't manage the `  destroyCookieValues(["s"]);` somehow globally, it will create sitiations like this where we forgot to add it in 1 place and it causes a bug.
> 
> So I did something to help out. I created the abstraction of the `setSearchParams` that automatically removed search params values that were removed from the cookie. However this still needs to be further tested and the import of `useSearchParams` needs to be replaced. Can you please take care of that. I also placed one inline comment to address.

ok will check

DonKoko (approved):

PR #1215, opened 2024-07-26
feat: url shortener

DonKoko:
- implement url shortener middleware
- add env variable for short url
- update docs to explain how to use url shortener
- update functions that generate code to handle the short url


rphlmr (dismissed):
lgtm

PR #1219, opened 2024-07-31
feat: Added a createScanNote function which creates a note whenever a scan is created

### LINKED ISSUE #1200: [Feature request]: Activity event when manually updating a GPS coordinate
### Contact Details

_No response_

I notice someone updated the GPS coordinate to the wrong place. I want to see who did that action and how.

IF manually updated or via scan add an entry in the asset activity including the user who performed the action.

_No response_

### Additional context

_No response_

hunar1312:
resolves #1200


app/modules/scan/service.server.ts
```diff
@@ -34,7 +37,7 @@ export async function createScan(params: {
     };
 
     /** If user id is passed, connect to that user */
-    if (userId) {
+    if (userId && userId != "anonymous") {
```
DonKoko, line 41: I am not sure what the "anonymous" is. I dont see us using this anywhere or passing it to `createScan`.

DonKoko: NVM, i found it in some old code. Didnt even know we handle it like this :)

app/modules/scan/service.server.ts
```diff
@@ -138,3 +145,61 @@ export async function getScanByQrId({ qrId }: { qrId: string }) {
...
+  latitude?: Scan["latitude"];
+  longitude?: Scan["longitude"];
+  manuallyGenerated?: boolean;
+}) {
+  try {
+    let message = "";
+    const { assetId, organizationId } = await getQr(qrId)
+    if (assetId) {
+      if (userId && userId != "anonymous") {
+        const user = await db.user.findFirst({
```
DonKoko: Okey a few things.
1. IN general we dont do those inline queries anymore, because we need to try/catch them every time so its a lot of repeating code.
2. You should simply use `getUserByID` as that already handles catching errors and so on.

app/modules/scan/service.server.ts
```diff
@@ -138,3 +145,61 @@ export async function getScanByQrId({ qrId }: { qrId: string }) {
...
+        }
+        return await createNote({
+          content: message,
+          type: "UPDATE",
+          userId,
+          assetId
+        });
+      } else {
+        if (organizationId) {
+          const { userId: ownerId } = await getOrganizationById(organizationId)
```
DonKoko: Can you please annotate this, so its clear we took a decision here.

app/modules/scan/service.server.ts
```diff
@@ -138,3 +145,61 @@ export async function getScanByQrId({ qrId }: { qrId: string }) {
...
+          message = `**${userName}** performed a scan of the asset QR code`
+        }
+        return await createNote({
+          content: message,
+          type: "UPDATE",
+          userId,
+          assetId
+        });
+      } else {
+        if (organizationId) {
```
DonKoko, line 191: This should not happen, that there is no organizationId. If there is an assetId there will always be organization id. PLease annotate that we are just doing this extra check but it should never happen.

app/modules/scan/service.server.ts
```diff
@@ -138,3 +145,61 @@ export async function getScanByQrId({ qrId }: { qrId: string }) {
...
+          where: {
+            id: userId,
+          },
+          select: {
+            firstName: true,
+            lastName: true,
+          },
+        });
+
+        const userName = user?.firstName + " " + user?.lastName
```
DonKoko: We had this issue before, i think if the user doesnt have first and last name it will print "undefined" as a string. Can you please double check it. Also you can see how we handle it in other notes.

app/modules/scan/service.server.ts
```diff
@@ -138,3 +145,61 @@ export async function getScanByQrId({ qrId }: { qrId: string }) {
...
+        if (manuallyGenerated) {
+          message = `**${userName}** manually updated the GPS coordinates to **${latitude} ${longitude}**`
```
DonKoko: lat and long need to be separated by a coma. Also I think the bold is too much. Lets do italic instead.

DonKoko (changes requested):
Some small changes required. Overall i tested it and It looks good:
Screenshot 2024-08-01 at 10 39 22

DonKoko (approved):

PR #1236, opened 2024-08-09
Created profile page for team's registered users

### LINKED ISSUE #729: [Feature request]: Team member page
### Contact Details

_No response_

I am trying to see all assets provisioned to one user. Now, I have to use filters. I do not get historic data just the currently provided assets to the user. 

Ideally a page where I can see information on the user and the assets that are under his/her custody.

**Visiting team member page**
https://www.figma.com/design/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?node-id=5392-12557

**Visiting page of team member that has no assets in custody**
https://www.figma.com/design/ykvc0MhPeNLsrJ4tJsKosQ/Shelf-%7C-UX-Scenarios?node-id=6982-67556

_No response_

### Additional context

_No response_

hunar1312:
resolves #729


app/components/workspace/users-actions-dropdown.tsx
```diff
@@ -41,10 +43,21 @@ export function TeamUsersActionsDropdown({
...
       >
-        <DropdownMenuTrigger className="size-6 pr-2 outline-none focus-visible:border-0">
-          <i className="inline-block px-3 py-0 text-gray-400 ">
-            {disabled ? <Spinner className="size-4" /> : <VerticalDotsIcon />}
-          </i>
+        <DropdownMenuTrigger className="size-6 w-full pr-2 outline-none focus-visible:border-0">
+          {useButtonTrigger ? (
+            <Button
+              variant="secondary"
+              to="#"
```
DonKoko: This should not have the `to` as it will make it behave like a link. We don't want that.

app/components/workspace/users-actions-dropdown.tsx
```diff
@@ -41,10 +43,21 @@ export function TeamUsersActionsDropdown({
         onOpenChange={(open) => setOpen(open)}
         open={open}
       >
-        <DropdownMenuTrigger className="size-6 pr-2 outline-none focus-visible:border-0">
-          <i className="inline-block px-3 py-0 text-gray-400 ">
-            {disabled ? <Spinner className="size-4" /> : <VerticalDotsIcon />}
-          </i>
+        <DropdownMenuTrigger className="size-6 w-full pr-2 outline-none focus-visible:border-0">
```
DonKoko: I am not a fan of this implementation tbh. I would rather we make the component receive an optional component to be used as a trigger as thats basically the exception. Then in the place where we need a custom button we can pass it when rendering `<TeamUsersActionsDropdown/>`

app/routes/_layout+/bookings.tsx
```diff
@@ -158,24 +158,28 @@ export type RouteHandleWithName = {
 
 export default function BookingsIndexPage({
   className,
+  includeBulkActions = true,
 }: {
   className?: string;
+  includeBulkActions?: boolean;
```
DonKoko: Again, the same. If this is true by default, then it should be managed the other way around. The prop should be called `disableBulkActions` and should be false by default.

app/routes/_layout+/bookings.tsx
```diff
@@ -158,24 +158,28 @@ export type RouteHandleWithName = {
...
    * 1. When we are on the index route
    * 2. When we are on the .new route - the reason we do this is because we want to have the .new modal overlaying the index.
    * 3. When we are on the assets.$assetId.bookings page
+   * 4. When we are on the settings.team.users.$userId.bookings
    */
   const shouldRenderIndex =
     currentRoute?.handle?.name === ("bookings.index" as string) ||
     currentRoute?.handle?.name === "bookings.new" ||
-    currentRoute?.handle?.name === "$assetId.bookings";
+    currentRoute?.handle?.name === "$assetId.bookings" ||
```
DonKoko: At this point that we have 4 options, its better to create an array with the options and use `.includes()`

app/routes/_layout+/settings.team.users.$userId.assets.tsx
```diff
@@ -0,0 +1,403 @@
...
+    const reason = makeShelfError(cause, { userId });
+    throw json(error(reason), { status: reason.status });
+  }
+}
+
+export default function UserAssetsPage() {
+  return <AssetsList />;
+}
+
+const AssetsList = () => {
```
DonKoko: I am a bit confused here. The idea was to just import the `AssetsList` from the asset index, so we use the same component. Not sure why you added it here again? assets index already exports this component.

DonKoko (changes requested):
Wrote some comments in the code.

app/routes/_layout+/settings.team.users.$userId.assets.tsx
hunar1312: Few reasons for a custom asset list: 
1. Different custom empty state
2. Removing teamMember filter and it's dropdown as well
3. To remove sorting filters
4. Different Navigate parameter

@DonKoko maybe we can remove the ListAssetContent and import that from assets Index

DonKoko: Yeah then all those things need to be adjusted. Still repeating the 90% same code 2 times will create more issues here.

> 1. Different custom empty state

Allow the AssetIndex component to receive the empty state and drill it down to the list

> Removing teamMember filter and it's dropdown as well

I believe we already have ways to do that. Again u can add an optional prop for it.

> To remove sorting filters

You can leave those. They should be in the design. I dont know why they are not.

> Not sure what you mean

Again, this is something that can be passed as props

DonKoko:
@hunar1312 something i noticed. cmd + click on the users asset index fails. Most likely due to how navigate it used.

DonKoko (approved):

PR #1249, opened 2024-08-14
Bug 1184 date issue fix

### LINKED ISSUE #1184: [Bug]: I've set up a custom field of type Date, and when viewing an asset with this field set, it always shows one day earlier than I entered.
### Contact Details

_No response_

### What happened?

I've set up a custom field of type Date, and when viewing an asset with this field set, it always shows one day earlier than I entered. When re-entering Edit mode, the original correct date still shows there.

wrong-date-view-mode

wrong-date-edit-mode

Show correct date

### Version

Hosted (app.shelf.nu)

Chrome

### Relevant log output

_No response_

rajeshj11:
/closes #1184


rajeshj11:
@DonKoko can you please review these changes

app/utils/custom-fields.ts
```diff
@@ -183,9 +206,17 @@ export const getCustomFieldDisplayValue = (
...
-      : format(new Date(value.valueDate), "PPP"); // Fallback to default date format
+    if (hints) {
+      const dateFormatter = getDateTimeFormatFromHints(hints, {
+        year: "numeric",
+        month: "2-digit",
+        day: "2-digit",
+      }).format(new Date(value.valueDate));
+      const [month, day, year] = dateFormatter.split("/");
+      // Rearrange the components into YYYY-MM-DD format
+      return `${year}-${month.padStart(2, "0")}-${day.padStart(2, "0")}`;
```
DonKoko: We don't do that. The idea is that the formatting is done based on the user locale and timezone. That is what our dateformatters do. You are formatting the date and then hardcoding a different format that doesnt fit with our app.

Here are some examples. My PC has dates in american style:

image
image
image

As you can see they follow the same format everywhere, except the 1 place where we have the custom field where you are hardcoding the value.

DonKoko (changes requested):
Again, you did a lot of changes to the functions that format our date without needing to do them. We do exactly the same for bookings form and then displaying the bookings date on the index for example. You dont need to add any new features or handle any new case. Just need to implement it 1:1 like in bookings.

rajeshj11:
closing this pr as it is resovled in #1268

PR #1258, opened 2024-08-16
fix: updating layout export in root

DonKoko:
Root route exports were not updated to the latest Remix standard so it was creating an issue with unstyled errors when a certain layout doesnt have error boundary and the errors bubbled up to the root.


rphlmr (approved):
LGTM

PR #1259, opened 2024-08-16
feature: creating function that deletes user

DonKoko:
To delete a user we need to do a few manual things
 1. Delete the user itself. To do this we need to do some manual stuff. We need to find all entries that belong to the user, but not to his personal org. Then we need to transfer those entries to the org owner. The models to handle are:
- Asset
- Category
- Tag
- Location
- CustomField
- Invite
- Booking
- TeamMember
- Image
2. Delete the auth account of the user
3. Delete the auth account of the user
4. Send email that the user is deleted

### DIFF
--- a/app/components/assets/notes/note.tsx
+++ b/app/components/assets/notes/note.tsx
@@ -8,7 +8,7 @@ import { ActionsDopdown } from "./actions-dropdown";
 
 export type NoteWithDate = WithDateFields<NoteType, string> & {
   dateDisplay: string;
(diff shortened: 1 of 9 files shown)


rphlmr (approved):
LGTM

DonKoko:
@rphlmr after we discussed this on discord, you actually changed my mind. We will take a different approach, so I will create a new branch based on this and do a new approach. Thanks for your help.

PR #1304, opened 2024-09-06
Dashboard's top custodians table to include custodies via bookings

### LINKED ISSUE #1251: Feature: Dashboard "Top custodians" include custodians via bookings
After bookings were released, we never update the table on the dashboard to include custodians via bookings.

hunar1312:
resolves #1251


app/routes/_layout+/dashboard.tsx
```diff
@@ -97,6 +97,50 @@ export async function loader({ context, request }: LoaderFunctionArgs) {
         });
       });
 
+    const bookings = await db.booking
```
DonKoko: You should use `getBookings` function.

DonKoko (approved):

PR #1312, opened 2024-09-17
Update README.md

carlosvirreira:
Improved the readme slightly. If you could remove the how to get started part if you find it not valuable it would be appreciated.


DonKoko (approved):
Made some small changes to the get started section.

PR #1316, opened 2024-09-18
feat: bulk add and remove tags from assets

DonKoko:


DonKoko:
@carlosvirreira any feedback on this?

PR #1319, opened 2024-09-19
feat: allow linking of QR codes on import

### LINKED ISSUE #1290: Feat: Allow importing with linking to existing QR code
Sometimes users want to use the import feature and link the newly created assets to existing QR codes, instead of generating a new one.

In order for this to work we have to add some limitations that need to be checked while importing for each asset. If those conditions fail, fail the whole import and give an error to the user listing which `qrIds` are not okey and need to be fixed. This means we need to check this either inside `extractCSVDataFromContentImport` or as the first step inside `createAssetsFromContentImport` so we prevent any other entries being created unless this check is good. Then we can throw an error explaining which rows have the error and which id.

This also needs to be updated in the documentation. 

### Rules
IMPORTANT: You can only import `unclaimed` or `unlinked ` codes 

1. For `unclaimed` - there are not really any limitations we need to place. This should work directly
2. For `unlinked` - we can only allowed if the code is already claimed within the current workspace the user is trying to import to
3. For linked codes - we don't allow any imports.

### To Do:

- [x] Update template for importing
- [x] Add support for `qrId` column
- [x] Implement new logic that checks if the value of `qrId` column is a from a valid qr code.
- [x] Update import page with explanation of how importing of qr works

DonKoko:
Resolves #1290 

This needs to be tested. It is deployed to staging. 
We have quite a few error scenarios. Here is what we are already handling in the code. Let me know if I missed an edge case as that could be.

Error cases to test 
1. Duplicate codes - multiple assets having the same code
2. Non existent codes - codes that don't exist in shelf
3. Linked codes - codes that are already linked to an asset or kit
4. Belonging to other org - codes that are part of other organizations
5. Empty - doing an import where `qrId` is added only for some of the assets being imported
6. No qrId column - when the `qrId` column is completely missing from the import file.

### DIFF
--- a/app/components/assets/import-content.tsx
+++ b/app/components/assets/import-content.tsx
@@ -112,6 +112,35 @@ export const ImportContent = () => (
       <b>"cf:purchase date, type:date"</b>
     </div>
 
+    <h4 className="mt-2">Importing with QR codes</h4>
+    <div>
+      You also have the option to se a Shelf QR code for each asset. This is
+      very valuable if you already have Shelf QR codes printed and you want to
+      link them to the assets you are importing.
+      <br />
+      This feature comes with the following limitations:
+      <ul className="list-inside list-disc pl-4">
+        <li>
+          <b>Existing code</b> - the QR code needs to already exist in shelf
+        </li>
+        <li>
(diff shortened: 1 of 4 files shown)


carlosvirreira:
I have tested all cases. For me it can get published.

PR #1321, opened 2024-09-19
Pwa install prompt for mobile users

### LINKED ISSUE #1302: Feat: PWA install prompt
Prompt users to install PWA when visiting shelf on mobile.

hunar1312:
resolves #1302


app/components/layout/install-pwa-prompt-modal.tsx
```diff
@@ -0,0 +1,49 @@
+import { useLoaderData, useFetcher } from "@remix-run/react";
+import { AnimatePresence } from "framer-motion";
+
+import type { loader } from "~/routes/_layout+/_layout";
+import { Button } from "../shared/button";
+
+export default function InstallPwaPromptModal() {
```
DonKoko: Please adjust the export to not be default.

DonKoko (commented):
Looking good so far.

hunar1312:
Direct PWA installation from prompt modal now works for chromium browsers.

Pending things (input needed) :
1. Replacing the banner videoof install prompt modal
2. Need updated and appealing texts for the prompt
3. Instructions to display for non-chromium browser (ios, firefox, etc.) users on how to install PWA.

DonKoko:
@hunar1312 can you please fix the tests.

app/routes/api+/user.prefs.hide-install-pwa-prompt-modal.ts
```diff
@@ -0,0 +1,26 @@
...
+import { makeShelfError } from "~/utils/error";
+import { data, error } from "~/utils/http.server";
+
+export async function action({ context, request }: ActionFunctionArgs) {
+  const authSession = context.getSession();
+  const { userId } = authSession;
+
+  try {
+    const cookieHeader = request.headers.get("Cookie");
+    const cookie = (await userPrefs.parse(cookieHeader)) || {};
```
DonKoko, line 12: I see you are using the `userPrefs` cookie, but I think we need to create a different one, as we want it to expire after 14 days so we can remind the user again about it.

app/components/icons/library.tsx
```diff
@@ -1619,3 +1619,22 @@ export const ScanIcon = (props: SVGProps<SVGSVGElement>) => (
     />
   </svg>
 );
+
+export const InstallIcon = (props: SVGProps<SVGSVGElement>) => (
+  <svg
+    width="20"
```
DonKoko: We set width and height to 100% so we can control its size based on the parent

app/components/icons/library.tsx
```diff
@@ -1619,3 +1619,22 @@ export const ScanIcon = (props: SVGProps<SVGSVGElement>) => (
...
+    height="20"
+    viewBox="0 0 20 20"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    {...props}
+  >
+    <path
+      d="M6.66667 10L10 13.3333M10 13.3333L13.3333 10M10 13.3333V6.66667M6.5 17.5H13.5C14.9001 17.5 15.6002 17.5 16.135 17.2275C16.6054 16.9878 16.9878 16.6054 17.2275 16.135C17.5 15.6002 17.5 14.9001 17.5 13.5V6.5C17.5 5.09987 17.5 4.3998 17.2275 3.86502C16.9878 3.39462 16.6054 3.01217 16.135 2.77248C15.6002 2.5 14.9001 2.5 13.5 2.5H6.5C5.09987 2.5 4.3998 2.5 3.86502 2.77248C3.39462 3.01217 3.01217 3.39462 2.77248 3.86502C2.5 4.3998 2.5 5.09987 2.5 6.5V13.5C2.5 14.9001 2.5 15.6002 2.77248 16.135C3.01217 16.6054 3.39462 16.9878 3.86502 17.2275C4.3998 17.5 5.09987 17.5 6.5 17.5Z"
+      stroke="#667085"
+      stroke-width="1.75"
```
DonKoko: This is not valid jsx and will give errors. Please fix all stroke attributes

app/components/layout/install-pwa-prompt-modal.tsx
```diff
@@ -0,0 +1,71 @@
+import { useLoaderData, useFetcher } from "@remix-run/react";
+import { AnimatePresence } from "framer-motion";
+
+import type { loader } from "~/routes/_layout+/_layout";
+import { usePwaManager } from "~/utils/pwa-manager";
+import { Button } from "../shared/button";
+import { useRef } from "react";
```
DonKoko: Please fix import order

app/components/layout/install-pwa-prompt-modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+  let optimisticHideInstallPwaPrompt = hideInstallPwaPrompt;
+  if (fetcher.formData) {
+    optimisticHideInstallPwaPrompt =
+      fetcher.formData.get("pwaPromptVisibility") === "hidden";
+  }
+  const hidePwaPromptForm = useRef<HTMLFormElement | null>(null);
+
+  const { promptInstall } = usePwaManager();
+  return optimisticHideInstallPwaPrompt ? null : (
+    <AnimatePresence>
```
DonKoko, line 21: I think just like this its not doing anything. You need to add the motion.div inside it. We have some examples. Would be nice to have it animated.

app/components/layout/install-pwa-prompt-modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <div className="dialog-backdrop !items-end !bg-[#364054]/70">
+        <dialog
+          className="dialog m-auto h-auto w-[90%] pb-8 sm:w-[400px]"
+          open={true}
+        >
+          <div className="relative z-10  rounded-xl bg-white p-4 shadow-lg">
+            <video height="200" autoPlay loop muted className="mb-6 rounded-lg">
```
DonKoko: This doesn't auto play on my iphone. I think you need to add the inline attribute for it to work.

app/components/layout/install-pwa-prompt-modal.tsx
```diff
@@ -0,0 +1,71 @@
...
+              </Button>
+            )}
+            <fetcher.Form
+              ref={hidePwaPromptForm}
+              method="post"
+              action="/api/user/prefs/hide-install-pwa-prompt-modal"
+            >
+              <input type="hidden" name="pwaPromptVisibility" value="hidden" />
+              <Button type="submit" width="full" variant="secondary">
+                Skip
```
DonKoko: Skip for 2 weeks

app/routes/_layout+/_layout.tsx
```diff
@@ -148,6 +150,14 @@ export default function App() {
               )}
             </div>
             <Toaster />
+            <ClientOnly fallback={null}>
+              {() =>
+                window.matchMedia("(max-width: 640px)").matches &&
```
DonKoko: Lets move this check into a function with a clear name, so its more obvious what it does.

app/utils/pwa-manager.tsx
```diff
@@ -0,0 +1,101 @@
...
+ * 🤷‍♂️ Arc Browser, even if it's based on Chromium, doesn't support prompt to install.
+ * `prompt` never moves from pending to resolved.
+ *
+ * @returns the BeforeInstallPromptEvent if available
+ */
+export const usePwaManager = () => {
+  const context = useContext(PwaManagerContext);
+
+  if (context === null) {
+    throw new Error(`usePwaManager must be used within a PwaManagerProvider.`);
```
DonKoko, line 57: Wondering what that will do. Will it block the page or just throw an error in the console. Do u have an example?

DonKoko (changes requested):
- [ ] I can still scroll behind when the install prompt is visible

Added some comments in the files themselves. We will provide the content for both iOS and android asap.

app/utils/pwa-manager.tsx
hunar1312: Hmm, it will throw an error in console if usePwaManager() hook is used outside of the PwaManagerProvider context, maybe we can replace it with ShelfError @DonKoko ?

DonKoko: No need. Its fine like this. That is how it should work. THanks for confirming.

hunar1312:
> * [ ]  I can still scroll behind when the install prompt is visible
So to disable scroll, I will have to add css on body/html tag conditionally, for that I need to access it from DOM and to avoid optimization issues we were not doing DOM manipulation

carlosvirreira:
https://github.com/user-attachments/assets/33e5e2e1-0762-4387-a026-6a9bccdcbe99

Can you add this to test hunar? curious if its even helpful.

carlosvirreira:
If the other assets are not ok. Here are alternatives.

PWA IOS chrome

PWA IOS

DonKoko (approved):
Thanks for the work @hunar1312

PR #1360, opened 2024-10-16
fix: overflow issue of kit name in asset overview page

### LINKED ISSUE #1352: [Bug]: Front end alignment issues
### Contact Details

_No response_

### What happened?

Text overflowing when long kit name

Screenshot 2024-10-15 at 17 27 49

On asset page
Screenshot 2024-10-15 at 17 29 12

If text is too large for container area, fine to add (....)

### Version

Hosted (app.shelf.nu)

Safari

### Relevant log output

_No response_

rajeshj11:
closes #1352

fix the kit name alignment issue


DonKoko (changes requested):
This works good, but we still have the other case.
If you add the kit with the very long name to a booking, and then check the booking page you can see the issue:
image
image
You can see in the second screenshot. Its the link that is causing the issue. Should be `items-start` and truncated as well.

rajeshj11:
> This works good, but we still have the other case. If you add the kit with the very long name to a booking, and then check the booking page you can see the issue: image image You can see in the second screenshot. Its the link that is causing the issue. Should be `items-start` and truncated as well.

will work on it

DonKoko (approved):

PR #1364, opened 2024-10-18
feat: add assest and kits to exisiting booking

### LINKED ISSUE #1245: Feature: Add assets to existing booking from asset page
Currently its only possible to create a new booking via the asset page. 
We should allow users to add to existing booking as well.

rajesh-jonnalagadda:
closes #1245


rajeshj11:
@DonKoko please review

app/routes/_layout+/kits.$kitId.tsx
DonKoko, line 1: This route should be a child of the asset overview route so when it opens you can still see the asset behind.

app/components/assets/actions-dropdown.tsx
```diff
@@ -275,4 +277,55 @@ const ActionsDropdown = () => {
   );
 };
 
+export const BookActionsDropDown = () => {
```
DonKoko: I am very very confused by the placement of this component. Why not in the file with the other relevant components or maybe in its own file?

app/utils/booking-drop-down-actions.tsx
DonKoko, line 1: This is a component file. Should not be placed here but rather in the correct folder in components.

app/utils/booking-drop-down-actions.tsx
```diff
@@ -0,0 +1,195 @@
...
+}) => {
+  const {
+    ref: dropdownRef,
+    defaultApplied,
+    open,
+    defaultOpen,
+    setOpen,
+  } = useControlledDropdownMenu();
+
+  const disabled_reason = links.find((link) => link.disabled)?.disabled_reason;
```
DonKoko: camelCase please

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+      userId: authSession?.userId,
+      request,
+      entity: PermissionEntity.booking,
+      action: PermissionAction.create,
+    });
+
+    const formData = await request.formData();
+    const { assetIds, bookingId, tab, kitIds } = parseData(
+      formData,
+      z.object({
```
DonKoko: Please make the code more modular and extract this into a Schema variable. We only do it inline when its super simple schema and this is not the case here.

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+
+  return (
+    <Form method="post">
+      <div>
+        <div className="mb-4 inline-flex items-center justify-center rounded-full border-8 border-solid border-primary-50 bg-primary-100 p-2 text-primary-600">
+          <BookingExistIcon />
+        </div>
+        <div className="mb-5">
+          <h3>Add to Existing Booking</h3>
+          <p>You can only add an asset to bookings that</p>
```
DonKoko: No need to split this in 2 paragraphs.

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+          <p>are in Draft or Reserved State.</p>
+        </div>
+        {tab === "assets" &&
```
DonKoko: I am a bit confused by the variable name `tab`. There are no tabs in this UI. When someone just looks at this code this variable name will make it very confusing for them on why and what the UI is. Please use something more descriptive of the value that is being stored.

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+  if (allAssets.some((asset) => asset.status === "CHECKED_OUT")) {
+    throw new Error(
+      "One or more assets are already checked out in the kit, so they cannot be added to the booking."
+    );
+  }
+
+  return allAssets.map((asset) => asset.id);
+};
+
+const getAvailableAssets = async (assetIds: string[]): Promise<string[]> => {
```
DonKoko: This function should not be in this file.
Also why not just use `getAssets`?

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+    booking = bookingDetails;
+  } else {
+    throw new ShelfError({
+      cause: null,
+      message: "Invalid operation. Please contact support.",
+      label: "Booking",
+    });
+  }
+
+  if (finalAssetIds.length === 0) {
```
DonKoko: This should not happen and should be controlled at an earlier step of the process.
To start with, the button should be disabled if the kit has no assets inside it.

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    const { finalAssetIds, bookingInfo } = await processBooking(
+      tab,
+      bookingId,
+      assetIds,
+      kitIds
+    );
+
+    function intersected(arr1: string[], arr2: string[]): boolean {
```
DonKoko: Please place this in the correct place. This is a very useful function. Placing it here only within this loader seems like a huge waste.

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+              key={kitId}
+              type="hidden"
+              name={`kitIds[${i}]`}
+              value={kitId}
+            />
+          ))}
+        <input key="tab" type="hidden" name="tab" value={tab} />
+
+        <div className=" relative z-50 mb-2">
+          <DynamicSelect
```
DonKoko: I would have never done this with the DynamicSelect, would have just made something simple. This works now so its fine, maybe a bit of an overkill but we will see. My main concern is that you had to adjust the model-filters which now means we are required to test like 10 different features.

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+              displayTo: formatDate(item?.metadata?.to),
+              status: item?.metadata?.status,
+            })}
+            renderItem={(item: any) =>
+              isValidBooking(item) ? (
+                <div
+                  className="flex flex-col items-start gap-1 text-black"
+                  key={item.id || item.name}
+                >
+                  <h3 className="max-w-[250px] truncate">{item.name}</h3>
```
DonKoko: This is way too huge. Should be just a div with 16px font size and semibold and the dates below should be font size 14px.

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,473 @@
...
+        ],
+      }
+    );
+  } catch (cause) {
+    const reason = makeShelfError(cause, { userId });
+    throw json(error(reason), { status: reason.status });
+  }
+}
+
+const getBookingDetails = async (
```
DonKoko: I see you added 4 functions in this file that I feel are just repeating existing stuff and can be managed way better.

1. They should not be in this file. Each function should be in the relevant service file for the model it refers to
2. All function should have a try/catch.

rajeshj11:
@DonKoko can you please review

rajeshj11:
@DonKoko does it still need work? please let me know.

note: As we already discussed in Discord.
if you think the pr is overkill. we can simply close it. no issues

app/components/shared/actions-drop-down.tsx
DonKoko, line 1: `dropdown` instead of `drop-down` in file name

app/modules/asset/service.server.ts
```diff
@@ -2824,3 +2824,29 @@ export async function relinkQrCode({
...
+    });
+    if (selectedAssets.some((asset) => asset.kit)) {
+      throw new ShelfError({
+        cause: null,
+        message: "Cannot add assets that belong to a kit.",
+        label: "Booking",
+      });
+    }
+    return selectedAssets.map((asset) => asset.id);
+  } catch (err) {
```
DonKoko: Please use `cause`. that is what we use everywhere. Also the message in the catch should be dynamic. something like this:
```ts
message: cause?.message ? cause.message : "some message"
```

app/modules/booking/service.server.ts
```diff
@@ -1536,3 +1536,36 @@ export async function addScannedAssetsToBooking({
...
+        cause: null,
+        message: "Booking is not in Draft or Reserved status.",
+        label: "Booking",
+      });
+    }
+    return booking;
+  } catch (cause) {
+    throw new ShelfError({
+      cause,
+      message: "Something went wrong while getting existing booking details.",
```
DonKoko: Dynamic message from cause.

app/modules/booking/service.server.ts
```diff
@@ -1536,3 +1536,36 @@ export async function addScannedAssetsToBooking({
     });
   }
 }
+
+export async function getExistingBookingDetails(bookingId: string) {
+  try {
+    const booking = await db.booking.findFirst({
```
DonKoko: Usually we use `findUniqueOrThrow` as this automatically handles 404s. Is there some specific reason to not use it here.

app/modules/kit/service.server.ts
```diff
@@ -1156,3 +1156,31 @@ export async function updateKitQrCode({
...
+        message:
+          "One or more assets are already checked out in the kit, so they cannot be added to the booking.",
+        label: "Booking",
+      });
+    }
+    return allAssets.map((asset) => asset.id);
+  } catch (cause) {
+    throw new ShelfError({
+      cause: cause,
+      message: "Something went wrong while getting available assets.",
```
DonKoko: Dynamic message from cause./

app/routes/_layout+/assets.$assetId.tsx
```diff
@@ -246,35 +246,7 @@ export default function AssetDetailsPage() {
...
-                      instead.
-                    </>
-                  ),
-                }
-              : false
-          }
-        >
-          Book
-        </Button>
+        <BookingActionsDropDown />
```
DonKoko: `BookingActionsDropdown` instead of `BookingActionsDropDown`

app/routes/_layout+/bookings.update-existing.tsx
```diff
@@ -0,0 +1,431 @@
...
+        message: "No assets available.",
+        label: "Booking",
+      });
+    }
+
+    return {
+      finalAssetIds,
+      bookingInfo: booking,
+    };
+  } catch (err) {
```
DonKoko: Use `cause` please. Also use the message from the cause if it exists. That way the user knows what actually happened.

app/routes/api+/model-filters.ts
```diff
@@ -68,11 +72,23 @@ export async function loader({ context, request }: LoaderFunctionArgs) {
     const { name, queryKey, queryValue, selectedValues, ...filters } =
       parseData(searchParams, ModelFiltersSchema);
 
-    const where: Record<string, any> = {
+    let where: Record<string, any> = {
       organizationId,
       OR: [{ id: { in: (selectedValues ?? "").split(",") } }],
     };
 
+    const genericKeys = ["deletedAt"];
```
DonKoko: I am having trouble understanding this. Can you explain why you are doing it and add some annotations. Then we can maybe see if there is a better way to manage it.

app/components/assets/booking-actions-dropdown.tsx
```diff
@@ -0,0 +1,61 @@
...
+      label: "Add to existing booking",
+      icon: "booking-exist",
+      disabled,
+      disabledReason: reason,
+      to: `/bookings/update-existing?indexType=assets&id=${asset.id}`,
+    },
+  ] as CustomLink[];
+
+  return (
+    <div className="actions-dropdown flex">
```
DonKoko, line 52: Although this works, that is now what we wanted to achieve. Please look at the example we gave.
It is not a singular dropdown, its a button group where the left item is the main button and the right item is a dropdown trigger with the different options. Let me know if you need further examples.

DonKoko (changes requested):
Already way better but still some changes pending.

app/routes/_layout+/bookings.update-existing.tsx
DonKoko, line 1: File name is too generic. It needs to be more specific to what it does. Also this should be a child of `assets.$assetId.overview`. Please name the file: `assets.$assetId.overview.add-to-existing-booking`

Then you would need to create also a file in `kits.$kitId.add-to-existing-booking`

And last you can create a function for the action, loader & component so you can use them in the 2 different places.

DonKoko (changes requested):
Files are still not as I like them

rajesh-jonnalagadda:
@DonKoko Could you please review this when you have some time? It’s been pending for a while.

rajesh-jonnalagadda:
> @DonKoko Could you please review this when you have some time? It’s been pending for a while.

if not needed please close it. i can understand

DonKoko (commented):
I made a lot of changes myself to optimize, cleanup and change stuff that were not used as we usually do. We are now in the process of testing this on our staging env.

carlosvirreira:
From my test / Observation:

- If an asset is marked as unavailable for booking. Please make the dropdown button disabled. "The asset is marked as unavailable for booking - please adjust this to proceed"
- 
Screenshot 2024-11-05 at 14 49 54

DonKoko:
Thanks @carlosvirreira . I will resolve it.

DonKoko (approved):

PR #1392, opened 2024-11-05
Improvement to Calendar View

### LINKED ISSUE #1350: [Feature request]: Improvement to Calendar View
### Contact Details

_No response_

I use the calendar a lot and I want some changes made to improve quality of life using shelf.

As a quality of life improvement it would be great to have an option to scale calendar view to see more data on the grid. Maybe toggle different views like WEEK, MONTH, DAY. It would significantly improve our workflow.

Also the calendar landscape view on my iphone is a bit glitchy .. sometimes it just does not switch from list view to calendar view.

_No response_

### Additional context

_No response_

Bhavyajain21:
fixes #1350

### DIFF
--- a/app/routes/_layout+/calendar.tsx
+++ b/app/routes/_layout+/calendar.tsx
@@ -79,16 +79,6 @@ export const loader = async ({ request, context }: LoaderFunctionArgs) => {
       entity: PermissionEntity.booking,
       action: PermissionAction.read,
     });
-    if (isPersonalOrg(currentOrganization)) {
-      throw new ShelfError({
-        cause: null,
-        title: "Not allowed",
-        message:
-          "You cannot use bookings in a personal workspaces. Please create a Team workspace to create bookings.",
-        label: "Booking",
-        shouldBeCaptured: false,
-      });
-    }
 
     const header = {
       title: `Calendar`,
@@ -131,7 +121,8 @@ export default function Calendar() {
   const [calendarView, setCalendarView] = useState(
     isMd ? "dayGridMonth" : "listWeek"
   );
-
+  const disabledButtonStyles =
+    "cursor-not-allowed pointer-events-none bg-gray-50 text-gray-800";
   const calendarRef = useRef<FullCalendar>(null);
   const ripple = useRef<HTMLDivElement>(null);
 
@@ -147,7 +138,7 @@ export default function Calendar() {
     updateTitle();
   };
 
-  const updateTitle = () => {
+  const updateTitle = (viewMode = calendarView) => {
     const calendarApi = calendarRef.current?.getApi();
     if (calendarApi) {
       const currentDate = calendarApi.getDate();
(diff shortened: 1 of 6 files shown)


Bhavyajain21:
image

DonKoko (changes requested):
I have done some testing, there are the things I noticed so far:
- [x] There should be an active state for the currently active view. You can check the simple & advanced buttons on the asset index for example on how the active state should look. It should basically be the same as the current hover effect
Screenshot 2024-11-06 at 15 18 23

- [ ] The title should be updated when the mode is changed. Here is how it should work:
Screenshot 2024-11-06 at 15 19 43

1. Month - keep as is
2. Week - Keep as is + add a subtitle saying "Week ${week start day} - ${week end date}"
3. Day - Prepend the date to the month + year. Add subtitle with the week day.


- [ ] When we are in day view, dont highlight in orange the whole day
- [x] In day view, when there are multiple events the UI breaks:
Screenshot 2024-11-06 at 15 24 14

- [ ] in week view, having multiple events looks broken

Screenshot 2024-11-06 at 15 25 17

carlosvirreira:
@DonKoko if you pull this, could you send me a screenshot(s)?

app/routes/_layout+/calendar.tsx
```diff
@@ -199,6 +224,19 @@ export default function Calendar() {
     }
   };
 
+  const handleViewChange = (view: string) => {
+    setCalendarView(view);
+    const calendarApi = calendarRef.current?.getApi();
+    calendarApi?.changeView(view);
+    if (view === "dayGridMonth") {
```
DonKoko: why not just:
`updateTitle(view)`. The `if` block seems redundant.

app/routes/_layout+/calendar.tsx
```diff
@@ -127,7 +129,11 @@ export default function Calendar() {
   const [calendarSubtitle, setCalendarSubtitle] = useState(
     isMd ? undefined : `${startingDay} - ${endingDay}`
   );
-
+  const [calendarView, setCalendarView] = useState(
```
DonKoko, line 136: This is problematic. Let me explain.
This will run every time the component re-renders.
So if you have the following scenario:

1. Switch to day view
2. Resize window
3. Component re-renders
4. You see month view but on the buttons day view is selected.

This needs to be updated on how we set the default one. You have have a `useEffect` that runs on render that updates the value based on the screen size + the current selected value.

app/routes/_layout+/calendar.tsx
```diff
@@ -217,27 +255,65 @@ export default function Calendar() {
...
+                  variant="secondary"
+                  className="p-[0.75em] text-gray-500"
+                  onClick={() => handleNavigation("next")}
+                >
+                  <ChevronRightIcon />
+                </Button>
+              </ButtonGroup>
+            </div>
+
             <ButtonGroup>
```
DonKoko: Should not be visible on mobile. Only view allowed on mobile is list.

DonKoko (changes requested):

app/routes/_layout+/calendar.tsx
Bhavyajain21: Fixed!

app/routes/_layout+/calendar.tsx
Bhavyajain21: This is fixed with a small change

app/routes/_layout+/calendar.tsx
Bhavyajain21: Fixed

DonKoko (changes requested):
- [ ] currently, when hovering an event the info shows up only when hovering the name. Should be for hovering the whole block
image
- [ ] currently single day events only have a bg-color on hover. This should be the case only for monthly view. In the other views, they should always have the events
- [x] The dot for 1 day events should only be visible on monthly view:
image
- [x] On week view, the title is still not correct. Please check how we do it on mobile. It should be 1:1 with how we do it there
- [x] Currently when you have 2 events overlapping on the same time, they are kinda squashed next to eachother. This makes it ugly and very condensed. What we want to do is make it like it is in the example on full calendar or as a matter of fact like in any calendar app. My example I use is the apple calendar app.
**Currently looks like:**
image

**Should look like:**
image

Please check the full calendar docs for this. Just quickly searched and found this: https://fullcalendar.io/docs/slotEventOverlap. You have set this to false, not sure why.

- [x] When I go on day view and I go to "Today" the whole day is still highlighted. This should not be the case for the day view, only for month/week.
- [x] When events are overlapping, we should figure out how to add a shadow to the events that are on top of other events. From what I can see FullCalendar adds a z-index to `fc-timegrid-event-harness` so we could somehow style based on this. You can just use one of the shadows we have available in tailwind.

DonKoko:
I am gonna add another comment related to the overlapping because I found something cool I wanted to share. Take a look at this video:

https://github.com/user-attachments/assets/255b52bc-184d-4116-87c8-0217d4257e19

DonKoko (changes requested):
- [ ] When the event is very short there is not enough space to show the text. I am wondering if FullCalendar has some way to manage this. Could you check it out please
image
image

- [ ] 1 day event background color has been added also to month view. The idea was to add it only to the other views (week & day)

PR #1401, opened 2024-11-07
fix: handing the case of checking out a booking after its end period has p…

### LINKED ISSUE #1369: [Bug]: Overdue status revision
### Contact Details

_No response_

### What happened?

If a user has made a reservation in the past. And the current date is after the stated start/end date of the booking. 

Shelf allows you to do a check-out anyways. This is in line with our focus on flexibility. The status shows 'Checked out'. 

Allowing users to check out items on a reservation, after the stated start and end period is ok. However, it would be best if the status of items and the booking would go to 'overdue' and the 'overdue' mail gets sent. 

### Version

Hosted (app.shelf.nu)

Safari

### Relevant log output

_No response_

DonKoko:
…assed

fixes #1369


DonKoko:
@carlosvirreira deployed to staging and ready to test. Please also test the classic use cases just to make sure nothing broken. I already did the tests and all seems good, but its important to get a second test. Thanks

DonKoko:
> Now the status goes to overdue as expected. Mail did not reach my inbox though (overdue mail) > [carlos@shelf.nu](mailto:carlos@shelf.nu)

This is on purpose. Let me explain the logic, considering the 2 actors are the ADMIN and CUSTODIAN.

1. Checking out the booking is the ADMIN's responsibility
2. The "Overdue reminder" email is being sent to the CUSTODIAN

So in this case we have the ADMIN being later at what they should have done earlier and then the CUSTODIAN receiving a notification warning them that their booking is overdue and they are late. 
That is the reason I have removed the email in this case.

If I am missing something, please let me know. Lets get the best solution.

carlosvirreira:
> > Now the status goes to overdue as expected. Mail did not reach my inbox though (overdue mail) > [carlos@shelf.nu](mailto:carlos@shelf.nu)
> 
> This is on purpose. Let me explain the logic, considering the 2 actors are the ADMIN and CUSTODIAN.
> 
> 1. Checking out the booking is the ADMIN's responsibility
> 2. The "Overdue reminder" email is being sent to the CUSTODIAN
> 
> So in this case we have the ADMIN being later at what they should have done earlier and then the CUSTODIAN receiving a notification warning them that their booking is overdue and they are late. That is the reason I have removed the email in this case.
> 
> If I am missing something, please let me know. Lets get the best solution.

Good points. I see the logic here brother. Then, I have no further observations.

PR #1428, opened 2024-11-15
feat: change user email

DonKoko:
- created a modal for changing email
- implemented action for sending email verification code & verifying code
- created `updateEmail` function that updates the email in `public.User` as well as in `auth` schema
- created custom email to be sent with the verification code
- kill all "other" sessions when changing email, so if the user is logged in on a compromised device, they will be logged out.

### DIFF
--- a/app/components/assets/booking-actions-dropdown.tsx
+++ b/app/components/assets/booking-actions-dropdown.tsx
@@ -41,7 +41,7 @@ export default function BookingActionsDropdown() {
       disabled,
       label: "Create new booking",
       icon: "bookings",
-      to: `/bookings/new?assetId=${asset.id}`,
+      to: "overview/create-new-booking",
     },
     {
       indexType: "asset",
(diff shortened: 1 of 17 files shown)


DonKoko:
@carlosvirreira this is being deployed to staging server now. Needs some detailed testing. Here are some cases to test. I will go over them again on staging as well.

- [ ] success case - successfully change email
- [ ] try to change email to an already taken email
- [ ] try to change email to the same as the current email
- [ ] type in different emails in the 2 fields
- [ ] enter wrong verification code
- [ ] Open 2 sessions in 2 browsers and change the email in 1 browser. Then check the other one and make sure the session is killed when you refresh the page

carlosvirreira:
Hi @DonKoko - this has been tested and it looks good. Only observation.

**Enter Wrong Verification Code**
> Outcome: Displayed error "Invalid or expired verification code"
> Issue found: Clicking 'Cancel' does nothing. so you are stuck there. You have to refresh page. (In case you submitted a change to a mail with typo [[giga-edge]]) 

//

Suggestion / Observations (Not requested)

- Password reset page: For me it was confusing, do i need to add the former password or just the new password. Consider renaming the 'Password' form label to 'New Password'

rphlmr (approved):
Lgtm!

rphlmr (dismissed):
Nice catch!

Suggestion:
What about blocking the `ChangeEmailForm` pencil button if the email is an SSO domain email?
With a popover displaying, "The email's domain is not allowed for security reasons. For more information, please get in touch with support."

PR #1442, opened 2024-11-21
Tag trim name

HarryEMartland:
This change trims the tag name when creating or updating.

When using a mobile device with autocomplete spaces are added at the end which the user may not be aware of. This change solves the problem by trimming the name before saving.


DonKoko (approved):

DonKoko (approved):

carlosvirreira:
ty for the contribution @HarryEMartland

PR #1448, opened 2024-11-22
Mobile Camera Image Upload

HarryEMartland:
this change prompts the user with their camera on mobile devices rather than file upload when creating or updates assets

adds the capture attribute to the input field see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture for reference

### DIFF
--- a/app/components/assets/form.tsx
+++ b/app/components/assets/form.tsx
@@ -12,6 +12,7 @@ import { z } from "zod";
 import { updateDynamicTitleAtom } from "~/atoms/dynamic-title-atom";
 import { fileErrorAtom, validateFileAtom } from "~/atoms/file";
 import type { loader } from "~/routes/_layout+/assets.$assetId_.edit";
+import { ACCEPT_SUPPORTED_IMAGES } from "~/utils/constants";
 import type { CustomFieldZodSchema } from "~/utils/custom-fields";
 import { mergedSchema } from "~/utils/custom-fields";
 import { isFormProcessing } from "~/utils/form";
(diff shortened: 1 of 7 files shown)


DonKoko:
hey @HarryEMartland. Thanks for the PR. 
First of all, cool API. Didn't know it existed.

I am not sure if this is an improvement as I am not convinced this is the best UX. I know for a fact that we have users that sometimes take a picture and sometimes use an image from their device. Adding this capture attribute makes it impossible to use an image from their device.

I double checked the default behavior without the capture and I can see this:
New asset  shelf nu

This is way more flexible as it allows both options. Does it not work like this on Android? I would love to know more about the motivation for this PR.

HarryEMartland:
This took me on a bit of a goose chase 🤣 

TLDR there looks like there may be a bug in Android 14 so if you filter the allowed files it does not prompt the user for the camera. I found [this stack overflow](https://stackoverflow.com/a/79163998/1894472) post where others are having this issue which includes a workaround. I have updated the MR to include this workaround and removed the capture attribute.

This is what Android 14 (Pixel 6) currently looks like (cut off but existing photos are shown below)
Screenshot_20241125-173402

And after this change:
Screenshot_20241125-173407

I agree it would be best to give the user both options but as you can see on Android 14, only selecting existing images was available. 

I am just using Shelf at home but as I could not take an image in the app I was not adding photos as it was too much effort. With this change it will be a lot easier.

HarryEMartland:
Just had a quick search in the code base and looks like there will be other affected uploads. If you are happy with this MR I could do some more MRs for the the other inputs. Maybe pull out the values to a constant? (those that are images)

DonKoko (changes requested):
hey @HarryEMartland , now it makes sense. Weird bug but glad you found a solution.
And indeed I agree, we should implement this in the other places as well. I like the suggestion of putting the `accept` value in a constant and using it wherever relevant.

These are the places we use `type="file"`. Most of them are for images so we would need to put the constant there.
Screenshot 2024-11-26 at 15 09 37

DonKoko (approved):

PR #1453, opened 2024-11-26
feat: group required and optional

### LINKED ISSUE #1447: [Feature request]: Form Asset Create / Edit - Updated Behaviour
### Contact Details

_No response_

As a facilities manager, I have to create a workflow for making an inventory of 1000s of assets. 

I will be using the help of not super technical colleagues. So I try to make the process as simple as possible. In my eyes, if they add the required fields from custom fields this is good. 

Currently on shelf the fields on asset create are grouped alphabetically. Ideally they are bundled on required/optional. 

Group the custom fields by required and optional on asset create/edit form.

Group by required

_No response_

### Additional context

_No response_

Bhavyajain21:
/fixes #1447
/claim #1447


app/components/assets/custom-fields-inputs.tsx
```diff
@@ -196,6 +196,25 @@ export default function AssetCustomFields({
       );
     },
   };
+  const requiredFields = customFields.filter((field) => field.required);
+  const optionalFields = customFields.filter((field) => !field.required);
+  type JsonifyObject<T> = {
```
DonKoko: This is not needed. We already have ways to solve this. You can look for different examples in the platform. Take a look at `SerializeFrom`. Import it from remix, it will help you with this issue.

app/components/assets/custom-fields-inputs.tsx
```diff
@@ -208,25 +227,49 @@ export default function AssetCustomFields({
           Manage custom fields
         </Link>
       </div>
+      <div className="border-b pb-5">
```
DonKoko: This whole div should be only rendered if there are any required fields. Same for the optional fields.

app/components/assets/custom-fields-inputs.tsx
```diff
@@ -208,25 +227,49 @@ export default function AssetCustomFields({
...
+        <h3 className="mb-4 text-[16px] font-semibold">Required Fields</h3>
```
DonKoko: No need to use an H3 and add styles on top. Just use h5 please.

Bhavyajain21: Fixed!

app/components/assets/custom-fields-inputs.tsx
Bhavyajain21: Fixed!

app/components/assets/custom-fields-inputs.tsx
Bhavyajain21: fixed!

DonKoko (approved):

PR #1546, opened 2024-12-30
fix: prisma schema mismatch with DB

DonKoko:
We created some indexes with a manual migration, so when we create a new migration they were getting automatically dropped as they are not reflected in schema.prisma.  This PR resolves this my adding those indexes to the schema so they dont get dropped every time.

- added all direct table indexes to schema. Those indexes were defined with a custom migration so we have to add them to the schema to prevent issues when generating new migrations
- fixed small issue with welcome layout
- implemented post-migration scripts that run after generating a new migration
- added a script protected-indexes which protects certain indexes from being dropped as they are not defined in `schema.prisma` due to prisma limitations - https://github.com/prisma/prisma/issues/12914
- created documentation about how we are protecting indexes and what it does
- updated package.json to make sure that post-migrations scripts are ran at the correct time

### DIFF
--- a/app/database/migrations/20241230135956_test_mg/migration.sql
+++ b/app/database/migrations/20241230135956_test_mg/migration.sql
@@ -1,5 +1,0 @@
--- DropIndex
-DROP INDEX "_AssetToBooking_Asset_idx";
-
--- DropIndex
-DROP INDEX "_AssetToTag_asset_idx";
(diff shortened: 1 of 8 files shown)


DonKoko (commented):
You did a good job:
image

carlosvirreira:
+1 Medal to @DonKoko 

Screenshot 2024-12-30 at 16 55 36

PR #1553, opened 2025-01-02
fix: change pdf generation to client side for better performance

DonKoko:


DonKoko:
@carlosvirreira this is quite tricky but its getting there. I am deploying it to staging so I can get some feedback.

PR #1554, opened 2025-01-02
feat: adds row level security to _AssetToTag table

HarryEMartland:
I have been getting warnings saying row level security is not enabled for `_prisma_migrations` and `_AssetToTag`. I have added a migration to enable it for `_AssetToTag` however I was struggling with `_prisma_migrations` I think due to the table not existing in the "shadow database". I have manually enabled it and things seem to still work so maybe we could add it to the docs or I could play so more with getting a migration working.

Screenshot 2025-01-02 at 16 42 54


DonKoko:
hey @HarryEMartland . Thanks for the PR. You are right about the RLS for __prisma_migrations. I also enabled that from the UI. As for `_AssetToTag` I am not sure what happened but it seems like we forgot it. Thanks for spotting it.

DonKoko (approved):

PR #1587, opened 2025-01-18
add email queue to handle email sending failures

mahendraHegde:
- ability to enqueue failed emails and retry them later
  - **why not push all emails to queue**? because, then we need need to increase pg-boss polling frequency(more queries to postgress every minute), if not users will always have to wait for X minute to receive any email, some tokens/OTPs will have lifetime of X minutes 
- We dont await  `triggerEmail` call inside `sendEmail` anymore, so that any user facing actions can finish without waiting for email send to complete, this is actually a tradeoff, because if email sending fails(1st attempt + 5 retries using queues) user never will be warned, but the case is very unlikely and the benefit supersedes the tradeoff.

### DIFF
--- a/app/emails/email.worker.server.ts
+++ b/app/emails/email.worker.server.ts
@@ -1,0 +1,59 @@
+import { transporter } from "~/emails/transporter.server";
+import { QueueNames, scheduler } from "~/utils/scheduler.server";
+import type { EmailPayloadType } from "./types";
+import { SMTP_FROM } from "../utils/env";
+import { ShelfError } from "../utils/error";
+
+export const registerEmailWorkers = async () => {
+  await scheduler.work<EmailPayloadType>(
+    QueueNames.emailQueue,
+    { newJobCheckIntervalSeconds: 60 * 3, teamSize: 2 },
+    async (job) => {
+      await triggerEmail(job.data);
+    }
+  );
+};
+
+export const triggerEmail = async ({
+  to,
+  subject,
+  text,
+  html,
+  from,
+  replyTo,
+}: EmailPayloadType) => {
+  try {
+    // send mail with defined transport object
+    await transporter.sendMail({
+      from: from || SMTP_FROM || `"Shelf" <updates@emails.shelf.nu>`, // sender address
+      replyTo: replyTo || "support@shelf.nu", // reply to
+      to, // list of receivers
+      subject, // Subject line
+      text, // plain text body
+      html: html || "", // html body
+    });
+  } catch (cause) {
+    throw new ShelfError({
+      cause,
+      message: "Unable to send email",
+      additionalData: { to, subject, from },
+      label: "Email",
+    });
+  }
+
+  // verify connection configuration
+  // transporter.verify(function (error) {
+  //   if (error) {
+  //     // eslint-disable-next-line no-console
+  //     console.log(error);
+  //   } else {
+  //     // eslint-disable-next-line no-console
+  //     console.log("Server is ready to take our messages");
+  //   }
+  // });
+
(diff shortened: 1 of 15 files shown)


app/emails/mail.server.ts
```diff
@@ -1,104 +1,39 @@
...
-  /** Override the default sender */
-  from?: string;
+import { Logger } from "~/utils/logger";
+import { QueueNames, scheduler } from "~/utils/scheduler.server";
+import { triggerEmail } from "./email.worker.server";
+import type { EmailPayloadType } from "./types";
+
+export const sendEmail = (payload: EmailPayloadType) => {
+  // attempt to send email, push to the queue if it fails
+  triggerEmail(payload).catch((err) => {
```
mahendraHegde, line 8: we dont `await`, so users dont have to wait for long running email api calls.

app/modules/booking/service.server.ts
```diff
@@ -1341,7 +1341,7 @@ export async function bulkDeleteBookings({
     }));
 
     // Send emails with rate limiting
```
DonKoko: Can you please remove the comment.

app/emails/email.worker.server.ts
```diff
@@ -0,0 +1,59 @@
+import { transporter } from "~/emails/transporter.server";
+import { QueueNames, scheduler } from "~/utils/scheduler.server";
+import type { EmailPayloadType } from "./types";
+import { SMTP_FROM } from "../utils/env";
+import { ShelfError } from "../utils/error";
+
+export const registerEmailWorkers = async () => {
+  await scheduler.work<EmailPayloadType>(
+    QueueNames.emailQueue,
+    { newJobCheckIntervalSeconds: 60 * 3, teamSize: 2 },
```
DonKoko: Can you please add a comment on how this relates to our limit per second of sending. Maybe just place the whole calculation we discussed on discord so we have it in the code for future reference.

app/modules/booking/email-helpers.ts
```diff
@@ -156,44 +156,35 @@ export const checkinReminderEmailContent = ({
     )}.`,
   });
 
-export async function sendCheckinReminder(
+export function sendCheckinReminder(
```
DonKoko, line 158: Okey so based on this we have to review every place where we call sendEmail and replace `await` with `void`

app/modules/booking/email-helpers.ts
```diff
@@ -156,44 +156,35 @@ export const checkinReminderEmailContent = ({
...
-    });
-  } catch (cause) {
-    throw new ShelfError({
-      cause,
-      message: "Something went wrong while sending the checkin reminder email",
-      additionalData: { booking },
-      label: "Booking",
-    });
-  }
+  sendEmail({
```
DonKoko, line 163: TS error here. If we dont await we have to always void.

DonKoko (changes requested):
Added some small comments. There is some ESlint warnings and also TS errors everywhere where we dont use `void` instead of `await` for sending emails.

app/modules/booking/email-helpers.ts
mahendraHegde: but the function is not async, what error you got? `npm run typecheck` runs fine for me.

app/modules/booking/email-helpers.ts
mahendraHegde: same as other, the sendEmail is not async anymore

mahendraHegde (commented):
I have addressed review comments

DonKoko (approved):
PR looks good. I have tested all emails and I think we have it handled. Lets see what comes up after release but I am quite confident.

I will approve it now and will release monday.

PR #1613, opened 2025-01-29
feat: multi domain support for SSO

DonKoko:


carlosvirreira:
do we have to update documentation for this? https://www.shelf.nu/knowledge-base/sso-with-google-workspace

PR #1628, opened 2025-02-05
Feature: zxing/wasm scanner

DonKoko:
- [x] There is a moment after permissions are given when the video initiates together with the canvas and the module. We need some loading state for that as it can be slow on slower devices and it looks weird.


carlosvirreira:
Feedback: 

- [ ] Action buttons go over select-able elements (custodian select) - prevents from completing action.
IMG_0013  IMG_0014

- [ ] Initiating a booking (Mobile) via calendar or booking index - modal has overflow. 
IMG_0015

- [ ] After using Scan to Add  - Switching from Draft to reserved status makes status badge act up (on all states, reserved, checked out, checked in (complete))IMG_0016

DonKoko:
@carlosvirreira all those issues are in live and not really related to the scanner. I have made a different issue to resolve them:  #1638

PR #1640, opened 2025-02-11
refactor: wasm/scanner - manage state and re-renders in a better way

DonKoko:


DonKoko:
For some reason the last 2 commits broke the scanner but just on production servers. Works fine on local but on staging it just doesnt detect the code at all. The commits that broke it are [27b6af8](https://github.com/Shelf-nu/shelf.nu/pull/1640/commits/27b6af83f8e04bb1e6fea650dfac8f30694ee898) or [3766b27](https://github.com/Shelf-nu/shelf.nu/pull/1640/commits/3766b27432e6dc5f5996d9ca260092d007104e5d). Needs further debugging.

Issue #1681, opened 2025-03-14
[Feature request]: Ability to change SMTP port


iDiegoNL:
### Is your feature request related to a problem? Please describe?

Currently, the SMTP port is hardcoded to `465`. This causes issues if the mail host (in our case AWS SES) uses a different port.

### Describe the solution you'd like

The ability to customize the SMTP port via an optional env variable, e.g. `SMTP_PORT`

PR #1683, opened 2025-03-14
Add env variable for the SMTP port

iDiegoNL:
Closes https://github.com/Shelf-nu/shelf.nu/issues/1681

This implementation should be fully backwards-compatible, since it falls back to the old default (port 465) if the env variable is not present.


carlosvirreira:
@iDiegoNL thanks for your contribution!!!

DonKoko (approved):
Thanks for the PR @iDiegoNL 🤘

DonKoko (approved):
I had to update the PR because the TS tests were failing.

PR #1695, opened 2025-03-20
feature: export of bookings to csv

DonKoko:
- creating export button
- Creating route that will return csv
- sending search params from bookings index to csv export loader including handling selectAll
- created function that creates the csv for bookings


carlosvirreira:
Tested - let's ship it!

rockingrohit9639 (approved):

PR #1715, opened 2025-04-03
Update shelf.config.ts.md

gruffvaughan:
Correction of assumed typo (true instead of false)

### DIFF
--- a/docs/shelf.config.ts.md
+++ b/docs/shelf.config.ts.md
@@ -26,7 +26,7 @@ sendOnboardingEmail: true;
 
 ## enablePremiumFeatures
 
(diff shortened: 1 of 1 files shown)


DonKoko:
hey @gruffvaughan . 
Thanks for the PR. I realize the variable name is confusing but I think you misunderstood.
Setting it to true, means that subscription plan checks will run. You can check https://github.com/Shelf-nu/shelf.nu/blob/main/app/utils/subscription.server.ts to see how it works.

I do have to change this env variable at some point, but I  havent got to it yet, as it requires updating quite a few servers for us.

Hope that makes it more clear.

gruffvaughan:
Super clear now that I just tried to implement it myself! Apologies for the overeager correction.

PR #1751, opened 2025-04-24
Improvement: implement thumbnails for asset images

DonKoko:
## Changes implemented

- adding new field in asset model for the asset thumbnail image
- updating uploading function
- updating form data parsing function
- updating functions that update assets
- update every use of `AssetImage`
- changed refresh image and generate thumbnail endpoints to work with GET requests so they dont cause re-validation of all parent routes and reloading loops

To make everything backwards compatible we implemented a hybrid solution as running migrations would have been very time consuming and resource heavy

- **Enhanced Refresh Endpoint** - When images expire and need refreshing, it checks for thumbnails and generates them if missing.
- **On-Demand Generation** - The `AssetImage` component checks if a thumbnail is missing and generates it immediately if needed, independent of the expiration status.
- **Smart Loading States** - The component shows loading spinners during both refresh and thumbnail generation operations.

### This approach ensures:

- No heavy migration script needed
- Thumbnails are generated only when needed
- System self-heals over time
- Minimal impact on server resources
- Better user experience with progressive enhancement


DonKoko (commented):
2 issues:

1. The PDF files exported from bookings are still huge. I will look into what we can do about it.
2. The export booking overview UI is broken on safari:
image

PR #1803, opened 2025-05-27
Feat: i18n implementation

tomStory9:
I opened this draft Pr for the multi-language support as i said in our previous disscusion , already started their is not much to do except all the translation

This PR addresses **discussion #1780**, which proposed making the app multilingual to support non english user 

---

### Summary

Adds internationalization (i18n) support to **shelf.nu** using [i18n-remix](https://github.com/sergiodxa/remix-i18next/tree/v6.4.1). The language selection strategy follows this priority:

1. Default language defined in app settings
2. Language stored in the `i18nremix` cookie

---

### Main Changes

| File / Directory         | Description                                           |
|--------------------------|-------------------------------------------------------|
| `package.json`           | Added `i18n-remix and dep`                |
| `public/locales/*`       | Added `fr.json`, `en.json` (initial translation files)|
| `app/i18n      `          | Compiled translation                                 |
| `app/*`                  | use of translation instead of plain text             |

### Checklist
- [x] dependy is installed
- [x] Unit and integration tests still pass
- [ ] Translations are complete (in progress)
- [x] Language management by cookie
- [ ] User documentation updated
- [ ] UI/UX review planned

### DIFF
--- a/app/i18n/i18next.server.ts
+++ b/app/i18n/i18next.server.ts
@@ -44,10 +44,11 @@ export async function createI18nInstance(
 }
 
 export async function initTranslationLoader(
-  request: Request
+  request: Request,
+  namespace: string = "common"
 ): Promise<TFunction> {
   const lng = getLng(request);
-  return remixI18n.getFixedT(lng);
+  return remixI18n.getFixedT(lng, namespace);
 }
 
 export default remixI18n;
--- a/app/routes/_auth+/login.tsx
+++ b/app/routes/_auth+/login.tsx
@@ -41,7 +41,7 @@ import {
 import { validEmail } from "~/utils/misc";
 
 export async function loader({ context, request }: LoaderFunctionArgs) {
-  const t = await initTranslationLoader(request);
+  const t = await initTranslationLoader(request, "auth");
   const title = t("login.title");
   const subHeading = t("login.subHeading");
   const { disableSignup, disableSSO } = config;
@@ -121,7 +121,7 @@ export const meta: MetaFunction<typeof loader> = ({ data }) => [
 ];
 
 export default function IndexLoginForm() {
-  const { t } = useTranslation();
+  const { t } = useTranslation("auth");
   const { disableSignup, disableSSO } = useLoaderData<typeof loader>();
   const zo = useZorm("NewQuestionWizardScreen", LoginFormSchema);
   const [searchParams] = useSearchParams();
@@ -137,15 +137,13 @@ export default function IndexLoginForm() {
     <div className="w-full max-w-md">
       {acceptedInvite ? (
         <div className="mb-8 text-center text-success-600">
-          Successfully accepted workspace invite. Please login to see your new
-          workspace.
+          {t("login.acceptedInvite")}
         </div>
       ) : null}
 
       {passwordReset ? (
         <div className="mb-8 text-center text-success-600">
-          You have successfully reset your password. You can now use your new
-          password to login.
+          {t("login.passwordReset")}
         </div>
       ) : null}
       <Form ref={zo.ref} method="post" replace className="flex flex-col gap-5">
@@ -185,7 +183,7 @@ export default function IndexLoginForm() {
         </Button>
         <div className="flex flex-col items-center justify-center">
           <div className="text-center text-sm text-gray-500">
-            Don't remember your password?{" "}
+            {t("login.forgotPassword")}{" "}
             <Button
               variant="link"
               to={{
@@ -193,7 +191,7 @@ export default function IndexLoginForm() {
                 search: searchParams.toString(),
               }}
             >
-              Reset password
+              {t("login.resetPassword")}
             </Button>
           </div>
         </div>
@@ -201,7 +199,7 @@ export default function IndexLoginForm() {
       {!disableSSO && (
         <div className="mt-6 text-center">
           <Button variant="link" to="/sso-login">
-            Login with SSO
+            {t("login.continueWithSSO")}
           </Button>
         </div>
       )}
@@ -213,10 +211,8 @@ export default function IndexLoginForm() {
           </div>
           <div className="relative flex justify-center text-sm">
             <span className="bg-white px-2 text-gray-500">
-              Or use a{" "}
-              <strong title="One Time Password (OTP) is the most secure way to login. We will send you a code to your email.">
-                One Time Password
-              </strong>
+              {t("login.otp1")}
+              <strong title={t("login.otpTitle")}>{t("login.otp2")}</strong>
             </span>
           </div>
         </div>
@@ -225,7 +221,7 @@ export default function IndexLoginForm() {
         </div>
         {disableSignup ? null : (
           <div className="mt-6 text-center text-sm text-gray-500">
-            Don't have an account?{" "}
+            {t("login.noAccount")}{" "}
             <Button
               variant="link"
               data-test-id="signupButton"
@@ -234,7 +230,7 @@ export default function IndexLoginForm() {
                 search: searchParams.toString(),
               }}
             >
-              Sign up
+              {t("login.signup")}
             </Button>
           </div>
         )}
--- a/i18n.json
+++ b/i18n.json
@@ -6,7 +6,7 @@
   },
   "buckets": {
     "typescript": {
-      "include": ["app/locales/[locale]/common.ts"]
+      "include": ["app/locales/[locale]/*.ts"]
     }
   },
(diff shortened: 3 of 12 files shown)


app/entry.server.tsx
DonKoko, line 1: I am not sure why you removed the logic with `handleBotRequest` and `handleBrowserRequest`. Considering how our app works and that everything is behind authentication, we should only consider the localization for user requests. Why did you decide to not implement the i18n logic inside `handleBrowserRequest` instead of refactoring the whole approach.

app/i18n/cookie.ts
```diff
@@ -0,0 +1,9 @@
+import { parse } from "cookie";
+import i18n from "./i18n"; // your i18n configuration file
+export function getLng(request: { headers: { get: (arg0: string) => any } }) {
+  const cookies = request.headers.get("cookie");
```
DonKoko: Please check `cookies.server.ts` and use our already existing functions for reading cookies. Also I don't see anywhere that you are creating the cookie. How do you plan to do that?

app/routes/_auth+/login.tsx
```diff
@@ -40,9 +41,11 @@ import {
 } from "~/utils/http.server";
 import { validEmail } from "~/utils/misc";
 
-export function loader({ context }: LoaderFunctionArgs) {
+export async function loader({ context, request }: LoaderFunctionArgs) {
+  let lng = getLng(request);
+  let t = await i18next.getFixedT(lng);
   const title = "Log in";
-  const subHeading = "Welcome back! Enter your details below to log in.";
+  const subHeading = t("greeting");
```
DonKoko: I know you are doing this just for testing, but please use the original texts that were there for your testing as once we start implementing all the translations it will be very easy to miss and forget this.

app/i18n/cookie.ts
tomStory9: The cookie is created by i18next plugin https://github.com/i18next/i18next-browser-languageDetector but it was not able to read it with getLocale method so i just write my own cookie reader

app/routes/_auth+/login.tsx
tomStory9: I'am sorry i will use origin text for this

app/entry.server.tsx
tomStory9: Sorry , for my poc i tried to simplify the entry.server as mush as i could so i refactored to just need to had my code in one function. I forgot to change to base. I will reverse this refactor and make translation only for user not bot

app/entry.server.tsx
```diff
@@ -149,20 +156,37 @@ function handleBotRequest(
...
 }
 
-function handleBrowserRequest(
+async function handleBrowserRequest(
   request: Request,
   responseStatusCode: number,
   responseHeaders: Headers,
   remixContext: EntryContext
 ) {
+  let instance = createInstance();
```
DonKoko: This whole code that creates the instance. Shall we add it to a function placed inside i18next.server.ts and just invoke it here. Will make things a bit more organized.

app/entry.server.tsx
```diff
@@ -149,20 +156,37 @@ function handleBotRequest(
...
+  let ns = i18next.getRouteNamespaces(remixContext);
+  let lng = getLng(request);
+  await instance
+    .use(initReactI18next) // Tell our instance to use react-i18next
+    .use(LanguageDetector)
+    .init({
+      ...i18n, // spread the configuration
+      ns, // The namespaces the routes about to render wants to use
+      lng, // The language to use
+      resources: { en: { common: en }, fr: { common: fr } },
```
DonKoko: If we store the `supportedLngs` in shelf.config we can then have a function that generates this object as we need to use it in a couple of different places. Adding a new language should be as simple as adding it to supportedLngs in shelf.config and creating the empty translation file.

app/i18n/i18n.ts
```diff
@@ -0,0 +1,13 @@
+export default {
+  detection: {
+    order: ["htmlTag"],
+    lookupCookie: "i18next",
+  },
+  // This is the list of languages your application supports
+  supportedLngs: ["en", "fr"],
```
DonKoko: Here again we are repeating this languages. This will be tricky to maintain. We need a central way to manage what languages we have. Maybe this should be stored in the shelf.config file.

app/i18n/i18next.server.ts
```diff
@@ -0,0 +1,21 @@
...
+let i18next = new RemixI18Next({
+  detection: {
+    supportedLanguages: i18n.supportedLngs,
+    fallbackLanguage: i18n.fallbackLng,
+  },
+  // This is the configuration for i18next used
+  // when translating messages server-side only
+  i18next: {
+    ...i18n,
+    resources: { en: { common: en }, fr: { common: fr } }, // You can add more languages here
```
DonKoko: Same as above.

app/routes/_auth+/login.tsx
```diff
@@ -40,9 +40,11 @@ import {
 } from "~/utils/http.server";
 import { validEmail } from "~/utils/misc";
 
-export function loader({ context }: LoaderFunctionArgs) {
+export async function loader({ context, request }: LoaderFunctionArgs) {
+  let lng = getLng(request);
```
DonKoko: So every time we need translations in a loader we need to call those 2:
```ts
let lng = getLng(request);
let t = await i18next.getFixedT(lng);
```

I think we should simplify that. Like in the components we have a hook, for loaders/actions we need to have a function that receives request and returns `t()` without having to do this manually every time. Also another important point, does this consume the request? Its important to know so we can then know if we should clone it before using it.

i18n.json
DonKoko, line 1: I can imagine this is the config for lingo to translate. We would need to add a file to the docs folder explaining how this is used for our open source users.

DonKoko (changes requested):
What I am still missing is a language selector that loads data from cookie and then  creates/updated cookie when changed. I think that's a very key part and can be a bit time consuming to build, so I would recommend to create it sooner rather than later.

lexcorp:
Hi... It´s possible include latin spanish language?

app/entry.server.tsx
tomStory9: we cant add it inside this file because vite will throw the following error 

[vite] Pre-transform error: Server-only module referenced by client

    './i18n/i18next.server' imported by 'app/entry.client.tsx'

  See https://remix.run/docs/en/main/guides/vite#splitting-up-client-and-server-code
  
  But i thinks we can create a utils and add code to create instance and return the translation function for loader

lexcorp:
Is there any way I can help?

app/config/shelf.config.ts
```diff
@@ -4,10 +4,14 @@ import {
   ENABLE_PREMIUM_FEATURES,
   FREE_TRIAL_DAYS,
   SEND_ONBOARDING_EMAIL,
+  SUPPORTED_LANGUAGES,
+  FALLBACK_LANGUAGE,
 } from "~/utils/env";
 import { Config } from "./types";
 
 export const config: Config = {
+  SUPPORTED_LANGUAGES: SUPPORTED_LANGUAGES || ["en", "fr"],
```
DonKoko: config properties should be camel cased

app/config/types.ts
```diff
@@ -45,4 +45,7 @@ export interface Config {
    * Disable SSO
    */
   disableSSO: boolean;
+
+  SUPPORTED_LANGUAGES: Array<string>;
```
DonKoko: same as above

vite.config.ts
```diff
@@ -30,10 +30,13 @@ export default defineConfig({
...
   },
   optimizeDeps: {
+    esbuildOptions: {
+      target: "esnext",
+    },
     include: ["./app/routes/**/*"],
   },
   build: {
-    target: "ES2022",
+    target: "esnext",
```
DonKoko: I am worried this will create conflicts with other packages and features.

tsconfig.json
```diff
@@ -1,5 +1,11 @@
 {
-  "include": ["env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": [
+    "env.d.ts",
+    "**/*.ts",
+    "**/*.tsx",
+    "public/locales/en/common.js",
```
DonKoko: This should use wildcard, instead of having to add each locale.

DonKoko (changes requested):

tsconfig.json
tomStory9: I forgot to delete these lines after my tests. My apologies.

PR #1810, opened 2025-05-29
improvement: bookings csv export labels:

DonKoko:
- update labels of dates to better reflect their meaning
- add status column


DonKoko:
@carlosvirreira FYI

PR #1826, opened 2025-06-06
Fix the docker container build and deployment

Lejo1:
Github is blocking the action because of a outdated workflow.
I changed all the versions to the current major version to prevent future problems of the same kind.

The working workflow: https://github.com/Lejo1/shelf.nu/actions/runs/15499705143
The current built image: https://github.com/Lejo1/shelf.nu/pkgs/container/shelf.nu/432904083?tag=dev-1f88cbe


DonKoko:
Thank you very much @Lejo1 . 
As you might have noticed the docker build is completely community supported so I am sure some of our users will be very grateful to you.

PR #1952, opened 2025-07-25
feat: partial check in & scan to check in:

DonKoko:


DonKoko:
- [ ] I managed to check out an asset in 2 bookings
- have a ongoing booking
- create a second ongoing booking.
- Use scan to add to add an asset to the second ongoing booking

- [ ] Assets that are checked out are showing labels when they shouldn't 
- [ ] Availability checks should still be implemented for kits the same way we did for assets with the new 2 rule system. Now if a kit has been checked in with partial, in a conflicting booking it will still show as already booked and it shouldn't as its avaialble
- [ ] Should update the booking assets preview sheet to show better information about asset including some AvailabilityLabels

PR #1975, opened 2025-08-13
Fix: Prisma config to load environment variables

DonKoko:
Add dotenv/config import to prisma.config.ts to ensure environment variables are loaded when using the new Prisma configuration format. This resolves the 'Environment variable not found: DIRECT_URL' error that occurred after migrating from package.json to prisma.config.ts.

### DIFF
--- a/prisma.config.ts
+++ b/prisma.config.ts
@@ -1,3 +1,4 @@
+import "dotenv/config";
 import { defineConfig } from "prisma/config";
 
 export default defineConfig({


MadeByDouglas:
So i am experiencing this error on a fresh build. I think I have everything setup, .env is in the root shelf.nu folder with all my supabase stuff, but when i try to run `npm run setup` it says it cannot find the the environment variable. it seems its skipping loading it?

```
Prisma config detected, skipping environment variable loading.
Prisma schema loaded from app/database/schema.prisma
Error: Prisma schema validation - (get-config wasm)
Error code: P1012
error: Environment variable not found: DIRECT_URL.
  -->  app/database/schema.prisma:4
   | 
 3 |   url       = env("DATABASE_URL")
 4 |   directUrl = env("DIRECT_URL")
   | 
```

MadeByDouglas:
OK I tried node 20 but same error. My .env file is in the root shelf.nu folder, is that right? 
I'm trying to think if i forgot anything, I cloned the repo, added the .env file with supabase info, then ran `npm run setup`, and always get this error

MadeByDouglas:
OK, I tried fresh npm install after removing package lock, then specifically installing `npm install dotenv` as that other issue mentioned. No dice.

Then I had claude look at it, turns out, when i created my .env file it was saved as a .env.txt file, even though I couldn't see that anywhere in my text editor or finder. Once it stripped the .txt its working. 

Anyway, appreciate your help, sorry for the hassle.

carlosvirreira:
Thank you for sharing the solution @MadeByDouglas - indeed, as Nikolay says, even small contributions are appreciated. Glad you can take advantage of the work done so far, let's make Shelf even better!

PR #2040, opened 2025-09-16
feat: add global command palette

carlosvirreira:
## Summary
- implement a command palette with keyboard access, debounced search, and grouped results for assets, navigation, and quick actions
- add reusable command UI primitives plus buttons and provider wiring in the main layout and header
- expose a dedicated API endpoint for asset lookups and add cmdk/fuse dependencies to support fuzzy matching

## Testing
- npm run lint
- npm run typecheck

------
https://chatgpt.com/codex/tasks/task_b_68c98bc49bf08326adbe6cf955cd3989


DonKoko (changes requested):
@carlosvirreira @codex I made a quick review by testing this. I haven't looked at the code in details yet but already found some major stuff that need to be addressed and polished.

- [ ] Search bar in command palette looks broken in terms of spacing:
Image

- [ ] The button to open the command palette doesn't look anything like shelf style buttons. I am also not sure if that is the best place to position it.
Image

- [ ] The searching is basically broken. When a search is performed there seems to be an endless loop where it constantly queries the endpoint. I haven't debug it in details but i suspec its because of the useEffect hell that is inside the command-palette component. This should not be needed at all when using remix, as everything gets auto re validated. Also the code is missing any inline documentation so when looking at the component its a bit harsh to figure out what its doing.

carlosvirreira:
Thanks for the feedback @DonKoko i am testing myself again and will drop a new PR with fixes

PR #2097, opened 2025-10-08
Fix calendar booking dialog missing custodian default

DonKoko:
## Summary
- include the self-service flag and current user id in the calendar loader payload
- ensure the create booking dialog on the calendar has access to the custodian defaults like other routes

## Testing
- npm run typecheck

------
https://chatgpt.com/codex/tasks/task_b_68e531c8ecac8320b42bc599d707099f


carlosvirreira:
Thanks for the fix!

PR #2113, opened 2025-10-13
feat: custom field deletion

DonKoko:
## Summary
- reset the custom field deletion dialog when it closes without relying on side effects
- perform custom field removal inside a transaction, using a single SQL update to drop index columns
- add coverage for the asset index cleanup helper and ensure deletion errors use ShelfError responses

## Testing
- npm run test -- app/modules/asset-index-settings/service.server.test.ts
- npm run lint

------
https://chatgpt.com/codex/tasks/task_b_68e5075fa40c83208f060998e08c09cb


DonKoko:
> In my opinion you should be able to do the same from both areas.

Deleting custom fields is tricky on DB level and requires confirmation by typing the field name. That is why I have excluded it from bulk actions

PR #2122, opened 2025-10-16
feat: show placeholders for missing asset table values

carlosvirreira:
## Summary
- add a shared EmptyTableValue helper for rendering muted placeholders with screen reader labels
- use the placeholder in asset index simple and advanced columns when optional data is missing
- cover both table modes with a regression test for empty tags, custody, location, and kit cells

## Testing
- npm run test -- assets-list

------
https://chatgpt.com/codex/tasks/task_b_68f0bdba697c832696c729a7297cb214


app/components/assets/assets-index/assets-list.test.tsx
DonKoko, line 1: This test file mocks a lot of modules that aren’t actually used. We can simplify it by keeping only the mocks required for useLoaderData and the components/hooks directly rendered by ListAssetContent or AdvancedIndexColumn. Removing the rest will make the test easier to read and maintain, without changing its behavior.

PR #2136, opened 2025-10-22
feat: refresh signup welcome and plan selection flow

carlosvirreira:
## Summary
- update the /welcome choose-purpose view with new copy, guardrail hint, plan badges, and analytics-aware CTA behaviour
- redesign the /select-plan route with static team plan cards, optional add-ons section, refreshed trial messaging, and navigation back link
- add focused tests covering the revised welcome selection states and plan selection layout

## Testing
- npx vitest run app/components/welcome/choose-purpose.test.tsx app/routes/_welcome+/select-plan.test.tsx

------
https://chatgpt.com/codex/tasks/task_b_68f7bb3dc5ac832683f243af3aab07b5

### DIFF
--- a/app/components/assets/assets-index/advanced-filters/value-field.tsx
+++ b/app/components/assets/assets-index/advanced-filters/value-field.tsx
@@ -865,11 +865,12 @@ function CustodyEnumField({
                   selectedIds.length <= 0 && "text-gray-500"
                 )}
               >
-                {value === "without-custody"
-                  ? "Without custody"
-                  : selectedIds.length > 0
+                {selectedIds.length > 0
                   ? selectedIds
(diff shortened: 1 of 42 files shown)


carlosvirreira:
hey @DonKoko works nice - thank you for your latest polish on top of all. 

Please check that we write 7-day trial as this is our current default. 

For the rest I find this to be a nice enhancement to what we have. 

Only thing I noticed that might be - unrelated is that when i clicked on the trial I got an error of not being able to create more workspaces. So - ideally this is something we check if it wont translate to prod.

Screenshot 2025-10-23 at 16 58 14

Screenshot 2025-10-23 at 16 58 42

PR #2142, opened 2025-10-22
fix: make SortBy control responsive on mobile

carlosvirreira:
## Summary
- render the SortBy selects inline on mobile viewports to avoid the Radix popover trap
- keep both selects controlled so search params stay in sync across views
- add tests covering the mobile inline experience and desktop popover trigger
- explicitly note that Playwright coverage remains out of scope per CTO guidance

## Testing
- npm run test -- sort-by

------
https://chatgpt.com/codex/tasks/task_b_68f8e2ed58208326b78fdd7f1dea307c


DonKoko (changes requested):
This PR doesn't exactly fix the existing bug, but rather completely replaces the sort by functionality on mobile.
A few things:
1. This new design basically takes half the screen which I don't think is good.

Before: 
IMG_FBEB5212B268-1

After:
Nikolay's inventory  shelf nu

2. The way this is implemented is very strange. You are changing both `select` elements to "controlled", but not really as there is no state, they are controlled by the search params. If that is the approach they should just be kept as uncontrolled. Moreover this doesn't seem to be related to the resolution of the issue as the issue was defined as something happening because of the Popover focus on mobile, and that popover is removed so this seems necessary

3. The code is quite messy and hard to understand. If it is really required to use a different component for mobile, it should be abstracted into its own file and everything should be kept clean and easy to understand and maintain

4. The tests have a lot of issues with them. They are not testing the actual code but rather testing the mocks. The actual important validations and behaviors of the component are not tested. IMO those tests are basically useless.

PR #2158, opened 2025-10-29
feat: add fallback invite link to email

carlosvirreira:
## Summary
- add fallback invite link text to the workspace invitation email
- cover the invitation email renderer with a test that asserts the fallback URL

## Testing
- npm run test -- invite-template

------
https://chatgpt.com/codex/tasks/task_b_6902007dbc18832687a11ece54111bc3

### DIFF
--- a/app/emails/invite-template.test.ts
+++ b/app/emails/invite-template.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+
+import { invitationTemplateString } from "~/emails/invite-template";
+import type { InviteWithInviterAndOrg } from "~/modules/invite/types";
+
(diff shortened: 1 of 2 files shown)


app/emails/invite-template.test.ts
DonKoko, line 1: This test is not needed. It doesn't test if the function works but rather the developer behavior.

DonKoko (changes requested):
I don't get the purpose of adding this text email link. Is there some issue with the clicking of the button? What is it solving? Also keep in mind that spam bots don't like links in emails so I am quite certain adding this won't be helpful for the issue we already have with our emails going in spam.

PR #2190, opened 2025-11-21
feat: change default asset index mode to ADVANCED

carlosvirreira:
Previously new users would see the SIMPLE asset index view by default. This changes the default to ADVANCED mode which provides more features and a richer interface for managing assets.


DonKoko:
Something went wrong with that PR. There are 46 files changes. Seems like claude did a lot of formatting changes that it shouldn't have done.

PR #2202, opened 2025-11-25
feat: change default asset index mode to ADVANCED

carlosvirreira:
Previously new users would see the SIMPLE asset index view by default. This changes the default to ADVANCED mode which provides more features and a richer interface for managing assets.


carlosvirreira:
@DonKoko 

Fresh account, signed up personal workspace > land into simple mode, no 403. 
invited admin user as owner, defaulted to advanced (no 403)
invited a base user to workspace, defaulted to simple asset index on the invited workspace (as expected)
invited self service user to workspace, defaulted to simple asset index (expected)
//
removed access from user, then from NRM re-invited > this time it defaulted back to simple index

removed same user, then invited him as admin but not via the nrm table > defaulted back to simple index
//

CSV export / bulk endpoints: Execute an export while logged in as ADMIN and OWNER to ensure the loader still fetches settings with the correct mode. > this works

carlosvirreira:
All tests pass. Let's sail. @DonKoko

Issue #1988, opened 2025-08-21
Feat: Create entities from inline context of dynamic select and dropdown


DonKoko:
In our `DynamicDropdown` and `DynamicSelect` we have the option of "Create new ...". 
<img width="629" height="628" alt="Image" src="https://github.com/user-attachments/assets/8b0be6c8-ee0a-4f01-87b2-d1f10b5e61a2" />

Currently this is very scuffed. Its just a link to a new tab so the UX is quite terrible. The idea is that we make this work inline, so the user doesn't have to leave the context.
So currently its used like this:
```ts
extraContent={
              <Button
                to="/categories/new"
                variant="link"
                icon="plus"
                className="w-full justify-start pt-4"
                target="_blank"
              >
                Create new category
              </Button>
            }
```
And I have checked and we only use it for "category" and "location". So the idea is to create a component that we can pass in the extra content that triggers a dialog to create a location or category. We would need to add api endpoints for the creation.
We also have to make sure that when this is created, we somehow reload the data of the DynamicDropdown/Select to the new added entry is included in the list.

PR #2219, opened 2025-11-27
Feat: create inline entities

DonKoko:
Closes https://github.com/Shelf-nu/shelf.nu/issues/1988

### DIFF
--- a/app/modules/asset-index-settings/service.server.ts
+++ b/app/modules/asset-index-settings/service.server.ts
@@ -1,4 +1,9 @@
-import type { AssetIndexMode, CustomField, Prisma } from "@prisma/client";
+import {
+  AssetIndexMode,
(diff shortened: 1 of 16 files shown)


DonKoko:
Last feedback from previous issue:
- [ ]  When a new entry is created, we have to make sure we somehow already query and set this as the selected entry in the select or dropdown
- [ ] The form buttons dont get a disabled state while submitting. This could create double submissions and data errors
- [ ] In the case of "category" we have a bunch more places where we could add this that it makes sense: BulkCategoryUpdateDialog, CategoriesInput(this is for custom fields so i think it makes sense to add it there)
- [ ] For locations it works a bit different. We actually have a LocationSelect component that is used in 3 different ways that allow updating locations, so I think there we have to include the "create new location" as well.

PR #2230, opened 2025-12-03
feat: colored tags

pj-gridx:
### Tag color support
- schema/migration adds color with default #f2f4f7  
- tag services accept/persist color  
- CSV/import tags get a random color  

### forms/UI
- tag create/edit use the color picker  
- tag badges support colored backgrounds/dots  
- asset/tag listings, asset overview, and booking stats render colored tags with safe optional color handling  
- color input widened for hex codes  

### booking pipeline
- tag color included in booking queries (constants, pdf helpers, routes)  
- booking service tests updated for color-aware expectations and future-dated fixtures  

### assets/types
- tag color added to asset loader/types  
- list item tags column and booking stats tolerate missing colors  

### housekeeping
- lockfile updated  
- typegen/typecheck cleaned up around optional colors  

## Summary by CodeRabbit

* **New Features**
  * Tags now display with color indicators and styled visual badges throughout the application, including asset views, booking lists, and tag management
  * Users can customize tag colors through a new color input field in tag creation and editing interfaces
  * Tags are automatically assigned a default color to ensure consistent visual representation

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/app/components/shared/tag.tsx
+++ b/app/components/shared/tag.tsx
@@ -21,7 +21,7 @@ export const Tag = forwardRef<HTMLSpanElement, TagProps>(function Tag(
     <span
       ref={ref}
       className={tw(
-        "inline-flex items-center rounded-2xl bg-gray-100 py-[2px] pl-[6px] text-[12px] font-medium text-gray-700",
+        "inline-flex items-center rounded-2xl bg-gray-100 py-0.5 pl-1.5 text-[12px] font-medium text-gray-700",
         withDot ? " gap-1 pr-2" : "px-2",
         className
       )}


carlosvirreira:
@coderabbit full review

DonKoko (changes requested):
Thanks for the PR.
Overall it looks good but you have 11k lines changed in `package-lock.json`. Please resolve this as those changes shouldn't be present in the branch. 
Once resolved I will run the branch locally and test it further.

app/components/shared/tag.tsx
DonKoko, line 1: This component is used in a lot of places, not only for the tags. You need to double check other uses and make sure its fully backwards compatible and didn't break anything.

app/components/assets/assets-index/list-item-tags-column.tsx
```diff
@@ -6,7 +6,11 @@ import { Tag as TagBadge } from "~/components/shared/tag";
 export const ListItemTagsColumn = ({
   tags,
 }: {
-  tags: Pick<Tag, "id" | "name">[] | undefined;
+  tags:
+    | (Pick<Tag, "id" | "name"> & {
+        color?: string | null;
```
DonKoko: Why not include `color` in the Pick?

app/components/booking/booking-statistics.tsx
```diff
@@ -24,7 +25,7 @@ export function BookingStatistics({
   assetsCount: number;
   totalValue: string;
   allCategories: { id: string; name: string; color: string }[];
-  tags: Pick<Tag, "id" | "name">[];
+  tags: (Pick<PrismaTag, "id" | "name"> & { color?: string | null })[];
```
DonKoko: Why not include color in the Pick?

app/routes/_layout+/settings.team.users.$userId.bookings.tsx
```diff
@@ -78,7 +78,7 @@ export async function loader({ context, request, params }: LoaderFunctionArgs) {
         }),
         tags: filterTags,
         extraInclude: {
-          tags: { select: { id: true, name: true } },
+          tags: { select: { id: true, name: true, color: true } },
```
DonKoko: We are repeating this `select` in a lot of places. I think its a good idea to create a constant for it so we can easily maintain it in the future, having to change only 1 place. There are plenty examples of this in our codebase where you can see how we do it.

app/routes/_layout+/tags.$tagId_.edit.tsx
```diff
@@ -27,6 +28,7 @@ import { zodFieldIsRequired } from "~/utils/zod";
 export const UpdateTagFormSchema = z.object({
   name: z.string().min(3, "Name is required"),
   description: z.string(),
+  color: z.string().regex(/^#/).min(7),
```
DonKoko: This will accept invalid values. Can you please update the regex:
```
color: z.string().regex(/^#[0-9a-fA-F]{6}$/),
```
I think this should be better.

app/routes/_layout+/tags.new.tsx
```diff
@@ -26,6 +28,7 @@ import { zodFieldIsRequired } from "~/utils/zod";
 export const NewTagFormSchema = z.object({
   name: z.string().min(3, "Name is required"),
   description: z.string(),
+  color: z.string().regex(/^#/).min(7),
```
DonKoko: Same as with update tag schema. Improve the regex

DonKoko (changes requested):
Overall very good PR. There were some conflicts with `main` I had to resolve so make sure to update your local branch before working on it further.
I have added some feedback, nothing major but important to make this complete.

- [x] On the advanced asset index, the styling of the tags is not applied properly. In this case "legendary" has a custom color but its showing as default
Image

- [ ] On the tags index as well as other views such as the asset page, simple asset index, the default color of tags is broken. It is not accessibility compliant and is very unreadable. Lets make it the default look like it does on the advanced index mode.
Image

- [x] On the tags index as well as some other places the tags are showing with a dot. Lets remove that to keep them clean.

app/routes/_layout+/tags.$tagId_.edit.tsx
pj-gridx: Good point, I changed the regex for the category badges as well.

app/components/assets/assets-index/list-item-tags-column.tsx
pj-gridx: It's included, the idea here is to be backwards compatible if there is no color set.

pj-gridx: Found the issue, there was a missing "color" in the query.server.ts

pj-gridx:
@DonKoko I adjusted the code, thanks for the detailed review.

**Tag visuals refined**
removed the decorative dots on tag badges, matched the tag layout to the badge styling for consistency, and tweaked the default tag color to improve contrast/visibility.

**Color data fixes**:
now pulling the actual tag color from the DB instead of a fallback, using a const select shape for id/name/color to keep typings accurate, and tightened the color regex to allow only 6‑digit hex codes.

DonKoko:
Thanks for the great PR @pj-gridx. I just had to make some minor adjustments but overall it was great. I appreciate your input.

pj-gridx:
@carlosvirreira @DonKoko thank you both for the great support!

PR #2282, opened 2026-01-26
(PRD) > Add server-side workspace persistence for cross-device consistency

carlosvirreira:
## Summary

This PR adds a `lastSelectedOrganizationId` field to the User model to persist workspace selection across devices, solving the issue where users land in the wrong workspace when logging in from a new device or after cookies are cleared.

## Problem

Users with multiple workspaces (personal + organization) experience confusion when:
- Logging in from a new device without cookies
- Clearing browser cookies (falls back to arbitrary first org)
- Using incognito/private browsing mode

This causes users to think their data is lost when it's actually in a different workspace they can't see.

## Solution

Implement a simple, elegant fallback hierarchy for workspace selection:

1. **Cookie value** (if valid and user still has access)
2. **Database `lastSelectedOrganizationId`** (if valid and user still has access) ← NEW
3. **Personal workspace** (safe default)
4. **First available organization** (edge case fallback)

## Key Changes

- **Schema**: Add nullable `lastSelectedOrganizationId` field to User model
- **Write Path**: Update `change-current-organization` action to persist workspace selection to database alongside cookie
- **Read Path**: Update organization context logic to check database field as fallback when cookie is missing/invalid

## Implementation Details

- **Minimal footprint**: ~20 lines of code across 3 files
- **Backward compatible**: Existing cookie system remains unchanged; database field acts as persistent backup
- **Safe defaults**: Falls back to personal workspace if user loses access to previously selected org
- **No new tables**: Leverages existing User model, no schema complexity

## Testing

- User switches org → DB field updated
- User logs in new device without cookie → lands in last selected org
- User removed from org → falls back to personal org gracefully
- New user with no history → lands in personal org
- Cookie still works for same-device preference

## Files Modified

- `app/database/schema.prisma` - Add field to User model
- `app/routes/api+/user.change-current-organization.ts` - Persist selection on switch
- `app/modules/organization/context.server.ts` - Update fallback logic

### DIFF
--- a/app/modules/organization/context.server.ts
+++ b/app/modules/organization/context.server.ts
@@ -86,8 +86,8 @@ async function getSelectedOrganizationUncached({
     cookieRefreshNeeded = true;
 
     // Piggyback on the already-fetched userOrganizations query
-    const lastSelectedOrganizationId =
-      userOrganizations[0]?.user?.lastSelectedOrganizationId ?? null;
+    const user = userOrganizations[0]?.user;
+    const lastSelectedOrganizationId = user?.lastSelectedOrganizationId ?? null;
 
     if (
       lastSelectedOrganizationId &&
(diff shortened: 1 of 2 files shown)


DonKoko (commented):
This PR has major issues. Both security and actual gaps in the implementation that could cause performance and UX degradation. I will resolve everything and update the PR.

PR #2283, opened 2026-01-26
feat: add positive state filter options for custody, location, and kit

carlosvirreira:
## Summary

- Add "In custody", "In a location", and "In a kit" filter options to remove the need for
  double negatives (e.g., "Custody is not: Without custody")
- Implement mutual exclusion logic to prevent selecting conflicting options (positive and
  negative) simultaneously
- Add 20 unit tests covering special filter value handling and edge cases
- Unify labels to consistent "In a X" pattern across all filters

## Changes

**UI Components:**
- `DynamicSelect` and `DynamicDropdown`: Added `withValueItem` prop for positive state options
- `value-field.tsx`: Added `filterConflictingSelections` helper and integrated new options

**Query Building:**
- `query.server.ts`: Handle `in-custody`, `in-location`, `in-kit` special values with proper
  SQL generation for `is`, `isNot`, and `containsAny` operators

**Display:**
- `format-filter-summary.ts`: Map special IDs to user-friendly labels

## Test plan

- [x] Unit tests pass for all special filter value combinations
- [ ] Verify "In custody" filter returns assets with custodians
- [ ] Verify "In a location" filter returns assets assigned to locations
- [ ] Verify "In a kit" filter returns assets in kits
- [ ] Verify selecting conflicting options auto-deselects the previous one
- [ ] Verify saved filter presets with new values load correctly
- [ ] Verify filter summary displays correct labels

### DIFF
--- a/app/components/settings/transfer-ownership-card.tsx
+++ b/app/components/settings/transfer-ownership-card.tsx
@@ -35,6 +35,10 @@ import When from "../when/when";
 
 type TransferOwnershipCardProps = {
   className?: string;
+  /** Form action URL. Defaults to "/settings/general" */
+  action?: string;
+  /** Organization name for confirmation input. If not provided, uses currentOrganization from context */
+  organizationName?: string;
 };
 
 export const TransferOwnershipSchema = z.object({
@@ -53,6 +57,8 @@ export const TransferOwnershipSchema = z.object({
 
 export default function TransferOwnershipCard({
   className,
+  action = "/settings/general",
+  organizationName,
 }: TransferOwnershipCardProps) {
   const { admins } = useLoaderData<typeof loader>();
   const { isOwner } = useUserRoleHelper();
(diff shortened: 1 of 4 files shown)


app/components/assets/assets-index/advanced-filters/value-field.tsx
```diff
@@ -40,6 +40,44 @@ import { extractQrIdFromValue } from "./helpers";
...
+ * When user selects a positive option (e.g., "in-custody"), automatically
+ * removes the negative option (e.g., "without-custody") and vice versa.
+ *
+ * @param newSelection - The new selection array from the dropdown
+ * @param previousSelection - The previous selection array
+ * @param positiveId - The ID of the positive option (e.g., "in-custody")
+ * @param negativeId - The ID of the negative option (e.g., "without-custody")
+ * @returns Filtered selection with conflicting options removed
+ */
+function filterConflictingSelections(
```
DonKoko, line 54: This works good mostly, however it doesn't unselect the entry completely. It doesnt show it in the select trigger but it still shows it checked in the list:

Image

This applies to all 3 models: kits, locations and custody.

PR #2295, opened 2026-01-29
feat: add asset removal from location page directly

DonKoko:
Add row-level and bulk asset removal actions to the location assets page, eliminating the need to navigate to manage-assets route just to remove assets.

- Add RemoveAssetFromLocation component with confirmation dialog
- Add AssetRowActionsDropdown for per-row remove action
- Add BulkRemoveAssetsFromLocationDialog for bulk removal
- Add ListBulkActionsDropdown with bulk select support
- Add action handler in location assets route for both single and bulk asset removal intents
- Wire up bulk selection checkboxes and actions column
- Fix double bottom border on custodian Th header
- Add empty Th for actions column alignment


carlosvirreira:
nice one! @DonKoko

PR #2319, opened 2026-02-09
fix(assets): use form data instead of search params for category in action

carlosvirreira:
## Summary
Refactored the order of operations in the asset creation action to clone the request and extract form data before retrieving custom fields. This ensures the category value from the form data is available when fetching active custom fields.

## Key Changes
- Moved request cloning and `formData()` extraction earlier in the function (before custom fields retrieval)
- Updated `getActiveCustomFields()` call to use category from form data as a fallback when not present in search params
- Maintains the same functionality while improving data flow and ensuring form data is available when needed

## Implementation Details
The change ensures that when custom fields are retrieved, the category parameter can be sourced from either:
1. URL search params (primary source)
2. Form data (fallback source)

This prevents potential issues where the category might only be available in the form submission and not in the URL parameters.

https://claude.ai/code/session_01HZ4WP53vzmUBFAoQfpKshE

### DIFF
--- a/app/routes/_layout+/assets.new.tsx
+++ b/app/routes/_layout+/assets.new.tsx
@@ -131,13 +131,9 @@ export async function action({ context, request }: LoaderFunctionArgs) {
 
     const formData = await clonedRequest.formData();
 
-    const searchParams = getCurrentSearchParams(request);
-
     const customFields = await getActiveCustomFields({
       organizationId,
-      category:
-        searchParams.get("category") ??
-        (formData.get("category") as string | null),
+      category: formData.get("category") as string | null,
     });
 
     const FormSchema = mergedSchema({


carlosvirreira:
hi @DonKoko This addresses a bug reported on Crisp — custom field values silently dropped when creating a new asset. Editing the same asset and re-entering the values works fine.

Root cause: the create action calls `getActiveCustomFields` using only `searchParams.get("category")`, which can be `null` when the form submits (React Router doesn't preserve search params with `action="."`). When category is null, only uncategorized fields are returned → Zod schema is built without the category-specific `cf-*` keys → `safeParse` silently strips them → asset is created with no custom field values.

The edit action already handles this correctly with a fallback:
```
searchParams.get("category") ?? String(formData.get("category"))
```

This PR applies the same fallback pattern to the create action and reorders the `formData` read to happen before `getActiveCustomFields` so the fallback is actually available.

PR #2368, opened 2026-02-27
fix: scope toast notifications to originating browser tab

### LINKED ISSUE #1000: [Bug]: Toast Notifications
### Contact Details

_No response_

### What happened?

When Shelf is open on multiple devices connected to the same account, notifications (asset modification, booking registration, etc.) appear on all devices simultaneously.

Steps to reproduce:

1. Open Shelf on two devices/web browser
2. Go to an asset
3. Click on Edit
4. Click on Save
5. See the "Asset updated" popup on each browser

The message should only be displayed on the browser that made the modification.

### Version

Self-hosted

Firefox, Chrome, Safari, Microsoft Edge

### Relevant log output

_No response_

iuryeng:
## Summary

Fixes #1000 — Toast notifications now only appear on the browser tab that
triggered the action, instead of popping up on every open tab.

### Problem

When a user has multiple tabs open, performing an action (e.g. editing an
asset) shows the success toast on **all tabs** because the SSE notification
handler only filters by `userId`, not by tab.

### Solution

Each browser tab gets a unique identifier (`tabId`) stored in
`sessionStorage` (which is isolated per tab). The fix flows through 4 layers:

1. **Client** — `useTabId()` hook generates a `tabId` per tab and patches
   `window.fetch` to send it as an `X-Tab-Id` header on every same-origin
   request
2. **Server middleware** — Hono middleware reads `X-Tab-Id` and stores it
   in `AsyncLocalStorage` for the duration of the request
3. **Notification emitter** — `sendNotification()` automatically tags each
   notification with the current request's `tabId` (zero call-site changes
   needed across 100+ usages)
4. **SSE handler** — Compares the notification's `tabId` against the SSE
   connection's `tabId` query param; only delivers if they match

Notifications from background jobs (pg-boss workers, scheduled tasks) run
outside a request context, so `tabId` is `undefined` and the toast is
broadcast to all tabs — preserving existing behavior.

### Files changed (7)

| File | Change |
|------|--------|
| `app/utils/tab-id.server.ts` | **New** — AsyncLocalStorage for per-request tabId |
| `app/hooks/use-tab-id.ts` | **New** — Client hook: tabId generation + fetch interceptor |
| `server/index.ts` | Middleware to read X-Tab-Id header |
| `app/utils/emitter/send-notification.server.ts` | Auto-include tabId |
| `app/atoms/notifications.ts` | Add optional `tabId` to type |
| `app/routes/api+/sse.notification.ts` | Filter by tabId |
| `app/components/shared/toast.tsx` | Pass tabId to SSE URL |

## Test plan

- [x] Open two browser tabs side by side on the same page
- [x] Edit an asset in one tab
- [x] **Before fix**: toast appears on both tabs
- [x] **After fix**: toast only appears on the tab where the action was performed
- [x] Background notifications (reminders, scheduled jobs) still appear on all tabs
- [x] ESLint, Prettier, TypeScript all pass

### Evidence

**Before (bug):**

bug-toast

**After (fix):**

bug-toast-fix

> Screenshots show the toast appearing on both tabs (before) vs only the
> active tab (after). Actual images will be uploaded in the PR comments.

## Summary by CodeRabbit

* **New Features**
  * Notifications are now scoped to individual browser tabs—toast alerts and incoming notifications only appear for the active tab they belong to.
  * Client requests and server processing now carry tab context so notifications target the correct tab.

* **Tests**
  * Added unit tests ensuring tab-scoped notification isolation across concurrent tabs.

### DIFF
--- a/apps/webapp/app/hooks/use-tab-id.ts
+++ b/apps/webapp/app/hooks/use-tab-id.ts
@@ -1,15 +1,38 @@
+/**
+ * Per-tab identifier hook for browser tabs.
+ *
+ * Provides a stable, unique ID for the current browser tab and patches
+ * `window.fetch` to inject an `X-Tab-Id` header on same-origin requests.
+ * The server uses this header to scope SSE notifications to the originating tab.
+ *
+ * - Persists the ID in `sessionStorage` under {@link SESSION_KEY}
+ * - Generates a fresh UUID for cloned/opened tabs (detected via `window.opener`)
+ * - Mutates `window.fetch` to add the `X-Tab-Id` header automatically
+ *
+ * @see {@link file://./../components/shared/toast.tsx} — primary consumer
+ * @see {@link file://./../utils/tab-id.server.ts} — server-side counterpart
+ */
 import { useEffect, useRef } from "react";
 
 const SESSION_KEY = "__shelfTabId";
 
-/** Returns a stable, unique identifier for the current browser tab.
- * Patches `window.fetch` to send it as `X-Tab-Id` on same-origin requests. */
+/**
+ * Returns a stable, unique identifier for the current browser tab.
+ *
+ * **Side effects:**
+ * - Persists the tab ID to `sessionStorage` under `SESSION_KEY`
+ * - Patches `window.fetch` to send `X-Tab-Id` on same-origin requests
+ *
+ * @returns A stable unique string identifier for the current browser tab
+ */
 export function useTabId(): string {
   const tabIdRef = useRef("");
 
   if (!tabIdRef.current && typeof window !== "undefined") {
-    tabIdRef.current =
-      sessionStorage.getItem(SESSION_KEY) || crypto.randomUUID();
+    const stored = sessionStorage.getItem(SESSION_KEY);
+    // Cloned tabs (window.open / target="_blank") inherit sessionStorage,
+    // so generate a fresh ID when an opener is detected.
+    tabIdRef.current = stored && !window.opener ? stored : crypto.randomUUID();
     sessionStorage.setItem(SESSION_KEY, tabIdRef.current);
   }
 
@@ -43,7 +66,12 @@ export function useTabId(): string {
         })();
 
       if (isSameOrigin) {
-        const headers = new Headers(init?.headers);
+        // Seed from Request.headers when input is a Request, then overlay
+        // any headers from init so callers' overrides take precedence.
+        const headers = new Headers(
+          input instanceof Request ? input.headers : undefined
+        );
+        new Headers(init?.headers).forEach((v, k) => headers.set(k, v));
         if (!headers.has("X-Tab-Id")) {
           headers.set("X-Tab-Id", tabId);
         }
--- a/apps/webapp/app/utils/tab-id.server.test.ts
+++ b/apps/webapp/app/utils/tab-id.server.test.ts
@@ -1,3 +1,12 @@
+/**
+ * Unit tests for the tab ID server utilities.
+ *
+ * Verifies that {@link getTabId} and {@link runWithTabId} correctly scope
+ * tab identifiers using `AsyncLocalStorage`, ensuring session/tab isolation
+ * across concurrent async operations in the webapp server.
+ *
+ * @see {@link file://./tab-id.server.ts}
+ */
 import { describe, expect, it } from "vitest";
 import { getTabId, runWithTabId } from "./tab-id.server";
 
@@ -18,20 +27,23 @@ describe("tab-id.server", () => {
     });
   });
 
-  it("does not leak tabId across concurrent contexts", async () => {
+  it("does not leak tabId across concurrent async contexts", async () => {
     const results: Array<string | undefined> = [];
 
     await Promise.all([
-      new Promise<void>((r) =>
-        runWithTabId("tab-A", () => {
+      new Promise<void>((resolve) =>
+        runWithTabId("tab-A", async () => {
+          // Cross an async boundary so the context must survive a microtask hop
+          await Promise.resolve();
           results.push(getTabId());
-          r();
+          resolve();
         })
       ),
-      new Promise<void>((r) =>
-        runWithTabId("tab-B", () => {
+      new Promise<void>((resolve) =>
+        runWithTabId("tab-B", async () => {
+          await Promise.resolve();
           results.push(getTabId());
-          r();
+          resolve();
         })
       ),
     ]);


apps/webapp/app/hooks/use-tab-id.ts
```diff
@@ -0,0 +1,52 @@
+import { useEffect, useRef } from "react";
+
+const SESSION_KEY = "__shelfTabId";
+
+/** Returns a stable, unique identifier for the current browser tab.
+ * Patches `window.fetch` to send it as `X-Tab-Id` on same-origin requests. */
+export function useTabId(): string {
```
DonKoko, line 28: `useTabId` double-patch risk (app/hooks/use-tab-id.ts)

The fetch patch works correctly as long as this hook is mounted in exactly one component at a time. If `useTabId` is ever called from a second mounted component, the second mount captures the already-patched `window.fetch` as its `originalFetch`. When either component unmounts and restores its saved reference, the remaining component points at the wrong function. This is not an issue now, but I would rather add this simple guard to make it safer in the future when we have to use this hook in other components.

A simple guard prevents this:
```ts
window.fetch = function patchedFetch(...) { /* ... */ };
  (window.fetch as any).__shelfTabIdPatch = true;
```

And at the top of the effect, before patching:
```ts
  if ((window.fetch as any).__shelfTabIdPatch) return;
```
  Also worth noting — the hook comment says it "patches window.fetch" but that side effect is invisible at the call site in toast.tsx. A brief inline comment there would help future readers.

apps/webapp/app/utils/tab-id.server.ts
DonKoko, line 1: Would be nice to add some tests for this.

```ts
  import { getTabId, runWithTabId } from "./tab-id.server";

  it("returns the tabId inside the callback context", () => {
    runWithTabId("tab-123", () => {
      expect(getTabId()).toBe("tab-123");
    });
  });

  it("returns undefined outside any context", () => {
    expect(getTabId()).toBeUndefined();
  });

  it("does not leak tabId across concurrent contexts", async () => {
    const results: Array<string | undefined> = [];
    await Promise.all([
      new Promise<void>((r) => runWithTabId("tab-A", () => { results.push(getTabId()); r(); })),
      new Promise<void>((r) => runWithTabId("tab-B", () => { results.push(getTabId()); r(); })),
    ]);
    expect(results).toContain("tab-A");
    expect(results).toContain("tab-B");
  });
```

The concurrent isolation test is particularly worth having — it's the main guarantee `AsyncLocalStorage` provides.

apps/webapp/app/atoms/notifications.ts
```diff
@@ -16,6 +16,7 @@ export interface NotificationType {
   icon: NotificationIcon;
   time?: number;
   senderId: User["id"] | null;
+  tabId?: string;
```
DonKoko: This `tabId` ends up stored in the Jotai atom on the client, but the client never reads it — the SSE filter already did the routing work server-side. It's not harmful, but it leaks a server-routing concern into the client data model.

apps/webapp/app/hooks/use-tab-id.ts
```diff
@@ -0,0 +1,52 @@
...
+      input: RequestInfo | URL,
+      init?: RequestInit
+    ) {
+      const url =
+        typeof input === "string"
+          ? input
+          : input instanceof URL
+          ? input.href
+          : input.url;
+      const isSameOrigin =
```
DonKoko, line 58: `isSameOrigin` check is slightly permissive

```ts
url.startsWith(window.location.origin)
```

This would pass for https://app.shelf.nu.evil.com if the origin is https://app.shelf.nu.

Maybe something like this:
```ts
const isSameOrigin =
    url.startsWith("/") ||
    (() => {
      try { return new URL(url).origin === window.location.origin; }
      catch { return false; }
    })();
```

This is not high risk but i think its still worth tightening.

apps/webapp/app/hooks/use-tab-id.ts
DonKoko: Resolved in 92e5ad9 — added `__shelfTabIdPatch` guard flag and safe cleanup that only restores if our patch is still active.

apps/webapp/app/utils/tab-id.server.ts
DonKoko: Resolved in 92e5ad9 — added `tab-id.server.test.ts` with tests for basic usage, explicit undefined, and concurrent async context isolation.

apps/webapp/app/atoms/notifications.ts
DonKoko: Resolved in 92e5ad9 — removed `tabId` from `NotificationType`. Server adds it explicitly via `getTabId()` in `send-notification.server.ts`, and SSE filtering uses untyped JSON, so no downstream breakage.

apps/webapp/app/hooks/use-tab-id.ts
DonKoko: Resolved in 92e5ad9 — replaced `startsWith` with proper `new URL(url).origin === window.location.origin` comparison wrapped in try/catch.

apps/webapp/app/hooks/use-tab-id.ts
```diff
@@ -0,0 +1,67 @@
+import { useEffect, useRef } from "react";
+
+const SESSION_KEY = "__shelfTabId";
+
+/** Returns a stable, unique identifier for the current browser tab.
+ * Patches `window.fetch` to send it as `X-Tab-Id` on same-origin requests. */
+export function useTabId(): string {
```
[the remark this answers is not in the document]
DonKoko: Resolved in 27502a6 — added file-level JSDoc describing the module's purpose, responsibilities (`SESSION_KEY`, `window.fetch` patching, `X-Tab-Id`), and system context. Also expanded the function-level JSDoc with side effects and return value.

apps/webapp/app/hooks/use-tab-id.ts
```diff
@@ -0,0 +1,67 @@
...
+  const tabIdRef = useRef("");
+
+  if (!tabIdRef.current && typeof window !== "undefined") {
+    tabIdRef.current =
+      sessionStorage.getItem(SESSION_KEY) || crypto.randomUUID();
+    sessionStorage.setItem(SESSION_KEY, tabIdRef.current);
+  }
```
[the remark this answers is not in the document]
DonKoko: Resolved in 27502a6 — added `window.opener` check so cloned/`target="_blank"` tabs generate a fresh UUID instead of reusing the inherited sessionStorage value.

apps/webapp/app/hooks/use-tab-id.ts
```diff
@@ -0,0 +1,67 @@
...
+            return false;
+          }
+        })();
+
+      if (isSameOrigin) {
+        const headers = new Headers(init?.headers);
+        if (!headers.has("X-Tab-Id")) {
+          headers.set("X-Tab-Id", tabId);
+        }
+        return originalFetch.call(window, input, { ...init, headers });
```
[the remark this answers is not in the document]
DonKoko: Resolved in 27502a6 — now seeds `Headers` from `input.headers` when input is a `Request`, then overlays `init?.headers` on top, so existing request headers are preserved.

apps/webapp/app/utils/tab-id.server.test.ts
```diff
@@ -0,0 +1,43 @@
+import { describe, expect, it } from "vitest";
+import { getTabId, runWithTabId } from "./tab-id.server";
```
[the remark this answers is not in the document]
DonKoko: Resolved in 27502a6 — added file-level JSDoc describing test suite purpose, responsibilities, and `@see` reference to the source module.

apps/webapp/app/utils/tab-id.server.test.ts
```diff
@@ -0,0 +1,43 @@
...
+          results.push(getTabId());
+          r();
+        })
+      ),
+    ]);
+
+    expect(results).toContain("tab-A");
+    expect(results).toContain("tab-B");
+    expect(results).toHaveLength(2);
+  });
```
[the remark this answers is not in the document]
DonKoko: Resolved in 27502a6 — both branches now use `async` callbacks with `await Promise.resolve()` before reading `getTabId()`, so the test actually crosses a microtask boundary.

apps/webapp/app/routes/api+/sse.notification.ts
```diff
@@ -6,14 +6,23 @@ import { Logger } from "~/utils/logger";
 
 export function loader({ context, request }: LoaderFunctionArgs) {
   const authSession = context.getSession();
+  const url = new URL(request.url);
+  const connectionTabId = url.searchParams.get("tabId");
 
   return eventStream(request.signal, function setup(send) {
     /** Notification is a stringified json object with the shape {@link Notification} */
     function handle(notification: string) {
+      const parsed = JSON.parse(notification);
```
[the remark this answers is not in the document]
DonKoko: Not addressing this one — the `notification` string always comes from `sendNotification()` which calls `JSON.stringify()` internally, so it cannot be malformed. Adding defensive parsing here would be dead code.

DonKoko:
@iuryeng i took the initiative to finalize the last details. Feels great to have this bug that is opened since May 2024 be finally resolved. Thanks for your contribution

PR #2485, opened 2026-04-13
fix: use correct SEQUENCE value in iCal export

pj-gridx:
The SEQUENCE field should be a simple integer (0 for initial version) according to RFC 5545, not a timestamp from Date.now(). iCal Import into Google Calendar will fail otherwise.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed calendar event sequencing in exported .ics files to use proper revision tracking, ensuring calendar clients correctly interpret event updates.


DonKoko (approved):

DonKoko:
Thanks for the PR @pj-gridx

PR #2512, opened 2026-04-30
fix(webapp): mobile dialog footer + Radix Select selection on iOS Safari

carlosvirreira:
## TL;DR

Two iOS Safari bugs reported on iPhone 14:

1. **Asset-picker `Confirm` button was unreachable** when adding assets to a booking or kit on mobile.
2. **Role / asset-filter / booking dropdowns wouldn't select an item** on iPad/iPhone (intermittent earlier report).

Both have the same family of root causes: iOS Safari's viewport-height semantics (`100vh` resolves to the *Large* viewport, ignoring the URL bar) and a defensive Radix Select touch handler that was a little too aggressive. This PR fixes both, plus a small audit-driven sibling sweep covering the same patterns elsewhere — bundled because they share the same review context.

**14 files / +46 / −21 lines.** Typecheck ✅, lint ✅, 1777 unit tests ✅.

---

### Reproduction

- iPhone 14, iOS Safari & Chrome
- Open a booking → tap "Manage assets"
- Same flow on a kit: open kit → tap "Manage assets"
- Modal opens; the bottom action bar (`Close` / `Confirm`) is below the URL bar / home indicator and not tappable.

### Root cause

The fullscreen `<dialog>` element in [`apps/webapp/app/styles/global.css`](apps/webapp/app/styles/global.css) was styled `fixed h-full w-full` on mobile. On iOS Safari, `h-full` (= `100vh` of `<html>`) resolves to the **Large Viewport Height (LVH)** — i.e., the viewport with the URL bar collapsed. The manage-assets footer is positioned at the bottom of the dialog with flex `mt-auto`, so it sits at the bottom of LVH, **under the visible URL bar**, with no way to scroll to it.

Textbook iOS Safari viewport issue:

| Unit | Meaning | Behaviour with mobile URL bar |
|------|---------|-------------------------------|
| `100lvh` | Large viewport height | Largest possible — what `100vh` resolves to on iOS |
| `100svh` | Small viewport height | Smallest possible — always safe |
| `100dvh` | Dynamic viewport height | Tracks visible area as toolbars show/hide ✅ |

This affects every modal that flows through `<dialog class="dialog">`:

- [`apps/webapp/app/components/layout/dialog.tsx`](apps/webapp/app/components/layout/dialog.tsx) — the shared `<Dialog>` component (used by `InviteUserDialog`, etc.)
- [`apps/webapp/app/components/layout/contextual-modal.tsx`](apps/webapp/app/components/layout/contextual-modal.tsx) — wraps nested routes as modals (used by manage-assets / manage-kits / many others)

### Fix

Single CSS rule change in `global.css`:

```diff
 dialog {
-  @apply fixed m-0 h-full w-full bg-transparent p-0
-         md:relative md:m-auto md:h-auto md:w-[600px] md:py-8;
+  @apply fixed m-0 h-dvh w-full bg-transparent p-0
+         pb-[env(safe-area-inset-bottom)]
+         md:relative md:m-auto md:h-auto md:w-[600px] md:pb-0 md:py-8;
 }
```

(The hook auto-shortened `h-[100dvh]` → `h-dvh` on commit — identical CSS, just Tailwind 3.4+ built-in utility syntax.)

### Why this shape

- **`dvh`** is supported in Safari 15.4+, Chrome 108+, Firefox 101+ — covers all currently-supported iOS versions.
- **`env(safe-area-inset-bottom)`** resolves to `0` on non-iOS browsers, so non-iOS users see no change. On iPhones with a home indicator (`~34px`) the dialog gets that much padding-bottom, lifting the footer above the home indicator.
- **`md:` overrides reset both** — desktop continues to render `h-auto md:w-[600px] md:py-8` (centered, content-sized).
- All current `<dialog>` consumers benefit automatically. No JS changes, no component refactor, no API surface change.

### Verification

Built a temporary isolated test route (deleted before commit) that mounted `Dialog` + `Select` with mock content. Verified on a 375×812 (iPhone 14) viewport via computed-style and bounding-box inspection:

- Dialog: `h = 812px` (= dvh), `position: fixed`, `padding-bottom = 0` (Chrome's safe-area-inset = 0 in mobile emulation)
- Footer: `y = 762, height = 38` → ends at `y = 800`, well inside visible area
- Confirm button: `y = 771, height = 20` → fully reachable
- Desktop (resized to 2026×2023): dialog falls back to `h-auto`, `width: 600px`, `padding-bottom: 0` — no regression

---

### Reported on

- Team invite role dropdown (Settings → Team → Invite user)
- Asset picker filters (Booking → Manage assets → Status / Availability filters)
- Booking / kit selectors

### Root cause

[`apps/webapp/app/components/forms/select.tsx`](apps/webapp/app/components/forms/select.tsx) `SelectContent` had this listener:

```tsx
const handler = (event: TouchEvent) => event.preventDefault();
contentEl.addEventListener("touchend", handler, { passive: false });
```

It was added defensively to suppress iOS Safari's "ghost click" — the synthetic click event Safari fires ~300ms after a touch, on the element where the finger was lifted. Without this, after a user taps a `SelectItem` and the dropdown closes, the ghost click would fire on whatever was *behind* the dropdown (e.g., a button on the page below).

The problem: `touchend` bubbles. So the listener also calls `preventDefault()` on `touchend` events bubbling from `SelectItem` taps. While Radix selects on `pointerup` (which fires before `touchend`), several focus and `aria-activedescendant` transitions inside Radix Select still rely on the post-`touchend` synthetic click — and on certain Safari versions, suppressing that click intermittently breaks selection.

### Fix

Two parts:

**(a) Targeted bail-out in the touchend handler** ([`select.tsx`](apps/webapp/app/components/forms/select.tsx)):

```diff
 const handler = (event: TouchEvent) => {
-  event.preventDefault();
+  const target = event.target as HTMLElement | null;
+  if (
+    target?.closest(
+      '[role="option"], [data-radix-select-item], button, a, input, select, textarea, [role="checkbox"], [role="menuitem"]'
+    )
+  ) {
+    return;
+  }
+  event.preventDefault();
 };
```

Items keep their click; the ghost-click defense is preserved on the content's padding/empty area.

**(b) Tap-target hygiene**:

- Added `touch-manipulation` to `SelectTrigger` (kills iOS 300ms tap delay).
- Added `touch-manipulation` and `min-h-[40px] md:min-h-0` to `SelectItem` (close-to-Apple-HIG 44px tap target on mobile, no change on desktop).

### Verification

Synthetic touch-event sequences on the test fixture (with the new handler in place):

| Target | `event.defaultPrevented` after handler | Expected behaviour | Result |
|---|---|---|---|
| `[role="option"]` (`SelectItem`) | `false` | Handler bails out → synthetic click fires → selection works | ✅ matches |
| `SelectContent` padding (non-interactive area) | `true` | Handler runs → ghost click suppressed → no errant click on page below | ✅ matches |

End-to-end click flow: open trigger → items render → click item → trigger updates to `Base`, value state updates to `BASE`, dropdown closes. ✅

---

## Sibling sweep

After fixing the two reported bugs, I did a codebase audit for the same patterns. The full audit found ~30 sibling locations across 5 categories. **This PR bundles only the mechanical, single-call-site, low-risk ones** — same review context, defensible from the iOS LVH/DVH/safe-area model, no JS surface changes.

| File | Change | Why |
|---|---|---|
| [`maintenance-mode.tsx:22`](apps/webapp/app/components/layout/maintenance-mode.tsx) | `h-screen` → `h-dvh` | Same iOS-LVH bug. Affects every user during a maintenance window. CTA button is centered, so dvh is sufficient — no `safe-area-inset` needed (existing `py-16` already provides clearance). |
| [`shared/toast.tsx:87`](apps/webapp/app/components/shared/toast.tsx) | Added `pb-[calc(var(--viewport-padding)_+_env(safe-area-inset-bottom))]` | Toasts currently sit under the iPhone home indicator. Tailwind's longhand `pb-` overrides the shorthand `p-` for the bottom side. Used `calc()` (not `max()`) — `max(25px, 34px) = 34px` puts the toast right at the home-indicator boundary; `calc(25px + 34px) = 59px` gives 25px breathing room. On non-iOS, `env(...) = 0` → unchanged behaviour. |
| [`dynamic-select.tsx:465`](apps/webapp/app/components/dynamic-select/dynamic-select.tsx) | `touch-manipulation` on option `<div>` | Pure CSS property. Removes 300ms iOS double-tap-zoom delay. Matches Bug 2's `SelectItem` fix. |
| [`dynamic-dropdown.tsx`](apps/webapp/app/components/dynamic-dropdown/dynamic-dropdown.tsx) ×4 | `touch-manipulation` on all 4 option-row classNames (select-all, withValueItem, withoutValueItem, regular items) | Same. Used `replace_all` on a common substring — atomic update of all 4 sites. |
| 8 preview dialogs ([`code-preview-dialog`](apps/webapp/app/components/code-preview/code-preview-dialog.tsx), [`booking-overview-pdf`](apps/webapp/app/components/booking/booking-overview-pdf.tsx), [`kit-image`](apps/webapp/app/components/kits/kit-image.tsx), [`audit-receipt-pdf`](apps/webapp/app/components/audit/audit-receipt-pdf.tsx), [`asset-image/component`](apps/webapp/app/components/assets/asset-image/component.tsx), [`image-with-preview`](apps/webapp/app/components/image-with-preview/image-with-preview.tsx), [`report-pdf`](apps/webapp/app/components/reports/report-pdf.tsx), [`compliance-report-pdf`](apps/webapp/app/components/reports/compliance-report-pdf.tsx)) | `h-[90vh]` → `h-dvh` on mobile (`md:h-[calc(100vh-4rem)]` desktop preserved) | The original `90vh` was likely an intentional workaround leaving 10% margin to avoid iOS URL-bar overlap. `100dvh` is the proper fix — preview tracks the visible viewport and adapts as the URL bar shows/hides. |

### Deferred to follow-up PRs (with rationale)

These same iOS root causes appear in higher-risk locations that warrant separate scope:

| File / area | Severity | Why deferred |
|---|---|---|
| [`scanner/drawer/base-drawer.tsx`](apps/webapp/app/components/scanner/drawer/base-drawer.tsx) + [`hooks/use-viewport-height.ts`](apps/webapp/app/hooks/use-viewport-height.ts) | HIGH | Same root cause, but `useViewportHeight()` reads `window.innerHeight` (= LVH on iOS) and is consumed by other code. Changing the hook's semantics across all callers warrants its own scope and visual review. |
| [`booking-assets-sidebar.tsx`](apps/webapp/app/components/booking/booking-assets-sidebar.tsx) + [`contextual-sidebar.tsx`](apps/webapp/app/components/layout/contextual-sidebar.tsx) | MEDIUM-HIGH | `h-screen` inside Radix `Sheet` — interacts with Sheet's own height management & animations. Needs a Sheet-specific UX pass. |
| ~24 bulk-action fixed-bottom footers (assets / kits / booking / audit / location / category / tag) | MEDIUM | Mechanical ×24 sweep with same fix shape (`pb-[env(safe-area-inset-bottom)]`), but each flow (delete, archive, bulk-update, etc.) needs a quick visual confirm. Worth one focused PR. |

I have the full categorized audit list (Categories 1–5, with severities) and can break it out into tracking issues if useful.

---

## Verification methodology

The shared dev DB has drifted past `main` (`relation "_AssetToBooking" does not exist` for raw queries in [`apps/webapp/app/modules/asset/query.server.ts`](apps/webapp/app/modules/asset/query.server.ts)) — looks like another branch converted the implicit M2M to an explicit join model and migrated the shared DB. **I did not touch the database or run any migrations** — that's a separate investigation, flagged below.

Without DB access for the live booking/kit flows, I:

1. Built a temporary test route (`/mobile-test`, deleted before this commit) that mounted `Dialog` + `Select` with mock content + 30 spacer rows to ensure the body overflowed.
2. Verified mobile-viewport (375×812) layout via `getComputedStyle` + `getBoundingClientRect` (more precise than screenshot pixel-counting).
3. Drove a synthetic iOS-style touch sequence (`new TouchEvent('touchstart' / 'touchend')`) on options and on padding to confirm the `preventDefault` bail-out behaves as designed.
4. Round-tripped end-to-end click selection on the test route.
5. Resized to desktop (2026×2023) to confirm no regression — dialog falls back to `h-auto`, `padding-bottom: 0`, `width: 600px`.

**The fixes are reasoning-grounded from the iOS LVH/DVH/safe-area spec and verified on a Chrome mobile-emulation viewport — but Chrome does not perfectly simulate iOS Safari's `env(safe-area-inset-bottom)` values or its specific touch-event ordering quirks.**

### Recommended pre-merge smoke test (real iPhone, ~60 seconds)

- [ ] Open a booking → "Manage assets" → confirm `Confirm` button is reachable
- [ ] Open team settings → "Invite user" → tap role dropdown → confirm an option can be selected
- [ ] Open any asset image → confirm preview is fully visible (not clipped)

Optional but nice:
- [ ] iPad (different touch-handler version on iPadOS than iOS Safari)
- [ ] Older iOS (15.4 minimum for `dvh` — anything older needs a fallback we'd then need to add)

---

## Validation

| Check | Result |
|---|---|
| `pnpm turbo typecheck` | ✅ exit 0 |
| `pnpm webapp:lint` | ✅ exit 0 |
| `pnpm webapp:test -- --run` | ✅ 126 files / **1777 tests** / 36s |

---

## Notable decisions / context for review

1. **Why I didn't remove the touchend listener entirely on Bug 2.** It was added defensively for a real iOS issue (ghost click after dropdown closes onto whatever's behind it). Removing it brings back that bug. The bail-out preserves the defense for taps on padding while letting items receive their click — narrow scope, narrow fix.

2. **Why I didn't touch the nested `<div className="max-h-[320px] overflow-auto">` wrapper used in 13 SelectContent call sites.** Those are intentional height caps that Radix's `position="popper"` doesn't provide on its own. They're not the bug; the touchend handler was. Touching all 13 would balloon the diff with no functional benefit.

3. **`vite.config.ts` is intentionally NOT in this PR.** I added a `DISABLE_HTTPS` conditional locally to drive the dev server over HTTP for testing (matches the local-only edit you already carry in `shelf-main`). Worth landing as its own tiny PR upstream so future fresh clones don't need this dance — different concern.

4. **Dev DB schema drift, flagged separately.** `prisma migrate status` reports "up to date", but raw queries against `_AssetToBooking` (referenced in [`asset/query.server.ts`](apps/webapp/app/modules/asset/query.server.ts) lines 884–1023) fail because the table doesn't exist. Either some branch converted the M2M to an explicit model and migrated the shared DB, or `main` is behind. Worth investigating when you have a moment — it's the reason I couldn't do the live-route verification.

5. **Why `calc()` not `max()` on the toast safe-area.** `max(25px, 34px) = 34px` puts the toast right at the home-indicator boundary (functional but cramped). `calc(25px + 34px) = 59px` gives a 25px breathing-room buffer above the indicator (cleaner look). On non-iOS, `env(...)` resolves to 0 → both expressions reduce to 25px (no change).

---

## Test plan

- [ ] **iPhone 14**: asset-picker `Confirm` button reachable in both booking + kit flows
- [ ] **iPhone 14**: team invite role dropdown — tap an option, confirm selection registers
- [ ] **iPhone 14 / any iPhone with home indicator**: image / PDF preview dialogs use full visible viewport, not clipped
- [ ] **iPad**: dropdown selection works (different touch-handler version than iPhone)
- [ ] **Desktop Chrome / Firefox / Safari**: smoke test that no dialogs / dropdowns / previews regressed
- [ ] **Older iOS** (15.x, if available): confirm `dvh` and `env()` work — they're spec'd Safari 15.4+, worth a check

## Summary by CodeRabbit

* **Bug Fixes**
  * Dialogs/modals now use dynamic viewport height for more accurate mobile/iOS sizing and consistent responsive padding
  * Preview, image, and PDF dialogs updated for improved height/padding behavior across breakpoints
  * Sidebars and maintenance overlay adjusted to better match visible viewport on mobile
  * Dropdowns and selects receive improved touch handling to reduce accidental interactions on mobile
  * Toast placement now respects device safe-area insets for correct on-screen positioning


carlosvirreira:
@coderabbit full review

carlosvirreira:
Thanks @coderabbitai for the review. Walked through each finding against the current code; here's what's in fixup commit `4fac48657` and what was declined.

## ✅ Accepted

### `p-0` / `py-0` clobbers the dialog safe-area padding (real correctness bug)

CodeRabbit flagged this on `image-with-preview.tsx` only, but it applies to **all 8 preview dialogs** the sibling sweep touched. Class selectors (specificity `0,0,1,0`) beat element selectors (`0,0,0,1`), so a consumer-side `p-0` or `py-0` on the className passed to `<Dialog>` overrides the `.dialog` rule's `pb-[env(safe-area-inset-bottom)]` on mobile — making the safe-area addition I shipped ineffective for these dialogs. Their action buttons (`Close` / `Cancel` / `Download`) could still sit at the iPhone home-indicator boundary.

Scoped to `md:` so the rule remains effective on mobile:

| File | Change |
|---|---|
| code-preview-dialog.tsx | `p-0` → `md:p-0` |
| kit-image.tsx | `p-0` → `md:p-0` |
| asset-image/component.tsx | `p-0` → `md:p-0` |
| image-with-preview.tsx | `p-0` → `md:p-0` |
| booking-overview-pdf.tsx | `py-0` → `md:py-0` |
| audit-receipt-pdf.tsx | `py-0` → `md:py-0` |
| report-pdf.tsx | `py-0` → `md:py-0` |
| compliance-report-pdf.tsx | `py-0` → `md:py-0` |

### `select.tsx` `as HTMLElement` cast → `instanceof Element` guard

`event.target` is `EventTarget | null` — can be `Document` or `Window`, neither of which has `closest()`. The unsafe cast would throw if that ever happened in practice; the runtime guard skips the check and falls through to `preventDefault()` cleanly.

## ❌ Declined (with reasons)

### JSDoc backfill on `code-preview-dialog.tsx`, `image-with-preview.tsx`, `compliance-report-pdf.tsx`

These files all existed before this PR. My change to each was a single character (`90vh` → `dvh`). Adding JSDoc to functions and files I didn't author would shift this PR's scope from "iOS Safari mobile fixes" to "iOS Safari fixes + JSDoc backfill across pre-existing files".

CLAUDE.md does require JSDoc on "every new file and every new export", so this is real tech debt — but addressing it on these specific files just because I touched one line in each is drive-by enforcement that bloats the diff. Happy to do it as a separate PR if it's a priority.

### `Omit<HTMLProps<HTMLImageElement>, 'src' | 'alt' | ...>` on `ImageWithPreviewProps`

Defensive typing improvement that prevents callers from clobbering controlled props via `restProps`. Real benefit, but the loose typing existed before this PR and is unrelated to the iOS Safari fixes. Same scope reasoning — separate PR.

---

Re-ran validation after the fixup: typecheck ✅, lint ✅, **1777 tests ✅**.

Note for real-device verification: the safe-area padding effect can only be observed on actual iOS — Chrome mobile emulation reports `env(safe-area-inset-bottom) = 0`, so this fixup commit's behavioural change is invisible in headless tests but real on iPhone.

carlosvirreira:
## Critical update — original PR didn't actually fix the headline bug

**TL;DR**: I drove the real `/bookings/$id/overview/manage-assets` flow on a DB-compatible branch and discovered our original `<dialog>` fix doesn't reach this code path. Pushed `a38328e5e` with the actual fix.

### What happened

I'd been verifying against an isolated test fixture that mounted the same `<Dialog>` component used in some flows. The fixture passed — but when I drove the **actual** booking → manage-assets flow end-to-end, the DOM looked nothing like what my fix targets:

```
<div class="fixed z-50 ...">                       ← Radix <Sheet> outer
  <div class="flex h-screen w-full flex-col">     ← THIS is the bug
    <header>...</header>
    ...content...
    <footer class="mt-auto ...">Confirm</footer>  ← falls under URL bar on iOS
  </div>
</div>
```

This route is rendered via [`ContextualSidebar`](apps/webapp/app/components/layout/contextual-sidebar.tsx) (Radix Sheet), **not** [`ContextualModal`](apps/webapp/app/components/layout/contextual-modal.tsx) (`<dialog>`). The route's loader returns `showSidebar: true, noScroll: true`, which activates the Sheet wrapper, not the dialog wrapper.

The wrapping `<div className="flex h-screen w-full flex-col">` uses `h-screen` (= `100vh`). On iOS Safari, `100vh` resolves to the LARGE viewport (~877px on iPhone 14), so the `mt-auto` footer ends up below the visible viewport (~812px), under the URL bar. **Exactly the reported bug.**

The original commits on this PR (dialog viewport + safe-area + select touchend) still fix:
- The shared `<Dialog>` component (used by `InviteUserDialog` and other modal dialogs) ✅
- ContextualModal-wrapped routes ✅
- 8 preview dialogs ✅
- Radix Select dropdowns ✅

But they **didn't reach the actual asset-picker route**, which is the user-visible bug.

### How I caught it

Earlier verification was Chrome mobile-emulation against an isolated `/mobile-test` route. Chrome can't surface the iOS LVH/DVH mismatch (in Chrome, `100vh` = visible viewport, same as `100dvh`). So my fix appeared to work in headless. But on iOS, only `100dvh` tracks the visible viewport.

To see the real flow, I cherry-picked this PR onto a temporary branch off `feat-quantities` (which has the `BookingAsset` migration, making it compatible with the current dev DB schema), created a booking, and watched what actually rendered.

### The fix (commit `a38328e5e`)

[`apps/webapp/app/components/layout/contextual-sidebar.tsx`](apps/webapp/app/components/layout/contextual-sidebar.tsx) and [`apps/webapp/app/components/booking/booking-assets-sidebar.tsx`](apps/webapp/app/components/booking/booking-assets-sidebar.tsx):

```diff
- <div className="flex h-screen w-full flex-col">
+ <div className="flex h-dvh w-full flex-col">
```

Two-line change. Was originally flagged as MEDIUM-HIGH severity in the audit and **deferred to a follow-up PR** — promoted into this PR after live verification showed the previous fixes don't reach the actual bug path.

### Validation

- Live: confirmed the wrapper now renders `flex h-dvh w-full flex-col` on the real `/bookings/$id/overview/manage-assets` route
- Static: typecheck ✅, lint ✅, 1777 tests ✅

### What still needs verification

I have **not** verified on real iOS Safari yet — Chrome can't show the LVH/DVH difference. Spinning up a local dev server bound to LAN for iPhone testing now.

DonKoko (approved):

PR #2582, opened 2026-06-01
fix(webapp): enforce SELF_SERVICE custody self-restriction in shared services (P1)

carlosvirreira:
## P1 — within-org privilege escalation (mobile custody)

**SELF_SERVICE users could assign or release custody of *any* asset, to/from *any* team member, via the mobile API.** The web routes enforce a self-restriction (a self-service user may only touch their *own* custody), but the four mobile custody endpoints share the `asset:custody` permission and **skipped that check**:

| | Web (enforced ✅) | Mobile (was missing ❌) |
|---|---|---|
| Assign | `assets.bulk-assign-custody.ts` → 403 if custodian ≠ self | `mobile+/custody.assign.ts`, `mobile+/bulk-assign-custody.ts` |
| Release | `assets.bulk-release-custody.ts` / single → 403 if not own | `mobile+/custody.release.ts`, `mobile+/bulk-release-custody.ts` |

**Scope:** within-org integrity/authz bug (an insider with a SELF_SERVICE seat). **Not** a cross-org leak — multi-tenant org-scoping was already enforced and is unchanged. Found via the web↔mobile contract audit.

## Fix — one enforcement point, compiler-forced

Moved the self-restriction **out of the web route layer and into the shared services** — `bulkAssignCustody`, `bulkReleaseCustody`, `releaseCustody` — each now takes a **required `role` param**. Every caller (web + all 4 mobile routes) must supply it, so the rule lives in exactly one place and a future mobile route can't silently skip it. Removed the now-redundant inline web guards. (`releaseCustody` checks before its transaction so the 403 returns cleanly instead of being re-wrapped by the catch.)

This is the repo's own "org-scope guard" pattern applied to the role check, and the "ask the working sibling first" principle made structural.

## Tests
- **New** unit tests for all three service guards (deny path → 403, mutation never runs): `asset/service.server.test.ts`, `custody/service.server.test.ts`.
- **Updated** the web bulk-assign + mobile release route tests to reflect that enforcement now lives in the service (route forwards `role`).
- **Full webapp suite green: 2135 tests.** tsc / eslint / prettier clean.

## Notes
- Kit custody (`kit/service.server.ts`) is a separate path with its own SELF_SERVICE handling — out of scope, untouched (its tests still pass, including its own 403 assertion).
- Sibling to the audit-scan P0 (#2580) from the same web↔mobile seam audit. Independent; can merge in any order.

## Summary by CodeRabbit

* **Bug Fixes**
  * Self-service users can no longer assign or release custody of assets belonging to other team members. Asset custody operations are now restricted to self-owned assets only.

* **Improvements**
  * Unified custody permission enforcement across web and mobile platforms for consistent authorization behavior.

### DIFF
--- a/apps/webapp/app/modules/asset/service.server.test.ts
+++ b/apps/webapp/app/modules/asset/service.server.test.ts
@@ -1,3 +1,4 @@
+import { OrganizationRoles, type AssetIndexSettings } from "@prisma/client";
 import { describe, expect, it, vi, vitest, beforeEach } from "vitest";
 import { extractStoragePath } from "~/components/assets/asset-image/utils";
 import { db } from "~/database/db.server";
@@ -10,7 +11,9 @@ import { ShelfError } from "~/utils/error";
 import { createSignedUrl } from "~/utils/storage.server";
 import {
   bulkAssignAssetTags,
+  bulkAssignCustody,
   bulkDeleteAssets,
+  bulkReleaseCustody,
   bulkUpdateAssetCategory,
   getActiveCustomFieldsForAsset,
   parseAssetValuation,
@@ -45,6 +48,9 @@ vitest.mock("~/database/db.server", () => ({
     },
     qr: {
       update: vitest.fn().mockResolvedValue({}),
+    },
+    teamMember: {
+      findFirst: vitest.fn().mockResolvedValue(null),
     },
   },
 }));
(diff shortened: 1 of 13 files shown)


carlosvirreira:
@codex review

carlosvirreira:
@DonKoko — two small, independent fixes from the web↔mobile contract audit, both ready:

**#2580 — P0 (companion, iOS+Android):** the offline audit scan queue silently lost scans on flaky WiFi and marked those assets MISSING on completion — corrupting the audit-of-record. Client-only fix. ⚠️ **Needs an offline-device pass before merge** (airplane-mode scanning, kill+resume mid-sync, complete-with-pending-queue) — I can't exercise RN offline locally.

**#2582 — P1 (webapp):** SELF_SERVICE users could assign/release **anyone's** custody via the mobile API (the web routes blocked it). Moved the self-restriction into the shared services with a required `role` param + unit tests. No device QA needed — full webapp suite green (2135 tests).

Both: CodeRabbit + Codex clean, CI green, 0 open threads. Independent — merge in any order whenever you're happy. 🙏

DonKoko (approved):

PR #2583, opened 2026-06-01
feat(companion): audits redesign for ownership & urgency clarity

carlosvirreira:
## What & why

The companion's audit screens made it hard to answer the two questions an audit exists for — **who owns this work** and **what's urgent** — and had a dead-end that hid active work. This reworks the audit surfaces around ownership + urgency, fixes a per-asset status vocabulary bug, restores the Complete action on not-yet-started audits, and fixes an orange-on-white contrast failure.

Three companion commits (folded into one PR as agreed):
- `feat` audits redesign — the bulk of this PR
- `fix` legible audit evidence badge + clean paused scanner (earlier)
- `chore` point preview build profile at production (build config)

## Changes — redesign

**Audits list**
- Default to the full workspace, server-sorted by urgency (`dueDate asc, nulls last`) instead of "assigned to me" — kills the "no audits" dead-end when an active *unassigned* audit exists. BASE/SELF_SERVICE stay server-scoped; admins keep the optional "Assigned to me" toggle.
- First-class per-card ownership: **Unassigned · anyone can scan** / **Assigned to you** / **N assigned**.
- Relative due labels: **Overdue 2d / Due today / Due tomorrow / Due in 3d / Due {date} / No due date**.

**Audit detail**
- One **progress hero** (`X of Y found`, bar, `N pending/missing · N unexpected`) replaces the four stat boxes + separate bar.
- Ownership always shown, including the unassigned case (previously rendered nothing).
- **Vocabulary fix:** per-asset status is state-correct — `PENDING` while active, `MISSING` once completed. Previously the "Pending" and "Missing" filters resolved to the *same* rows. Filters are now context-aware.
- **Complete on PENDING:** show "Complete Audit" on not-yet-started audits (gate on status, not the ACTIVE-only server flag). `completeAuditSession` already accepts PENDING; the button was hidden, leaving a field worker who searched and found nothing with no action.

**Home**
- Active-audit cards use the same shared formatter + ownership line, so Home and the list no longer disagree on a deadline (was "Due 2 Jun" vs "Due tomorrow").

**Consistency / a11y**
- New `lib/audit-format.ts` — single source of truth for due-date wording across list, detail, and Home.
- New `primaryText` theme token: brand orange as body text on white was **3.1:1** (WCAG AA fail); now **5.4:1** light / **~7:1** dark. Icons keep the brighter brand orange (3:1 OK for graphics).

**API**
- Dashboard endpoint returns `assigneeCount` + `isAssignedToMe` for active audits (mirrors the audits list endpoint) so Home can show ownership. Companion treats them as **optional** and hides ownership gracefully against an older server (version-skew safe).

## Test plan

Verified on iOS simulator against real workspace data (+ one throwaway audit created → verified → deleted; workspace left clean):

| State | Verified |
|---|---|
| Active / unassigned / no due date | list + detail |
| Completed + missing | hero "N missing" (red), filters → All / Found / **Missing** / Unexpected, rows "Missing" |
| Completed + assigned-to-you | list |
| Active + assigned + due-soon | "Due tomorrow" (amber), "Assigned to you", "Start scanning" (PENDING) / "Continue scanning" (ACTIVE) |
| Urgency sort | due asc, nulls last (confirmed with 2 cards) |
| **Complete on PENDING** | full flow dogfooded: button → confirm ("N will be marked missing") → completed → rows missing |
| Home active-audit card | tappable → detail → back (anchored, no stranding) |
| Dark mode | both screens |
| Contrast | `primaryText` legible light + dark |

Companion typecheck: clean (0 errors).

## Reviewer notes / caveats
- ⚠️ **Webapp typecheck for `dashboard.ts` must be confirmed in CI** — the `react-router typegen` step couldn't run in the local mobile-only clone (env limitation, not the code). The change mirrors the existing audits-list endpoint's Prisma shape exactly.
- Deploy ordering: Home ownership lights up once the dashboard change is live; the companion degrades gracefully until then.

## Out of scope (follow-up)
- Per-asset "Can't find → mark missing" with optional note/photo evidence (a granular alternative to the bulk Complete). The bulk **Complete Audit** flow covers the core need today.

_Before/after screenshots (light + dark) captured during verification — can attach on request._

## Summary by CodeRabbit

* **New Features**
  * Redesigned audit detail with a single progress hero (found vs expected, percent complete) and clearer ownership display (assigned users or explicit "Open" state).
  * Server-driven completion eligibility now controls the "Complete Audit" CTA.

* **Improvements**
  * Completion-aware asset filters: PENDING vs MISSING distinctions, invalid filters reset, and visible filter pills reflect current audit state.
  * Standardized due-date tiers/wording and updated action hints ("Start scanning"/"Continue scanning").
  * Scan UI fully hides when paused; evidence badge and theme color refinements; preview build env added.


apps/companion/app/(tabs)/audits/[id].tsx
```diff
@@ -449,11 +454,26 @@ function AuditDetailContent() {
...
+
+  // Context-aware asset filters: a live audit has Pending items (not yet
+  // Missing); a completed one has Missing items (no longer Pending). Showing
+  // both at once was the source of two filters resolving to the same rows.
+  const visibleFilters = ASSET_FILTERS.filter((f) =>
+    f.value === "MISSING"
+      ? isCompleted
+      : f.value === "PENDING"
+      ? isActive
+      : true
```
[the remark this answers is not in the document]
Comment (carlosvirreira) [reply apps/companion/app/(tabs)/audits/[id].tsx]: Fixed in e38c33b5a. Extracted a shared `isAssetFilterVisible(value, status)` helper used by **both** the filter pills and a new effect that resets `assetFilter` to `ALL` whenever the current selection isn't valid for the audit's status (PENDING drops on completion, MISSING on re-activation). So the selection can never strand on a hidden tab with an empty list. Typecheck clean. Good catch 👍

carlosvirreira:
@codex review

apps/companion/lib/audit-format.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!dueDate) return { label: "No due date", tier: "none" };
+  const dayMs = 24 * 60 * 60 * 1000;
+  const ms = new Date(dueDate).getTime() - Date.now();
+  if (ms < 0) {
+    const d = Math.max(1, Math.ceil(-ms / dayMs));
+    return { label: `Overdue ${d}d`, tier: "overdue" };
+  }
+  const days = Math.round(ms / dayMs);
+  if (days === 0) return { label: "Due today", tier: "soon" };
+  if (days === 1) return { label: "Due tomorrow", tier: "soon" };
```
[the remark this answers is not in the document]
carlosvirreira: Fixed in 629a62437. Switched `formatDue` to compare **local calendar days** (start-of-day to start-of-day) instead of `Math.round(ms / day)`, so a due *time* never bleeds into the wrong day — "Due today" / "Due tomorrow" / "Due in Nd" are now calendar-accurate. Also shows a bare "Overdue" when past the due moment but still the same day (vs forcing "Overdue 1d"). Typecheck clean. Nice catch 👍

carlosvirreira:
@codex review

apps/companion/app/(tabs)/audits/[id].tsx
```diff
@@ -719,7 +797,16 @@ function AuditDetailContent() {
...
+            {/*
+              why: gate on `isActive` (PENDING || ACTIVE), not the server's
+              `canComplete` flag — that flag is ACTIVE-only, which hid this
+              button on a fresh PENDING audit, leaving a field worker who
+              searched and found nothing with no action but "Start Scanning".
+              `completeAuditSession` already accepts a PENDING audit (it only
+              rejects COMPLETED/CANCELLED/ARCHIVED) and marks unscanned assets
+              MISSING — the server is the real gate, so this is pure UI.
+            */}
+            {isActive && (
```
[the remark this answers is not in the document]
Comment (carlosvirreira) [reply apps/companion/app/(tabs)/audits/[id].tsx]: Fixed in c9c53586f. Made the detail endpoint's `canComplete` **authoritative** instead of gating client-side: it now returns true only when the audit is PENDING/ACTIVE **and** the caller is eligible — an assignee, or an admin/owner when the audit has no assignees — mirroring `requireAuditAssignee` in audits.complete.ts exactly. The client just consumes the flag again, so the CTA shows precisely when completion will succeed: keeps the PENDING-visibility fix, drops the 403-after-confirm on someone else's audit. Good catch 👍

carlosvirreira:
@codex review

apps/companion/app/(tabs)/audits/[id].tsx
```diff
@@ -98,6 +117,19 @@ function AuditDetailContent() {
...
   const [assetFilter, setAssetFilter] = useState<AssetFilterValue>("ALL");
 
+  // why: when the audit's status changes while this screen is open — it
+  // completes (here, or via another user + a focus refetch) or re-activates —
+  // a previously-selected filter can vanish from the pills (PENDING drops on
+  // completion, MISSING on re-activation). Left as-is, `filteredAssets()` then
+  // matches a status that no longer exists and shows an empty list with no
+  // active tab. Snap the selection back to "All" so it's always a visible
+  // pill. (Codex review, PR #2583.)
+  useEffect(() => {
```
Comment (DonKoko) [inline apps/companion/app/(tabs)/audits/[id].tsx]: This doesn't seem right. It has `assetFilter` as dependency and changes `assetFilter`. Also it misses dependencies.

DonKoko (changes requested):
I flagged 1 specific thing that I think will break. I recommend running react doctor on the companion app's changed files to see what it flags.

apps/companion/app/(tabs)/audits/[id].tsx
Comment (carlosvirreira) [reply apps/companion/app/(tabs)/audits/[id].tsx]: Good catch — you're right, that effect both depended on and wrote `assetFilter`, and keyed off the whole `audit` object (new identity on every focus refetch), so it re-ran constantly. Fixed in `9b3f70251` by removing the effect entirely and **deriving** the applied filter at render:

```ts
const effectiveFilter: AssetFilterValue =
  audit && isAssetFilterVisible(assetFilter, audit.status) ? assetFilter : "ALL";
```

`effectiveFilter` is what drives `filteredAssets`, the active pill, and the empty-state copy; the raw `assetFilter` is only ever *written* by the pill `onPress`. No state-syncing effect, no depend-and-set, no missing deps — and when a hidden filter becomes valid again the selection naturally reapplies (the old effect permanently snapped it to ALL). Typecheck passes.

Re: react doctor — ran it on the changed companion files (`--diff origin/main`). Triage posted as a PR comment below; tl;dr none of its findings are introduced by this PR (they're app-wide pre-existing patterns + documented `eslint-disable` focus-effect suppressions), so I've kept this PR scoped to the bug you flagged rather than bundling an app-wide refactor.

carlosvirreira:
### react-doctor triage (companion changed files)

Ran `react-doctor v0.2.16 --lint --diff origin/main` over the 9 companion files this PR touches. **38 findings — none introduced by this PR.** Breakdown:

**1. Missing effect dependencies ×9** — every one sits on a line already carrying `// eslint-disable-next-line react-hooks/exhaustive-deps` + a `// why:` (the focus-effects intentionally key off `currentOrg?.id` only, to avoid refetch storms on identity-only org changes). react-doctor doesn't honor `eslint-disable` (noted in our CLAUDE.md), so it re-flags deliberate, documented suppressions. The one effect that was a genuine anti-pattern — the `assetFilter` reset you flagged — is **removed** in `9b3f70251` (derived value, no effect).

**2. App-wide RN conventions (pre-existing, present across the whole companion app):**
- Touchable→Pressable ×5
- JS-thread `Animated`→Reanimated ×3
- Giant component ×4 / many-useState→useReducer ×3
- Lazy ref init `new Animated.Value(0)` ×2
- JSX-as-prop (`refreshControl={<RefreshControl/>}`) ×3 — idiomatic RN, effectively a false positive for `RefreshControl`

**3. Cosmetic ×3:** `…` vs `...` in a loading string, one module-scope-able `formatShortDate` helper.

**Recommendation:** keep this PR scoped to the filter bug. The app-wide items (Pressable + Reanimated migrations especially) are a worthwhile but separate cleanup — happy to open a follow-up issue/PR for the companion as a whole rather than half-migrate it inside an audits-redesign PR. react-doctor's own output says the same: *"ignore … unrelated pre-existing code."*

If you'd rather I fold any specific subset in here, point me at which and I'll do it. Otherwise this is ready for another look — @carlosvirreira could you kick `@codex review` on the new commit?

apps/companion/app/(tabs)/audits/index.tsx
```diff
@@ -81,18 +57,24 @@ function AuditsListContent() {
...
+  // landing on an empty "nothing assigned to you" screen — when in fact
+  // there were active, unassigned audits anyone could pick up. The list
+  // is server-sorted by due date (most urgent first) and every card now
+  // states its ownership ("Unassigned · anyone can scan" / "Assigned to
+  // you" / "N assigned"), so the urgency-first flat list is legible
+  // without forcing a scope filter. Admins can still narrow to their own
+  // work with the toggle below. BASE/SELF_SERVICE are server-scoped to
+  // their assignments regardless, and the effect below keeps their
+  // (hidden) toggle pinned to true so visible state matches the server.
+  const [assignedToMe, setAssignedToMe] = useState(false);
```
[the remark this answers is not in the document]
carlosvirreira: Good catch — confirmed and fixed in `b2b5f16f0`. `userCanSeeOrgWideAudits(undefined)` returns `false`, so on a cold mount (roles not yet loaded) the effect flipped `assignedToMe → true`, and since it only ever snaps *to* true, an ADMIN/OWNER never got reset to the full-workspace default once their role arrived — exactly the empty-list dead-end this screen removes.

Fix gates the snap on roles actually being loaded:

```ts
const rolesKnown = Array.isArray(currentOrg?.roles);
useEffect(() => {
  if (rolesKnown && !canWidenScope && !assignedToMe) {
    setAssignedToMe(true);
  }
}, [rolesKnown, canWidenScope, assignedToMe]);
```

While roles are unknown the `false` default holds; once they load, ADMIN/OWNER stay on the full list (`canWidenScope` true → no-op) and BASE/SELF_SERVICE snap to true to match the server's forced scoping. Mid-session downgrade still works since `rolesKnown` stays true. Typecheck passes.

apps/companion/eas.json
DonKoko, line 1: Why are we adding the URLs and Public key to the app code in this config file? Those are supposed to be just env variables.

DonKoko (approved):

apps/companion/eas.json
carlosvirreira: Agreed — makes sense to keep them out of the repo even though they're public. Since the production block is already on main and moving these to EAS env vars touches the EAS project, I'll handle it as a dedicated chore: create `EXPO_PUBLIC_API_URL`, `EXPO_PUBLIC_SUPABASE_URL` and `EXPO_PUBLIC_SUPABASE_ANON_PUBLIC` as EAS environment variables (preview + production), then strip the inline `env` blocks from `eas.json` across all profiles so nothing's hardcoded. That part needs EAS project access — @carlosvirreira can run `eas env:create`, or point me at it. Good to keep #2583 scoped to the audits work and land this cleanup in its own PR?

carlosvirreira: Done — opened #2588: the three values are now EAS environment variables (preview + production) and the inline `env` blocks are removed from `eas.json`. Nothing hardcoded in the repo anymore (dev keeps localhost).

PR #2587, opened 2026-06-02
ci(react-doctor): extend react-doctor to the companion app

DonKoko:
Fan the React Doctor GitHub Action out over a matrix of [webapp, companion] so the Expo/React Native code gets the same PR-time health check the webapp already has. react-doctor auto-detects React Native and applies RN-specific rules (deprecated modules, reanimated, FlatList renderItem, Dimensions.get, etc.) on top of the generic React hygiene rules.

- workflow: matrix over both apps (fail-fast: false), parameterized working-directory, path-prefix, and per-app sticky comment header so the two jobs keep separate PR comments. Same gate as webapp: --diff + --fail-on error (errors block, warnings advisory).
- pr-comment script: accepts a path-prefix arg, derives the app name and <app>:doctor command, renders repo-root-relative links. Webapp default unchanged.
- companion: pin react-doctor@0.0.39 (matches webapp) + add doctor script; add companion:doctor root convenience script.
- fix the 2 existing error-level findings so companion starts green: import SafeAreaView from react-native-safe-area-context instead of react-native in team-member-picker and location-picker.
- docs: document companion:doctor and correct the stale "not a CI gate" note (react-doctor does gate PRs on newly-introduced errors).

## Summary by CodeRabbit

* **New Features**
  * CI health scan now runs per-app (webapp and companion) with separate, app-scoped PR comments and correct repo-relative links.
  * New local npm scripts to run health scans for each app (webapp and companion).

* **Bug Fixes**
  * Adjusted React Native safe-area usage to the recommended provider.

* **Documentation**
  * Docs and guidance updated to cover companion app scan and CI behavior (failures only on newly introduced errors).

### DIFF
--- a/.github/scripts/react-doctor-pr-comment.mjs
+++ b/.github/scripts/react-doctor-pr-comment.mjs
@@ -5,29 +5,45 @@
  * filters out the knip dead-code findings (out of scope for PR review),
  * and prints a Markdown comment body to stdout.
  *
- * Invoked from `.github/workflows/react-doctor.yml` after the doctor scan.
+ * Invoked from `.github/workflows/react-doctor.yml` (once per app in the
+ * matrix) after the doctor scan.
  *
  * Usage:
- *   node .github/scripts/react-doctor-pr-comment.mjs <diagnostics-dir> <exit-code>
+ *   node .github/scripts/react-doctor-pr-comment.mjs \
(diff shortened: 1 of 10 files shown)


DonKoko:
@coderabbit can you make a review considering the last updates I did.

PR #2602, opened 2026-06-05
feat(bookings): subscribable iCal calendar feed

carlosvirreira:
## Subscribable iCal calendar feed for bookings

Lets a workspace member subscribe their Google / Apple / Outlook calendar to a **live feed of their Shelf bookings** via a secret URL, so bookings appear (and stay updated) in the calendar they already use.

### Why

Direct customer signal: a live trial (photography/media equipment) asked specifically for Shelf Bookings ↔ Google Calendar — *"automatic synchronization … or API/webhook options so we can build our own."* Calendar sync is also table-stakes for booking/rental tools. This is the smallest, highest-demand slice of the integrations roadmap.

### What it does

- **Public feed route** `GET /api/calendar/feed/:token.ics` — authenticated by a secret URL token (calendar clients can't send cookies), registered in `publicPaths`, **IP rate-limited**, served inline as `text/calendar`.
- **Token** lives on `UserOrganization.calendarTokenId` (`@unique`, no new model). Generate / regenerate / revoke via `POST /api/calendar-subscription` and a **Subscribe** dialog on the calendar page (copy URL, `webcal://` add, regenerate, stop sharing).
- **Scoping mirrors the in-app `/calendar`**: members who can see all bookings get the whole workspace (so an admin can publish a shared team calendar); self-service/base members get only their own. Custodian names are hidden when the member isn't entitled. DRAFT/ARCHIVED/CANCELLED excluded; windowed ~−1 month → +1 year.
- **Refactor:** extracted `buildBookingVEvent` / `buildBookingICalendar`; the existing per-booking "Add to calendar" download now reuses them — **byte-identical output** (characterization test included).

### Tiering

Free for all **Team** workspaces (bookings are already Team-gated, so the audience is already paid); self-host free via the `premiumIsEnabled` kill-switch. The future paid line is two-way Calendar-API sync, not this read-only feed.

### Migration

Additive, non-breaking: one nullable `calendarTokenId TEXT` column + a unique index on `UserOrganization`. The migration SQL was hand-authored to match the schema change (regenerate with `prisma migrate dev` against a dev DB if preferred). Rollback is trivial.

### Security

Ran an independent adversarial security review before pushing. Verdict: **safe to release**, no cross-org IDOR — `organizationId` is never user-supplied (it comes from the token's row and is the hard base `WHERE` clause; complies with `org-scope-user-supplied-ids`). One required fix was applied: **added `calendarIpRateLimit`** on the public route (it previously had none).

Conscious, documented decisions:

- **Token stored plaintext** — a feed URL must be re-displayable when the member returns to copy it (hashing would force "shown once," wrong for an iCal feed; Google stores its secret iCal address the same way). Mitigated by rotate/revoke + a limited blast radius (one member's bookings).
- **`takeAll`** kept (a `perPage` cap would silently drop bookings) — bounded by the window + single-org scope + the rate limit.
- `resolveCalendarVisibility` duplicates the `roles.server.ts` derivation (documented; a good future cleanup is to extract one shared helper).

### Testing

- New unit tests: iCal builders, **feed scoping** (proves a restricted member is forced to `custodianUserId = userId` — can only narrow), token lifecycle, visibility truth-table, URL helpers.
- `pnpm webapp:validate` run locally: Prettier + ESLint + typecheck pass; the full suite passes except 5 unrelated scanner/editor/activity-csv suites that **timed out under full-suite load** — they pass in isolation (verified). CI should run them in a cleaner environment.

### Recommended before merge

- A quick manual subscribe in a real Google Calendar (the one thing not done end-to-end live).
- Let CodeRabbit / Codex / the commit-time security agent do their pass.

### Reviewer notes

- Public route + token resolution: `app/routes/api+/calendar.feed.$token[.ics].ts`, `app/modules/calendar-subscription/service.server.ts`.
- Scoping: `getBookingsForICalFeed` in `app/modules/booking/service.server.ts`.

## Summary by CodeRabbit

* **New Features**
  * Added calendar booking subscription from the Calendar page with “Subscribe”, “Regenerate”, and a safer revoke flow (“Stop sharing” → “Yes, stop”).
  * iCal feed links are now generated and copyable, with clearer inline errors during subscription updates.
* **Refactor**
  * Standardized iCal (.ics) generation for consistent VEVENT/VCALENDAR formatting.
* **Bug Fixes**
  * Improved iCal feed visibility and status scoping, including correct handling for regenerated and revoked subscriptions.
* **Tests**
  * Added automated coverage for iCal formatting and subscription/booking scoping rules.
* **Chores**
  * Added per-subscription token handling and rate limiting for iCal feed requests.


carlosvirreira:
Thanks for the thorough review — all findings addressed in `47d1903ea`. Summary:

### Fixed
- **Secret token logged (Hex 🟡 / CodeRabbit 🔴):** removed `{ token }` from the `makeShelfError` call in the feed route's catch block — the token is the sole credential and `additionalData` is written to logs + Sentry. The cause chain still identifies the failure.
- **Team-member-scoped bookings omitted (Codex P2 / CodeRabbit 🟠):** `getBookingsForICalFeed` now passes `custodianTeamMemberIds: [teamMember.id]` alongside `custodianUserId`, so `getBookings` ORs the two (the documented legacy case where a booking is on the team member with a null `custodianUserId`). Verified safe — it can only ever *narrow* to the member's own bookings. Scoping test updated to assert the OR. _(Note: the in-app `/calendar` has the same latent gap — it spreads the singular `custodianTeamMemberId`, which `getBookings` ignores — but fixing the calendar itself is out of scope here.)_
- **DTSTAMP churns every poll (CodeRabbit 🟠):** `buildBookingVEvent` now uses the booking's `updatedAt` for `DTSTAMP` instead of `new Date()`, so an unchanged booking serializes identically across polls. Both call sites (feed + per-booking download) pass it. `SEQUENCE` stays `0` (constant → no churn; a changed `DTSTAMP` on edit is enough for clients to re-sync).
- **Auth-bypass / rate-limit scope (CodeRabbit 🟠 ×2):** tightened the `publicPaths` entry and the middleware mount to `/api/calendar/feed/*` only (the cookie-authed `/api/calendar-subscription` stays out of the bypass), and re-keyed the limiter by the feed **path/token** (renamed `calendarFeedRateLimit`) instead of client IP — so shared Google/Apple/Outlook egress IPs aren't cross-throttled and a leaked URL is capped per-feed. Generic invalid-token floods are cheap indexed 404s, best absorbed at the edge.
- **Missing JSDoc (CodeRabbit nitpicks):** added function-level JSDoc to the dialog component, the action, and the loader.
- **Import order (ESLint CI):** fixed.

### Not changed
- _Extract a booking fixture factory in the tests_ — left the inline fixtures as-is; for these small, focused suites the explicit data reads clearer than a shared factory. Happy to refactor if preferred.

Lint, typecheck, and the calendar test suites pass locally.

carlosvirreira:
@coderabbit full review

Issue #2611, opened 2026-06-08
Harden mobile SSO handoff transport (PKCE / per-device verifier + Universal Links)


DonKoko:
### Context
PR #2607 introduced web-delegated SSO for the companion app. The server side is solid: a 256-bit, hash-stored, single-use, 60s-TTL authorization code bound to a server-verified `userId`, exchanged over a back channel at `/api/mobile/exchange` (rate-limited via `/api/mobile/*` → `mobileIpRateLimit`, 30/min/IP).

The residual risk is the **mobile transport layer**, deferred in the original PR:

1. The code is delivered to the app over the `shelf://auth-callback?code=` **custom URI scheme**. On Android (and older iOS) any installed app can register the same scheme. `WebBrowser.openAuthSessionAsync` keeps the code inside the auth-session sandbox in the happy path (return-value binding), but if the redirect ever fires outside that managed session, the deeplink hits the OS dispatcher and is interceptable.
2. **PKCE is not implemented** — RFC 8252 (OAuth for Native Apps) calls this out as the primary defense for native handoffs.

### Proposed work
- [ ] **Per-device verifier (smallest, highest-value change):** app generates a random `verifier`, sends `sha256(verifier)` to `/sso-login`; persist the hash on the `MobileAuthCode` row; require the matching `verifier` at `/api/mobile/exchange`. An intercepted code becomes unredeemable by any other app. (PKCE-equivalent, minimal schema add.)
- [ ] Migrate the callback from custom scheme (`shelf://`) to **Universal Links (iOS) / App Links (Android)** so the OS verifies app ownership of the domain and no other app can claim the callback.
- [ ] Confirm `openAuthSessionAsync` return-value binding behavior on each target OS version (Android scheme-collision in particular).

### Acceptance
- An authorization code intercepted by a hostile app cannot be redeemed.
- Callback delivery no longer relies on an unverifiable custom URI scheme.

### References
- PR #2607
- RFC 8252 (OAuth 2.0 for Native Apps), RFC 7636 (PKCE)
- `apps/companion/lib/web-auth.ts`, `apps/webapp/app/routes/_auth+/oauth.callback.mobile.tsx`, `apps/webapp/app/routes/api+/mobile+/exchange.ts`, `apps/webapp/app/modules/auth/mobile-sso.server.ts`

_Surfaced during security review of PR #2607._

carlosvirreira:
[...]

PR #2619, opened 2026-06-10
fix(mobile-sso): prod hardening — Sentry 5xx capture, migrate-status gate, wider code TTL

carlosvirreira:
## Why

Follow-up hardening for mobile SSO (merged in #2607), making the prod rollout safe to pilot. Three gaps:

1. **The mobile SSO server flow was ~invisible to Sentry.** Both `exchange.ts` and `oauth.callback_.mobile.tsx` catch every error and `return data({ error })` without re-throwing or logging. Since `Sentry.captureException` only fires via `Logger.error` (or an uncaught throw), even genuine 5xx — a Supabase mint outage, a broken auth contract, a DB/migration fault — never reached Sentry. Confirmed empirically: the prod migration-drift 500 (missing `MobileAuthCode` table) paged no one and was only found by manual probing.

2. **A migration can silently fail to apply on deploy.** `fly.toml`'s `release_command = prisma migrate deploy` applies migrations during deploy, but if it ever exits `0` without applying, the deploy still goes green and the missing table only surfaces later as runtime 500s.

3. **The single-use code TTL was tight for edge cases.** The code is minted in the web callback *after* SSO/MFA, so 60s only had to cover the `shelf://` hand-back + the exchange POST — but iOS briefly backgrounding the app during hand-back, or a slow-network exchange, could still expire it.

## What

- **`exchange.ts` / `oauth.callback_.mobile.tsx`** — in the `catch`, log `status >= 500` via `Logger.error` (→ Sentry) and keep a sampled, non-alerting trail of the expected 4xx (expired / invalid / replayed code, SSO disabled, malformed payload) via `Logger.handledClientError`. Success path unchanged.
- **`deploy.yml`** — after the production deploy, run `prisma migrate status` on the running machine over `flyctl ssh console` and fail the workflow loudly if the schema isn't in sync. So a `release_command` that exits `0` without applying a migration is caught here, not by a user.
- **`mobile-sso.server.ts`** — widen the single-use code TTL `60s → 180s`. No security impact (single-use, hashed); just margin for the deeplink hand-back / slow exchange. The comment is corrected to note the code is minted *after* SSO/MFA, so this window never spanned the human-paced login.

## Why this is low-risk

- Additive observability + a wider (still short, single-use) code window — success path and response shapes unchanged.
- The CI gate **reuses the existing `FLY_API_TOKEN`** and runs **on the prod machine**, which already carries `DATABASE_URL`/`DIRECT_URL` as Fly secrets — no new secrets, no DB creds in CI.
- The gate is **post-deploy**: a failure is an alert to investigate, never a rollback or outage.

## Testing

- Pre-commit hook green: typecheck ✅, eslint ✅, prettier ✅, commitlint ✅
- TTL change: no test pins the value (tests assert `expiresAt > now`); behavior unchanged beyond the wider window.
- `deploy.yml` validated as YAML; new step parses as `🔎 Verify migrations applied (production)`.

## One thing to confirm on the first prod run

The exact `flyctl ssh console -C "/bin/sh -lc 'cd /src/apps/webapp && npx prisma migrate status'"` invocation can't be exercised from CI here. Please eyeball the step's output on the first `main` deploy — it greps for `Database schema is up to date`. If `flyctl ssh` / `npx prisma` resolves differently on the machine, the command is easy to adjust (the grep + fail-safe logic is already defensive).

## Summary by CodeRabbit

* **New Features**
  * Extended mobile authentication code validity from 60 seconds to 3 minutes.

* **Bug Fixes**
  * Improved observability for mobile OAuth callback and SSO token exchange by explicitly logging handled exceptions.
  * Added security-focused error handling to avoid exposing sensitive refresh-token data in logs/telemetry.
  * Enhanced monitoring by redacting sensitive fields in tracking “extra” details, including safe handling of nested/circular data.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -116,33 +116,3 @@ jobs:
         run: flyctl deploy --remote-only --build-arg COMMIT_SHA=${{ github.sha }} --app ${{ steps.app_name.outputs.value }} --dockerfile apps/webapp/Dockerfile
         env:
           FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
-
-      # Migrations apply via fly.toml's `release_command` (`prisma migrate
-      # deploy`) during the deploy above. If that ever exits 0 without applying
-      # (e.g. a schema-path/env hiccup at release time), the deploy still goes
-      # green and a missing table only surfaces later as runtime 500s — which is
-      # how a prod migration once drifted unnoticed. This gate asks the running
-      # production machine whether Prisma considers the schema in sync and fails
-      # loudly otherwise, so drift is caught here rather than by a user. It
-      # reuses FLY_API_TOKEN (no DB credentials in CI) and runs on the machine,
-      # which already carries DATABASE_URL/DIRECT_URL as Fly secrets. It runs
-      # after deploy, so a failure is an alert to investigate — it does not roll
-      # the release back.
-      - name: 🔎 Verify migrations applied (production)
-        if: ${{ github.ref == 'refs/heads/main' }}
-        run: |
-          set -uo pipefail
-          echo "Checking 'prisma migrate status' on production…"
-          OUTPUT=$(flyctl ssh console --app ${{ steps.app_name.outputs.value }} \
-            -C "/bin/sh -lc 'cd /src/apps/webapp && npx prisma migrate status'" 2>&1) || true
-          echo "--- prisma migrate status ---"
-          echo "$OUTPUT"
-          echo "-----------------------------"
-          if echo "$OUTPUT" | grep -q "Database schema is up to date"; then
-            echo "✅ Production database schema is up to date."
-            exit 0
-          fi
-          echo "::error::Prisma reports the production schema is NOT up to date — a migration may have failed to apply during release. Run 'npx prisma migrate status' / 'migrate deploy' against production and investigate before treating this deploy as healthy."
-          exit 1
-        env:
-          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -116,3 +116,33 @@ jobs:
         run: flyctl deploy --remote-only --build-arg COMMIT_SHA=${{ github.sha }} --app ${{ steps.app_name.outputs.value }} --dockerfile apps/webapp/Dockerfile
         env:
           FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
+
+      # Migrations apply via fly.toml's `release_command` (`prisma migrate
+      # deploy`) during the deploy above. If that ever exits 0 without applying
+      # (e.g. a schema-path/env hiccup at release time), the deploy still goes
+      # green and a missing table only surfaces later as runtime 500s — which is
+      # how a prod migration once drifted unnoticed. This gate asks the running
+      # production machine whether Prisma considers the schema in sync and fails
+      # loudly otherwise, so drift is caught here rather than by a user. It
+      # reuses FLY_API_TOKEN (no DB credentials in CI) and runs on the machine,
+      # which already carries DATABASE_URL/DIRECT_URL as Fly secrets. It runs
+      # after deploy, so a failure is an alert to investigate — it does not roll
+      # the release back.
+      - name: 🔎 Verify migrations applied (production)
+        if: ${{ github.ref == 'refs/heads/main' }}
+        run: |
+          set -uo pipefail
+          echo "Checking 'prisma migrate status' on production…"
+          OUTPUT=$(flyctl ssh console --app ${{ steps.app_name.outputs.value }} \
+            -C "/bin/sh -lc 'cd /src/apps/webapp && npx prisma migrate status'" 2>&1) || true
+          echo "--- prisma migrate status ---"
+          echo "$OUTPUT"
+          echo "-----------------------------"
+          if echo "$OUTPUT" | grep -q "Database schema is up to date"; then
+            echo "✅ Production database schema is up to date."
+            exit 0
+          fi
+          echo "::error::Prisma reports the production schema is NOT up to date — a migration may have failed to apply during release. Run 'npx prisma migrate status' / 'migrate deploy' against production and investigate before treating this deploy as healthy."
+          exit 1
+        env:
+          FLY_API_TOKEN: ${{ secrets.FLY_API_TOKEN }}
--- a/apps/webapp/app/modules/auth/mobile-sso.server.ts
+++ b/apps/webapp/app/modules/auth/mobile-sso.server.ts
@@ -35,10 +35,15 @@ import { mapAuthSession } from "./mappers.server";
 const label: ErrorLabel = "Auth";
 
 /**
- * Lifetime of a mobile auth code. Deliberately short — it only needs to survive
- * the deeplink round-trip from the web callback back into the app.
+ * Lifetime of a mobile auth code. Short by design — the code is minted in the
+ * web callback *after* SSO/MFA completes, so this window only spans the
+ * `shelf://` deeplink hand-back plus the app's exchange POST, not the
+ * human-paced IdP login. 180s (rather than a tighter 60s) leaves margin if iOS
+ * briefly backgrounds the app during the hand-back or the exchange runs on a
+ * slow network, without meaningfully widening the attack surface of a
+ * single-use, hashed code.
  */
-const MOBILE_AUTH_CODE_TTL_MS = 60_000;
+const MOBILE_AUTH_CODE_TTL_MS = 180_000;
 
 /**
  * SHA-256 hex digest. Auth codes are high-entropy (256-bit) single-use tokens,
--- a/apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
+++ b/apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
@@ -18,6 +18,7 @@ import {
   logException,
   parseData,
   payload,
+  readFormData,
 } from "~/utils/http.server";
 import { resolveUserAndOrgForSsoCallback } from "~/utils/sso.server";
 
@@ -99,6 +100,9 @@ export async function action({ request }: ActionFunctionArgs) {
 
     switch (method) {
       case "POST": {
+        // why: readFormData (not request.formData()) so a malformed body / wrong
+        // Content-Type is downgraded to a non-captured 400, rather than a
+        // TypeError that logException would otherwise surface as a captured 5xx.
         const {
           refreshToken,
           firstName,
@@ -110,7 +114,7 @@ export async function action({ request }: ActionFunctionArgs) {
           stateProvince,
           postalCode,
           country,
-        } = parseData(await request.formData(), MobileCallbackSchema);
+        } = parseData(await readFormData(request), MobileCallbackSchema);
 
         // Don't trust client tokens — re-derive the session from the refresh
         // token server-side (same trust boundary as the web callback).
--- a/apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
+++ b/apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
@@ -13,6 +13,7 @@ import { appendToMetaTitle } from "~/utils/append-to-meta-title";
 import { createSSOFormData } from "~/utils/auth";
 import { makeShelfError, notAllowedMethod, ShelfError } from "~/utils/error";
 import { getActionMethod, parseData, payload } from "~/utils/http.server";
+import { Logger } from "~/utils/logger";
 import { resolveUserAndOrgForSsoCallback } from "~/utils/sso.server";
 
 /**
@@ -138,6 +139,17 @@ export async function action({ request }: ActionFunctionArgs) {
     throw notAllowedMethod(method);
   } catch (cause) {
     const reason = makeShelfError(cause);
+    // why: the client component renders `result.error` and never re-throws, so
+    // without an explicit log a genuine 5xx (refresh-token exchange, user/org
+    // provisioning, or code mint failing) would never reach Sentry. Capture
+    // server faults as errors; keep a sampled, non-alerting trail of the
+    // expected 4xx (SSO disabled, malformed callback payload) for diagnostics.
+    // A missing status defaults to 500 (server fault).
+    if ((reason.status ?? 500) >= 500) {
+      Logger.error(reason);
+    } else {
+      Logger.handledClientError(reason);
+    }
     return data(
       { error: { message: reason.message } },
       { status: reason.status }
--- a/apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
+++ b/apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
@@ -13,8 +13,12 @@ import { appendToMetaTitle } from "~/utils/append-to-meta-title";
 import { createSSOFormData } from "~/utils/auth";
 import { mobilePkceChallengeCookie } from "~/utils/cookies.server";
 import { makeShelfError, notAllowedMethod, ShelfError } from "~/utils/error";
-import { getActionMethod, parseData, payload } from "~/utils/http.server";
-import { Logger } from "~/utils/logger";
+import {
+  getActionMethod,
+  logException,
+  parseData,
+  payload,
+} from "~/utils/http.server";
 import { resolveUserAndOrgForSsoCallback } from "~/utils/sso.server";
 
 /**
@@ -161,17 +165,11 @@ export async function action({ request }: ActionFunctionArgs) {
     throw notAllowedMethod(method);
   } catch (cause) {
     const reason = makeShelfError(cause);
-    // why: the client component renders `result.error` and never re-throws, so
-    // without an explicit log a genuine 5xx (refresh-token exchange, user/org
-    // provisioning, or code mint failing) would never reach Sentry. Capture
-    // server faults as errors; keep a sampled, non-alerting trail of the
-    // expected 4xx (SSO disabled, malformed callback payload) for diagnostics.
-    // A missing status defaults to 500 (server fault).
-    if ((reason.status ?? 500) >= 500) {
-      Logger.error(reason);
-    } else {
-      Logger.handledClientError(reason);
-    }
+    // why: the client renders `result.error` and never re-throws, so without an
+    // explicit log a genuine 5xx (refresh-token exchange, user/org provisioning,
+    // or code mint failing) would never reach Sentry. `logException` mirrors
+    // `error()`'s logging (5xx → Sentry, handled 4xx → trail, aborts skipped).
+    logException(reason);
     return data(
       { error: { message: reason.message } },
       {
--- a/apps/webapp/app/routes/api+/mobile+/exchange.ts
+++ b/apps/webapp/app/routes/api+/mobile+/exchange.ts
@@ -18,6 +18,7 @@ import {
 } from "~/modules/auth/mobile-sso.server";
 import { makeShelfError, notAllowedMethod, ShelfError } from "~/utils/error";
 import { getActionMethod } from "~/utils/http.server";
+import { Logger } from "~/utils/logger";
 
 const ExchangeSchema = z.object({
   code: z.string().min(1, "Authorization code is required"),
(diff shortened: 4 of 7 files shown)


carlosvirreira:
**Reviewer summary** (@DonKoko)

Additive prod hardening for the mobile SSO flow — three changes; success path and response shapes unchanged:

1. **Sentry capture for server faults** — `exchange.ts` and `oauth.callback_.mobile.tsx` caught errors and returned JSON without logging, so genuine 5xx never reached Sentry. They now `Logger.error` on `status >= 500` and keep a sampled, non-alerting trail of the expected 4xx.
2. **Post-deploy migrate-status gate** in `deploy.yml` — after the prod deploy, runs `prisma migrate status` on the running machine via `flyctl ssh` (reuses `FLY_API_TOKEN`, no DB credentials in CI) and fails the workflow if the schema is out of sync. It runs *after* deploy, so a failure is an alert to investigate — not a rollback.
3. **Single-use code TTL** widened `60s -> 180s`.

⚠️ One thing to eyeball on the first `main` deploy: the gate greps for `Database schema is up to date` — if `flyctl ssh` / `npx prisma` resolves differently on the machine, the command is a one-line tweak.

Independent of #2620 (PKCE) — either merge order. Part of the mobile SSO rollout (iOS 1.0.2 is in App Store review; Android SSO is tracked in #2611).

carlosvirreira:
@DonKoko — could you merge this one too when you get a sec?

Now that #2620 (PKCE) is merged and deploying, it adds a new `MobileAuthCode.codeChallenge` column. Having the **post-deploy migrate-status gate** from this PR live means any migration that doesn't apply on a deploy is caught at deploy time rather than surfacing later at runtime. This PR is additive and independent (the Sentry 5xx capture and the `60s -> 180s` code TTL come with it).

**What we'll do to test, check, and close out the mobile SSO work:**

1. Confirm the #2620 migration (`codeChallenge` column) applied to prod — a real SSO login exercises the new write/read path end-to-end.
2. Run the full PKCE SSO login from the companion app: `/sso-login` → IdP → `shelf://` deeplink → `POST /api/mobile/exchange` with the verifier → fresh session.
3. Monitor Sentry `label:Auth` on `/api/mobile/exchange` and `/oauth/callback/mobile` for any mint/exchange failures (the expected 4xx/429 are filtered out, so anything there is a real issue).
4. Once that's green, the server side of mobile SSO is complete; the companion-app PKCE half is tracked in #2611.

**Android / Play Store heads-up:** to ship the Android build we'll need a **Google Play service-account key added to `eas.json`** (currently missing) so `eas submit --platform android` works — flagging it now so it's ready when the Android binary is built.

Thanks!

apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
```diff
@@ -160,6 +161,17 @@ export async function action({ request }: ActionFunctionArgs) {
...
   } catch (cause) {
     const reason = makeShelfError(cause);
+    // why: the client component renders `result.error` and never re-throws, so
+    // without an explicit log a genuine 5xx (refresh-token exchange, user/org
+    // provisioning, or code mint failing) would never reach Sentry. Capture
+    // server faults as errors; keep a sampled, non-alerting trail of the
+    // expected 4xx (SSO disabled, malformed callback payload) for diagnostics.
+    // A missing status defaults to 500 (server fault).
+    if ((reason.status ?? 500) >= 500) {
+      Logger.error(reason);
```
DonKoko: 🔴 **Blocker: this exfiltrates the Supabase refresh token to Sentry.**

The reachable failure here is `refreshAccessToken(refreshToken)` (line 113), which throws:

```ts
// modules/auth/service.server.ts:537-547
throw new ShelfError({ cause, message: "...", label, additionalData: { refreshToken } });
```

No `status` → `ShelfError.status` defaults to 500 → `(reason.status ?? 500) >= 500` is true → `Logger.error(reason)` → `Sentry.captureException`. In `instrument.server.ts`, `makeSentryContext` does `extra: { ...(exception.additionalData || {}) }` and only the `__authSession` cookie is redacted — so the **raw refresh token is written to the Sentry event's `extra.refreshToken`** (and again inside `extra.cause.raw`).

New behavior: before this PR the catch only did `return data(...)`, so the token never left the server. An expired/invalid refresh token is a routine case, so this fires in normal use.

Fix before merge:
1. Remove `refreshToken` from `additionalData` in `refreshAccessToken` (`service.server.ts:545`).
2. Add a key denylist in `makeSentryContext` so `refreshToken`/`accessToken`/`password`/`codeVerifier`/etc. can never be spread into `extra` again.
3. Separately: an invalid refresh token should be `status: 401`, not a defaulted 500 — that also routes it to `handledClientError` (which does **not** serialize `additionalData`) instead of a paged error.

apps/webapp/app/routes/api+/mobile+/exchange.ts
```diff
@@ -77,6 +78,18 @@ export async function action({ request }: ActionFunctionArgs) {
...
   } catch (cause) {
     const reason = makeShelfError(cause);
+    // why: this resource route returns failures as JSON (the companion app
+    // parses `{ error }`) and never re-throws, so without an explicit log a
+    // genuine 5xx (Supabase mint outage, broken auth contract, DB/migration
+    // fault) would never reach Sentry — exactly how a prod migration-drift 500
+    // once went unnoticed. Capture server faults as errors; keep a sampled,
+    // non-alerting trail of the expected 4xx (expired / invalid / already-used
+    // code) for diagnostics. A missing status defaults to 500 (server fault).
+    if ((reason.status ?? 500) >= 500) {
```
DonKoko: 🟡 Non-blocking, but three things on this block (it's identical in `oauth.callback_.mobile.tsx`):

1. **Duplicated + reinvents an existing helper.** `error()` in `utils/http.server.ts` already does `Logger.error` + `Logger.handledClientError` (with an abort guard). You can't use it as-is here — it also fires `sendNotification` and returns a richer payload you intentionally avoid — but the logging half should be a shared `logException(reason)`, not copy-pasted.
2. **Abort noise.** `makeShelfError` maps client disconnects to status 499 (`< 500`) → this routes them to `handledClientError`, which emits a log-trail entry. The canonical `error()` deliberately skips `label === "Request aborted"`. Add the same guard.
3. **Scope.** All ~33 routes in `api+/mobile+/` swallow 5xx this same way with zero logging — this fixes 2. A shared helper makes finishing the job a one-liner per route.

(Exchange itself is safe re: the token leak — `redeemMobileAuthCode` doesn't put the code/verifier in `additionalData`.)

.github/workflows/deploy.yml
```diff
@@ -116,3 +116,33 @@ jobs:
...
+      # (e.g. a schema-path/env hiccup at release time), the deploy still goes
+      # green and a missing table only surfaces later as runtime 500s — which is
+      # how a prod migration once drifted unnoticed. This gate asks the running
+      # production machine whether Prisma considers the schema in sync and fails
+      # loudly otherwise, so drift is caught here rather than by a user. It
+      # reuses FLY_API_TOKEN (no DB credentials in CI) and runs on the machine,
+      # which already carries DATABASE_URL/DIRECT_URL as Fly secrets. It runs
+      # after deploy, so a failure is an alert to investigate — it does not roll
+      # the release back.
+      - name: 🔎 Verify migrations applied (production)
```
DonKoko: 🔴 Please remove this whole step.

- Out of scope for a mobile-SSO PR.
- Doesn't match how we actually run migrations (manual/controlled, not relied on via `release_command`). A post-deploy hard-fail when the schema isn't synced will false-alarm during the normal manual-migration window.
- Brittle: it greps Prisma's exact output string, and since `prisma` is a **devDependency** that `pnpm prune --prod` strips from the release image, `npx prisma migrate status` will pull the Prisma CLI from npm **onto the live prod machine** on every `main` deploy (npm-registry flake → false deploy failure).

Side note: `fly.toml` still has `release_command = "npx prisma migrate deploy"` (added in e9c59d60c). If migrations are manual now, that line is also misleading — worth reconciling, but separately from this PR.

apps/webapp/app/modules/auth/mobile-sso.server.ts
```diff
@@ -35,10 +35,15 @@ import { mapAuthSession } from "./mappers.server";
...
+ * Lifetime of a mobile auth code. Short by design — the code is minted in the
+ * web callback *after* SSO/MFA completes, so this window only spans the
+ * `shelf://` deeplink hand-back plus the app's exchange POST, not the
+ * human-paced IdP login. 180s (rather than a tighter 60s) leaves margin if iOS
+ * briefly backgrounds the app during the hand-back or the exchange runs on a
+ * slow network, without meaningfully widening the attack surface of a
+ * single-use, hashed code.
  */
-const MOBILE_AUTH_CODE_TTL_MS = 60_000;
+const MOBILE_AUTH_CODE_TTL_MS = 180_000;
```
DonKoko, line 46: 🟢 60→180s is fine — single-use, hashed, minted *after* SSO/MFA, and no test pins the value. The widened window only matters for legacy (non-PKCE) builds where an intercepted code could be replayed before the app redeems it; PKCE (#2620) is the real mitigation, so no concern.

DonKoko (changes requested):
## Review — mobile SSO prod hardening

Thanks @carlosvirreira. The Sentry-logging idea is sound in principle — I traced the chain and these routes catch-and-return (nothing else captures them), and `beforeSend` correctly drops 4xx / respects `shouldBeCaptured`. But two things need to change before this merges, plus some cleanups.

### 🔴 Blocker — this leaks Supabase refresh tokens to Sentry
The new `Logger.error(reason)` in `oauth.callback_.mobile.tsx` ships the user's **Supabase refresh token** to Sentry on the (routine) refresh-failure path. `refreshAccessToken()` throws `new ShelfError({ cause, additionalData: { refreshToken } })` with no `status` → defaults to 500 → `Logger.error` → `Sentry.captureException` → `makeSentryContext` spreads `additionalData` into `extra`, and `instrument.server.ts` only redacts the `__authSession` cookie, not `additionalData`. So `extra.refreshToken` lands in Sentry verbatim.

This is a **regression introduced by this PR**: before it, this catch never logged, so the token never left the server. A refresh token is a long-lived bearer credential — shipping it to a third-party error tracker (team-wide access + retention) is a real exposure. Details inline.

Minimum fix: stop putting the token in `additionalData` (`service.server.ts:545`), and add a redaction denylist in `makeSentryContext` so secret-ish keys (`refreshToken`, `accessToken`, `password`, `codeVerifier`, …) can never be spread into `extra`. Worth grepping every `additionalData:` for the same class.

### 🔴 Remove the migrate-status CI gate entirely
Out of scope for a mobile-SSO PR, and it doesn't match how we actually apply migrations (manual/controlled). It's also brittle — see inline. Please drop the whole step.

### 🟡 Cleanups (non-blocking)
- The 5xx/4xx logging branch is **duplicated verbatim** across both files and re-implements `error()` (`utils/http.server.ts`). Extract a tiny `logException(reason)` helper and reuse.
- The same root cause ("mobile resource routes swallow 5xx → invisible to Sentry") applies to **all ~33 routes in `api+/mobile+/`**, not just these two — none of the others log. Fine to scope, but a shared helper makes finishing the job trivial; please note the gap.
- TTL 60→180s is fine.

apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
```diff
@@ -138,6 +139,17 @@ export async function action({ request }: ActionFunctionArgs) {
...
```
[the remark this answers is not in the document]
carlosvirreira: Fixed in 80698656a — removed `refreshToken` from `additionalData` in `refreshAccessToken()`, and added a `redactSecrets` backstop in `makeSentryContext` that strips secret-keyed values (token/password/secret/verifier/cookie/authorization/api-key) before they are spread into the Sentry event `extra`. So the token can no longer reach Sentry from this path, or from any other flat `additionalData`.

apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
carlosvirreira: Fixed in 80698656a — removed `refreshToken` from `additionalData` in `refreshAccessToken()`, and added a `redactSecrets` backstop in `makeSentryContext` that strips secret-keyed values (token/password/secret/verifier/cookie/authorization/api-key) before they are spread into the Sentry event `extra`. So the token can no longer reach Sentry from this path, or from any other flat `additionalData`.

apps/webapp/app/routes/api+/mobile+/exchange.ts
carlosvirreira: Done in 80698656a — extracted a shared `logException(reason)` in `utils/http.server.ts` that mirrors `error()` logging, including the `label === "Request aborted"` (499) skip. Replaced the duplicated branch in both `exchange.ts` and `oauth.callback_.mobile.tsx`, and noted that all ~33 `api+/mobile+/` routes swallow 5xx the same way — `logException` is the shared piece to finish that off route by route.

.github/workflows/deploy.yml
carlosvirreira: Removed in 80698656a — dropped the whole step. Left `fly.toml`'s `release_command` untouched, to reconcile separately as you suggested.

apps/webapp/app/modules/auth/mobile-sso.server.ts
carlosvirreira: 👍 Kept at 180s.

carlosvirreira:
@DonKoko — addressed all the review feedback in 80698656a:

- 🔴 **Refresh-token leak:** removed `refreshToken` from `additionalData` in `refreshAccessToken()`, plus a `redactSecrets` backstop in `makeSentryContext` that strips secret-keyed values (token/password/secret/verifier/cookie/authorization/api-key) from the Sentry event `extra` app-wide — defense-in-depth for any stray `additionalData`.
- 🔴 **Migrate-status gate:** removed entirely. Left `fly.toml`'s `release_command` for the separate reconcile you mentioned.
- 🟡 **Shared helper:** extracted `logException()` (mirrors `error()`'s logging, including the `Request aborted`/499 skip) and used it in both routes; noted the ~33-route `api+/mobile+/` scope.
- 🟢 TTL kept at 180s.

Green locally: typecheck, 2335 unit tests, eslint, prettier, and the pre-commit security review (now "approve, low risk", flagging the token removal + redaction as net security improvements). All five threads replied + resolved — ready for another look when you have a sec.

apps/webapp/app/routes/_auth+/oauth.callback_.mobile.tsx
```diff
@@ -160,6 +165,11 @@ export async function action({ request }: ActionFunctionArgs) {
     throw notAllowedMethod(method);
   } catch (cause) {
     const reason = makeShelfError(cause);
+    // why: the client renders `result.error` and never re-throws, so without an
+    // explicit log a genuine 5xx (refresh-token exchange, user/org provisioning,
+    // or code mint failing) would never reach Sentry. `logException` mirrors
+    // `error()`'s logging (5xx → Sentry, handled 4xx → trail, aborts skipped).
+    logException(reason);
     return data(
```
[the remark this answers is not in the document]
carlosvirreira: Done in 5bae2ef53 — switched to `readFormData(request)`, so a malformed body / wrong `Content-Type` is downgraded to the intended non-captured 400 instead of a `TypeError` that `logException` would surface as a captured 5xx. Good catch.

apps/webapp/server/instrument.server.ts
```diff
@@ -153,6 +153,33 @@ function handleBeforeSendError<E extends Event>(event: E, hint: EventHint) {
...
+ */
+function redactSecrets(data: unknown): Record<string, unknown> {
+  if (!data || typeof data !== "object") {
+    return {};
+  }
+  const out: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(data as Record<string, unknown>)) {
+    out[key] = SENSITIVE_KEY_PATTERN.test(key) ? "[redacted]" : value;
+  }
+  return out;
```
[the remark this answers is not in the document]
carlosvirreira: Done in 5bae2ef53 — `redactSecrets` is now recursive and cycle-safe (WeakSet), so a secret nested under a non-sensitive key (e.g. `{ session: { refreshToken } }`) is scrubbed too, not just top-level keys. Also added `jwt`/`credential` to the sensitive-key pattern.

apps/webapp/app/utils/http.server.ts
```diff
@@ -296,6 +296,28 @@ export function payload<T extends ResponsePayload>(data: T) {
...
+ * which also fires a user notification and returns a richer payload). Mirrors
+ * `error()`'s logging exactly: 5xx (and uncaught) reach Sentry via
+ * `Logger.error`; handled 4xx land on the low-severity log trail via
+ * `Logger.handledClientError` (and are dropped from the Sentry error pipeline
+ * by `handleBeforeSendError`); client disconnects (status 499, "Request
+ * aborted") are intentionally skipped from both.
+ *
+ * @param cause - The normalized `ShelfError` to log
+ */
+export function logException(cause: ShelfError) {
```
DonKoko, line 313: 🟡 Suggestion: make `logException` the canonical logger and have `error()` **delegate** to it.

Right now the logging core is duplicated — `error()` (just below) repeats the exact `if (label !== "Request aborted") { Logger.error; Logger.handledClientError }` this function does. Two copies of the same behavior will drift over time.

To be clear, `logException` is the right thing to add: the mobile routes can't call `error()` directly because it also fires `sendNotification` and returns a richer `{ message, label, additionalData, traceId }` payload these session-less JSON endpoints intentionally don't expose. So keeping a logging-only primitive is correct — the duplication should just be collapsed the other way, with `error()` calling `logException`:

```ts
export function error(cause: ShelfError, shouldSendNotification = true) {
  logException(cause);

  if (
    cause.label !== "Request aborted" &&
    cause.additionalData?.userId &&
    typeof cause.additionalData?.userId === "string" &&
    shouldSendNotification
  ) {
    sendNotification({ /* ... */ });
  }

  return { error: { /* ... */ } };
}
```

Behavior-preserving (the inlined block is byte-for-byte what `logException` does), and leaves a single source of truth for "how we log a caught ShelfError." `error()` then reads as `logException()` + notification + payload.

apps/webapp/server/instrument.server.ts
```diff
@@ -176,7 +231,7 @@ function makeSentryContext(exception: unknown) {
       shelf_trace_id: exception.traceId || "Unknown",
     },
     extra: {
-      ...(exception.additionalData || {}),
+      ...redactSecrets(exception.additionalData),
```
DonKoko, line 234: 🟡 (CodeRabbit flagged this too as a nitpick — agreeing it's worth taking.) The redaction here only covers `additionalData`, but `cause.raw` a few lines down still serializes the raw cause chain unredacted:

```ts
cause: {
  message: (exception.cause as Error | null)?.message,
  raw: JSON.stringify(exception.cause), // ← bypasses redactValue
},
```

Nothing leaks through it *today* (the refresh token was removed from the error entirely, and the remaining `cause` is the Supabase error object), so this is defense-in-depth only. But the `redactSecrets` docstring claims it "guarantees a stray one (at any depth) can't reach Sentry" — which isn't true while `cause.raw` bypasses it. The one-liner closes the gap and makes the doc honest:

```diff
-        raw: JSON.stringify(exception.cause),
+        raw: JSON.stringify(redactValue(exception.cause, new WeakSet())),
```

DonKoko (commented):
Two non-blocking follow-ups after the latest round — the `logException` extraction, the `readFormData` switch, and the recursive `redactSecrets` all look good. Nice work removing the refresh-token leak and dropping the migrate-status gate.

apps/webapp/server/instrument.server.ts
carlosvirreira: Done in `9bf1e633b` — `cause.raw` now serializes `redactValue(exception.cause, new WeakSet())`, so the raw cause chain runs through the same recursive redaction as `additionalData` instead of bypassing it. The `redactSecrets` "no secret at any depth" guarantee now holds end-to-end. Defense-in-depth as you noted — nothing leaks through it today.

apps/webapp/app/utils/http.server.ts
carlosvirreira: Done in `9bf1e633b` — `error()` now delegates to `logException(cause)` and the duplicated abort-guard + `Logger.error` / `Logger.handledClientError` block is gone, so there is a single source of truth for logging a caught ShelfError. Behavior-preserving (the inlined block was byte-for-byte `logException`); `error()` now reads as `logException()` + notification + payload, as suggested.

carlosvirreira:
@DonKoko both follow-ups from your last review are addressed in `9bf1e633b` (pushed) — both threads replied + resolved, **0 open threads** now:

- **`cause.raw` redaction** (`instrument.server.ts`) — `cause.raw` now serializes `redactValue(exception.cause, new WeakSet())` so the raw cause chain runs through the same recursive redaction as `additionalData` instead of bypassing it. The `redactSecrets` "no secret at any depth" guarantee now actually holds end-to-end. Defense-in-depth as you noted — nothing leaks through it today.
- **`logException` as the canonical logger** (`http.server.ts`) — `error()` now delegates to `logException(cause)` and the duplicated abort-guard + `Logger.error` / `Logger.handledClientError` block is gone. Single source of truth; behavior-preserving (the inlined block was byte-for-byte `logException`). `error()` now reads as `logException()` + notification + payload, exactly as you suggested.

Both non-blocking and behavior-preserving — typecheck + lint green. Ready for another look whenever.

DonKoko (approved):

PR #2621, opened 2026-06-11
Security: Unsafe redirect constructed from request URL in splat route

qdzsh:
## Summary

Security: Unsafe redirect constructed from request URL in splat route

## Problem

**Severity**: `Medium` | **File**: `apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx:L10`

The loader splits `request.url` on a user-controlled `selectedUserId` and redirects to the trailing portion without validating that the result is a safe local path. A crafted URL could cause an open redirect or redirect to an unexpected route if the extracted suffix begins with `//` or contains a scheme-like value.

## Solution

Parse the URL and only redirect to a validated relative pathname/search/hash component. Reject any target that is not a local path, and avoid deriving redirect targets by string splitting on untrusted input.

## Changes

- `apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened redirect validation to block unsafe, malformed, and protocol-relative targets so users aren’t redirected off-site.
* **Refactor**
  * Unified redirect handling to preserve query and hash fragments and enforce same-origin rules for reconstructed wildcard paths.
* **Tests**
  * Added open-redirect edge-case tests covering backslash, encoded, protocol-relative, and internal-route attempts.
* **Documentation**
  * Expanded guidance on wildcard path reconstruction and safe redirect behavior.

### DIFF
--- a/apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx
+++ b/apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx
@@ -7,13 +7,16 @@ import { getParams } from "~/utils/http.server";
  * This is mostly made for handling users cmd+click on asset or booking in user page
  */
 export const loader = ({ request, params }: LoaderFunctionArgs) => {
-  const { userId: selectedUserId } = getParams(
+  const { "*": splat } = getParams(
     params,
-    z.object({ userId: z.string() })
+    z.object({ userId: z.string(), "*": z.string() })
   );
+  const url = new URL(request.url);
+  const redirectTo = `/${splat}${url.search}${url.hash}`;
 
-  /** Split the url based on the user id */
-  const extraParts = request.url.split(selectedUserId);
+  if (!redirectTo.startsWith("/") || redirectTo.startsWith("//")) {
+    throw new Response("Invalid redirect", { status: 400 });
+  }
 
-  return redirect(extraParts[extraParts.length - 1]);
+  return redirect(redirectTo);
 };
--- a/apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx
+++ b/apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx
@@ -1,19 +1,19 @@
 import { redirect, type LoaderFunctionArgs } from "react-router";
 import { z } from "zod";
-import { getParams } from "~/utils/http.server";
+import { getParams, safeRedirect } from "~/utils/http.server";
 
 /**
  * Splat route to handle requests with data after userId in the url.
  * This is mostly made for handling users cmd+click on asset or booking in user page.
  *
- * Extracts the splat segment from the request URL and redirects to the
- * reconstructed local path, preserving the query string and hash. Only
- * same-origin relative paths are allowed; protocol-relative (`//`) or otherwise
- * non-local targets are rejected to prevent open redirects.
+ * Reconstructs the target from the splat segment (preserving query string and
+ * hash) and routes it through `safeRedirect`, the canonical guard that resolves
+ * the destination and falls back to "/" for anything that is not same-origin —
+ * closing open-redirect vectors like `//host`, `/\host` and `%5C`-decoded
+ * backslashes.
  *
  * @param args - Loader arguments containing the incoming `request` and route `params`.
- * @returns A redirect `Response` pointing to the validated relative path.
- * @throws {Response} 400 response when the resolved redirect target is not a safe local path.
+ * @returns A redirect `Response` to the validated same-origin path.
  */
 export const loader = ({ request, params }: LoaderFunctionArgs) => {
   const { "*": splat } = getParams(
@@ -21,11 +21,6 @@ export const loader = ({ request, params }: LoaderFunctionArgs) => {
     z.object({ userId: z.string(), "*": z.string() })
   );
   const url = new URL(request.url);
-  const redirectTo = `/${splat}${url.search}${url.hash}`;
 
-  if (!redirectTo.startsWith("/") || redirectTo.startsWith("//")) {
-    throw new Response("Invalid redirect", { status: 400 });
-  }
-
-  return redirect(redirectTo);
+  return redirect(safeRedirect(`/${splat}${url.search}${url.hash}`));
 };
--- a/apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx
+++ b/apps/webapp/app/routes/_layout+/settings.team.users.$userId.$.tsx
@@ -3,8 +3,17 @@ import { z } from "zod";
 import { getParams } from "~/utils/http.server";
 
 /**
- * Splat route to handle requests with data after userId in the url
- * This is mostly made for handling users cmd+click on asset or booking in user page
+ * Splat route to handle requests with data after userId in the url.
+ * This is mostly made for handling users cmd+click on asset or booking in user page.
+ *
+ * Extracts the splat segment from the request URL and redirects to the
+ * reconstructed local path, preserving the query string and hash. Only
+ * same-origin relative paths are allowed; protocol-relative (`//`) or otherwise
+ * non-local targets are rejected to prevent open redirects.
+ *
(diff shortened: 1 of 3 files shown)


DonKoko (commented):
### The structural change is right, but the guard is still bypassable — and we already have a helper for this

Two things:

**1. This route shouldn't hand-roll its own redirect check — we already have `safeRedirect()`** (`apps/webapp/app/utils/http.server.ts:234`). It's the canonical guard used in ~11 places (login, `oauth.callback`, `accept-invite`, `change-current-organization`, the `*.edit` routes, etc.), and our convention is to route user-supplied redirects through it rather than re-implementing the check inline.

**2. The bigger issue: the `startsWith("//")` check (here *and* in `safeRedirect`) doesn't actually close the open redirect.** It only catches protocol-relative `//host`. Browsers treat `\` like `/` in http(s) URLs, so a backslash slips straight through:

```
splat = "\evil.com"   →  redirectTo = "/\evil.com"
"/\evil.com".startsWith("//")  →  false   (passes the guard)
new URL("/\evil.com", origin).origin  →  "https://evil.com"   (browser redirects off-site)
```

And `\` is reachable as `%5C` in the path, which React Router decodes into the splat param — so this is exploitable, not theoretical. Both `\evil.com` and `\\evil.com` get through today.

### Suggested direction

Harden `safeRedirect` itself (origin-resolution instead of prefix-matching) and route this loader through it. That fixes this route **and** the other ~11 call sites that share the same gap:

```ts
export function safeRedirect(to, defaultRedirect = "/") {
  const safeList = [SERVER_URL, `https://${URL_SHORTENER}`];
  if (!to || typeof to !== "string") return defaultRedirect;

  // Allowed absolute domains (e.g. URL shortener) pass through unchanged
  if (safeList.some((u) => to.startsWith(u))) return to;

  // Otherwise must be a local absolute path
  if (!to.startsWith("/") || to.startsWith("/__")) return defaultRedirect;

  // Final defense: resolve and confirm it stays same-origin.
  // Catches "//host", "/\host", "/\\host" and %5C-decoded backslashes,
  // which prefix checks miss.
  try {
    const resolved = new URL(to, SERVER_URL);
    if (resolved.origin !== new URL(SERVER_URL).origin) return defaultRedirect;
  } catch {
    return defaultRedirect;
  }
  return to;
}
```

Then the loader collapses to:

```ts
const url = new URL(request.url);
return redirect(safeRedirect(`/${splat}${url.search}${url.hash}`));
```

I audited all current `safeRedirect` call sites: every one passes a relative/literal path and feeds the result straight into `redirect(...)`, and nothing relies on the `URL_SHORTENER` absolute branch — so this hardening is behavior-preserving for legit inputs. The existing tests (`"my-url"` → `/`, `"/items"` → `/items`) still hold; please also add cases for `//evil.com`, `/\evil.com`, and `%5Cevil.com` so the bypass can't regress.

The move away from `request.url.split(selectedUserId)` is a genuine improvement either way — reconstructing from the splat param is much less fragile than string-splitting on the id. 👍

qdzsh:
Thanks for the thorough review, both points are spot on, especially the backslash bypass. Done:

1. **Routed through `safeRedirect()`**: dropped the hand-rolled check in the loader; it's now just `return redirect(safeRedirect(`/${splat}${url.search}${url.hash}`))`.

2. **Hardened `safeRedirect()` itself**: replaced the `startsWith("//")` prefix check with origin resolution (`new URL(to, SERVER_URL).origin !== new URL(SERVER_URL).origin`). That closes `//host`, `/\host`, `/\\host` and the `%5C`-decoded backslash, all of which resolved off-origin under the old check. Allowed absolute domains still pass through `safeList` and same-origin paths return unchanged, so it's behavior-preserving for the ~14 existing call sites.

Added regression tests for `//evil.com`, `/\evil.com`, `/\\evil.com`, the `%5C`-decoded form, and `/__manifest`, alongside the existing `my-url` and `/items` cases.

Thanks again for pointing at the canonical helper rather than letting this route reinvent it.

DonKoko (commented):
### Follow-up review — core fix is 👍, one same-class hole remains in the same function

The main hardening is exactly right: routing the loader through `safeRedirect`, resolving the target by origin, and the `//host` / `/\host` / `%5C` bypass tests. That closes the original finding cleanly.

**But there's one remaining open redirect in `safeRedirect` itself** — the absolute-domain allow-list still matches by prefix:

```ts
const safeList = [SERVER_URL, `https://${URL_SHORTENER}`];
if (safeList.some((safeUrl) => to.startsWith(safeUrl))) {
  return to;
}
```

`SERVER_URL` has its trailing slash stripped (`env.ts:144` → `https://app.shelf.nu`), so `startsWith` matches any host that merely *begins* with it:

```
"https://app.shelf.nu.evil.com/phish"  startsWith "https://app.shelf.nu"  → true → returned as-is
"https://eam.sh.evil.com/phish"        → same, via URL_SHORTENER
"https://app.shelf.nu@evil.com/phish"  → userinfo trick, host is evil.com
```

All three resolve off-origin in the browser. And this branch **is reachable** — `login`, `oauth.callback`, `accept-invite`, and `change-current-organization` all pass a user-supplied `redirectTo` (query/form) straight into `safeRedirect`, e.g. `/login?redirectTo=https://app.shelf.nu.evil.com`. It's pre-existing, but it's the same bug class this PR is hardening, in the function being hardened — worth closing here while we're in it.

The fix is to validate the absolute branch by **origin** too, not prefix. This preserves the existing `"my-url" → "/"` contract (relative inputs throw in `new URL(to)` and fall through):

```ts
export function safeRedirect(to, defaultRedirect = "/") {
  if (!to || typeof to !== "string") return defaultRedirect;
  if (to.startsWith("/__")) return defaultRedirect;

  // Absolute URL? Validate by ORIGIN, never by prefix
  // ("https://app.shelf.nu.evil.com" begins with SERVER_URL but is off-origin).
  try {
    const abs = new URL(to); // throws for relative inputs → falls through
    const allowed = new Set([new URL(SERVER_URL).origin]);
    if (URL_SHORTENER) allowed.add(new URL(`https://${URL_SHORTENER}`).origin);
    return allowed.has(abs.origin) ? to : defaultRedirect;
  } catch {
    // not absolute — fall through to relative-path handling
  }

  if (!to.startsWith("/")) return defaultRedirect;
  try {
    if (new URL(to, SERVER_URL).origin !== new URL(SERVER_URL).origin) {
      return defaultRedirect;
    }
  } catch {
    return defaultRedirect;
  }
  return to;
}
```

I ran this against the full set — existing contract (`my-url`/`//`/`//my-url`/`/items`), your new bypass cases, and the three prefix payloads above — all pass. Worth adding these to the new test block so the prefix bypass can't regress either:

```ts
expect(safeRedirect("https://app.shelf.nu.evil.com/phish")).toBe("/");
expect(safeRedirect("https://app.shelf.nu@evil.com/phish")).toBe("/");
```

Nice turnaround on the rest of it.

qdzsh:
Good catch, you're right, the absolute allow-list was still prefix-matching, which is the same bug class in the same function. Closed it: the absolute branch now parses with `new URL(to)` and compares against an origin allow-set (`SERVER_URL` plus optional `URL_SHORTENER`) instead of `startsWith`. Relative inputs throw there and fall through to the same-origin path check, so the `"my-url"` to `"/"` contract is unchanged.

Added the two regression tests you suggested (`app.shelf.nu.evil.com` and the `@evil.com` userinfo trick), and re-verified the whole set (existing contract, the `//` / `/\` / `%5C` cases, and the prefix/userinfo payloads) against both a localhost and a production-style `SERVER_URL`. All pass. Thanks for the careful second pass.

DonKoko (approved):
LGTM ✅ Origin-based validation closes the prefix/userinfo bypass too, and dropping the `https://${URL_SHORTENER}` prefix entry fixes the latent "unset shortener matches all https" case. Tests cover the bypasses and pass locally (44/44). Nice work.

PR #2634, opened 2026-06-17
fix(companion): align expo-image-manipulator to SDK 54 (14.0.8) to fix iOS build

carlosvirreira:
## What

`expo-image-manipulator@55.0.17` breaks the **iOS build** — its iOS Swift calls `FileSystemUtilities.isReadableFile`, which doesn't exist in the `expo-modules-core@3.0.29` that Expo SDK 54 ships:

> type 'FileSystemUtilities' has no member 'isReadableFile'

`14.0.8` is the version SDK 54 bundles (`expo install --check` recommends it). Its iOS code uses `FileSystemUtilities.permissions` + Foundation's `FileManager.isReadableFile` — all present in core 3.0.29.

## Correction of the version call from #2628

On #2628 I argued `~55.0.17` was intentional and that `expo install` would reintroduce the Android crash. That was wrong on the iOS axis — `55.x` *runs* on Android but its iOS Swift expects a newer core, so it fails to compile for iOS. `14.0.8` is correct for **both** platforms. CodeRabbit's `expo install` suggestion was right.

## Verification

- **iOS:** EAS iOS-simulator build compiles cleanly on `14.0.8` (the prior `55.0.17` build errored at the exact `isReadableFile` symbol).
- **JS:** the only consumer — `apps/companion/components/audit/evidence-modal.tsx` (HEIC→JPEG via `manipulateAsync` + `SaveFormat.JPEG`) — has an identical API/signature in `14.0.8`. No code change.
- **Android:** `14.0.8` has no `AnyTypeCache` reference (the v56-class regression), so the **already-shipped Android build is unaffected**.

## Scope

Targeted single-package realignment. `expo-image@55.0.6` is intentionally **left as-is** — it's kept iOS-compile-safe by its committed patch (`patches/expo-image@55.0.6.patch`). Don't touch it.

This unblocks the next iOS release (so iOS can ship the SSO update Android just got).

## Summary by CodeRabbit

* **Chores**
  * Updated `expo-image-manipulator` dependency version in the companion app.


carlosvirreira:
**Live iOS SSO pass ✅** — covering the #4 ask from #2628 ("run one live iOS SSO sign-in and note the result").

Built this branch for the iOS Simulator (prod env), installed + launched it:
- App **launches and renders** correctly on `expo-image-manipulator@14.0.8` — confirms the fix at runtime, not just at compile time.
- Tapped **"Sign in with SSO"** → Microsoft sign-in → **landed logged-in on the Assets screen**. iOS SSO works end-to-end (in-frame `ASWebAuthenticationSession` exchange; iOS never hits the `auth-callback` route).

So this PR fixes the iOS build **and** is verified to run + sign in via SSO on iOS. Android (already live) is unaffected.

PR #2679, opened 2026-06-30
fix(sentry): kit location 23514 + /kits and asset overview TypeErrors

DonKoko:
## Summary

Triage of three Sentry issues from the 2026-06-30 sweep on shelf-webapp.

- **SHELF-WEBAPP-1P4** — `updateKitLocation` raised Postgres 23514 when an INDIVIDUAL asset in the kit already held a manual `AssetLocation` row (`assetKitId IS NULL`). The `enforce_individual_asset_single_location` trigger rejects the kit-driven INSERT because INDIVIDUAL assets must hold at most one `AssetLocation` row. Fix: query the manual rows for the kit's INDIVIDUAL assets and exclude them from the `createMany` payload — manual placements override kit-driven cascades. Same correction applied to `bulkUpdateKitLocation` (verified to have the identical antipattern). Regression tests added for both.
- **SHELF-WEBAPP-1P2** — TypeError on `/kits` (`Cannot read properties of undefined (reading 'some')`) in a `useMemo` → `Array.map` → `.some(...)` chain. `assetKits` is typed as always-present by the loader, but at least one production crash shows a runtime-undefined edge case. Defensive `(?? [])` guards on three call sites. The deeper *why is `assetKits` undefined?* is a follow-up.
- **SHELF-WEBAPP-1P3** — TypeError on `/assets/:id/overview` (`undefined is not an object (evaluating 's.name')`). `ua-parser-js` can return empty sub-objects for malformed UA strings — `scan-details` now optional-chains `ua.browser?.name` and `ua.os?.name` with an `"Unknown"` fallback. Also makes `resolveTeamMemberName` null-safe (early return `""` for undefined input) as defense-in-depth.

Two paired transient-DB issues from the same sweep (1P5 + 1P6) were resolved directly in Sentry as a one-second Supabase pooler blip — no code change needed.

## Test plan

- [x] \`pnpm webapp:validate\` green (2953 tests + 1 skipped; lint, prettier, typecheck, security-review all pass)
- [x] New regression tests cover both \`updateKitLocation\` and \`bulkUpdateKitLocation\` manual-placement paths
- [ ] Manual SQL probe on staging to find a kit with the bug class, then UI-move it
- [ ] Confirm Sentry quiet on \`firstSeen:-24h\` for 1P4 / 1P2 / 1P3 post-deploy

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability in kit availability/booking calculations when asset-kit data is missing, including kit status display.
  * Made scan details more defensive by showing “Unknown” for missing or malformed browser/OS data.
  * Preserved user-manual placements for individual assets during kit location updates; stopped emitting activities and system notes for intentionally skipped assets.
  * Made team member name rendering null-safe when member data is missing.
* **Tests**
  * Added/expanded regression coverage for kit location cascade behavior and the manual-placement skip guard.

### DIFF
--- a/apps/webapp/app/components/assets/assets-index/use-kit-availability-data.ts
+++ b/apps/webapp/app/components/assets/assets-index/use-kit-availability-data.ts
@@ -33,12 +33,14 @@ export function useKitAvailabilityData(items: Items) {
         mainImage: item.image,
         thumbnailImage: item.imageExpiration,
         status: item.status,
-        // why: assetKits is typed as always-present from the loader, but at least
-        // one production crash shows a runtime-undefined edge case. Empty array is
-        // a safe stand-in — an empty kit has nothing available to book.
-        availableToBook: (item.assetKits ?? []).some(
-          (ak) => ak.asset.availableToBook
-        ),
+        // why: match the list view's semantic in kits._index.tsx — a kit is
+        // bookable only when ALL slices are bookable (booking reserves the
+        // whole kit). Undefined assetKits is treated as not-bookable since we
+        // can't verify the slices.
+        availableToBook:
+          item.assetKits == null
+            ? false
+            : !item.assetKits.some((ak) => !ak.asset.availableToBook),
       },
     }));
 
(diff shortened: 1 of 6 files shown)


DonKoko:
@coderabbit make another review, considering all the changes we have done in the commits since your prev review.

PR #2718, opened 2026-07-21
fix(bookings): guard mobile fulfil-and-checkout against cross-user checkout

carlosvirreira:
## What

`POST /api/mobile/bookings/fulfil-and-checkout` (added in #2710) had no booking-ownership check.

`SELF_SERVICE` and `BASE` both hold `booking:checkout` in the permission map, so `requireMobilePermission` passes for **any** booking id a restricted user sends. And unlike the scan-add path — whose cross-user guard lives inside `processBooking` — `fulfilModelRequestsAndCheckout` does not check ownership itself. Web's `fulfil-and-checkout` loader blocks this with `canUserManageBookingAssets`, so the mobile route was strictly **more permissive than web**: a restricted user could fulfil the model reservations of, and check out, another user's booking.

Fix: apply the shared `validateBookingOwnership` guard (creator-or-custodian for SELF_SERVICE/BASE, no-op for ADMIN/OWNER) immediately after the org-scoped booking lookup — the same pattern the sibling mobile `cancel` / `archive` / `delete` / `duplicate` routes already use.

## Exposure

Introduced by #2710 (merged 2026-07-13). **Not in any store build** — the newest companion build (29) predates that merge — but the endpoint is live wherever `main` is deployed, so it's reachable by any authenticated mobile client.

## Testing

`booking-authorization.server.test.ts` (8 tests) plus the booking module + mobile API route suites: **481 tests passing**. Typecheck and prettier clean. No route-level test added: the four sibling guarded mobile routes have none either, the helper is unit-tested centrally.

## Related, not fixed here

Several other mobile booking routes lack the route-level guard (`checkout`, `checkin`, `partial-checkout`, `partial-checkin`, `reserve`, `update`, `remove-assets`, `model-requests`). Some are covered at the service layer (`add-scanned-assets` is, via `processBooking`), so each needs individual verification rather than a blanket patch. Worth a follow-up sweep — happy to do it if you want.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an explicit ownership and permission check for mobile booking checkout.
  * Ensures callers can only “check out” bookings they are authorized to access, preventing unauthorized cross-account actions.

### DIFF
--- a/apps/webapp/app/routes/api+/mobile+/bookings.fulfil-and-checkout.ts
+++ b/apps/webapp/app/routes/api+/mobile+/bookings.fulfil-and-checkout.ts
@@ -6,8 +6,10 @@ import {
   requireMobilePermission,
   requireOrganizationAccess,
   assertMobileCanUseBookings,
+  getMobileUserContext,
 } from "~/modules/api/mobile-auth.server";
 import { fulfilModelRequestsAndCheckout } from "~/modules/booking/service.server";
+import { validateBookingOwnership } from "~/utils/booking-authorization.server";
 import { getClientHint, type ClientHint } from "~/utils/client-hints";
 import { makeShelfError } from "~/utils/error";
 import {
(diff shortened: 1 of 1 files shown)


carlosvirreira:
@coderabbit full review

DonKoko (approved):

PR #2737, opened 2026-07-23
feat(custom-fields): add sorting and grouping for asset custom fields

### LINKED ISSUE #2606: [Feature request]: Sorting/grouping for asset custom fields
### Contact Details

_No response_

We have added the following custom fields to our workspace, but they are showing in alphabetical order, not the order we added them.  This is causes confusion as they are not logically sorted/grouped for users.

Image

I'd like to group the fields and sort them within the groups.  As an example, I would have the following groups and sorting for our fields:

Item details
- Brand
- Model
- Serial number

Procurement
- Purchase date
- Supplier
- Purchase amount
- PO number

Warranty & protection
- Warranty months
- Insurance cover

(The groups above happen to be in the correct order alphabetically, but the groups should also be sortable to give the most logical order.)

We could add numbers at the beginning of each custom field name for sorting, but this still wouldn't give logical groupings.  It would also make importing more difficult as you'd need to include those in the CSV file headings.

### Additional context

_No response_

Pallavikumarimdb:
## Description
Implements the ability to group and sort custom fields on asset forms. This introduces a new `CustomFieldGroup` model, settings UI for managing groups, and updates the asset creation/edit forms to render custom fields grouped under their respective headers and sorted by position.

## changes
- **Database Schema**:- Added `CustomFieldGroup` model and associated it with `CustomField`.
- **CRUD Operations**:- Added backend service methods to support Custom Field Group creation, updating, deletion, and positioning.
- **Settings UI**:- Created routes and components under `/settings/custom-fields/groups` to CRUD groups and reorder them via drag-and-drop.
- **Asset Forms UI**:- Updated `custom-fields-inputs.tsx` to group and sort active custom fields by their assigned groups and specified positions before rendering.

Closes: #2606 

## Summary by CodeRabbit

* **New Features**
  * Added custom field groups with group and position controls when creating or editing custom fields.
  * Added a “Manage groups” page to create, delete, and reorder groups.
  * Custom fields now display in grouped sections, including “Other Fields” for ungrouped fields, following configured ordering.
* **Bug Fixes**
  * Improved custom field and group ordering consistency.
* **Tests**
  * Updated custom field test fixtures with grouping and position defaults.


DonKoko (changes requested):
Thanks for this, @Pallavikumarimdb — and thanks for sticking with it through the earlier bot rounds. Grouping custom fields is something people have asked for, and the shape of the solution (a first-class `CustomFieldGroup` with its own settings page, plus ordering) is the right one. Nice touch on the delete dialog explaining that fields survive the group being removed.

I've done a full pass. There's a fair bit below, but most of it collapses into one schema decision, so please read the first section before working through the rest.

---

## 🖼️ First: could you add screenshots to the PR description?

This PR changes real UI in three places and there's nothing visual to review against. Could you edit the original PR description (top post) and add screenshots of:

1. **`/settings/custom-fields/groups`** — the new "Manage Custom Field Groups" page, both empty state and with a few groups in it
2. **The custom field create/edit form** — showing the new "Group" select and "Position" input
3. **The asset create/edit form** — custom fields rendered under group headings, ideally next to a "before" shot of the current Required/Optional layout

A short screen recording of creating a group → assigning a field → seeing it grouped on the asset form would be even better, since reordering is interactive.

Putting them in the description rather than a comment keeps them visible for whoever reviews next and for the release notes later. 🙏

---

## 🔴 Blockers

### 1. The composite `[groupId, organizationId]` foreign key — please replace with a plain relation

This is the one to fix first, because three of the four blockers are downstream of it.

```prisma
group CustomFieldGroup? @relation(fields: [groupId, organizationId], references: [id, organizationId], onDelete: NoAction)
```

There's no precedent for this in the schema — `references: [id, organizationId]` appears zero times across 58+ models. It's what forces the `onDelete: NoAction`, the redundant `@@unique([id, organizationId])`, the unusual `connect: { id_organizationId: {…} }` call shape, and the manual "null out the children, then delete the parent" dance in `deleteCustomFieldGroup`.

And it isn't buying protection we don't already have: `createCustomField` and `updateCustomField` both validate that the group belongs to the caller's org before writing, which is exactly what [`.claude/rules/org-scope-user-supplied-ids.md`](.claude/rules/org-scope-user-supplied-ids.md) prescribes. The DB-level composite key is belt-and-braces that costs us a lot of complexity.

Please use the ordinary form instead:

```prisma
groupId String?
group   CustomFieldGroup? @relation(fields: [groupId], references: [id], onDelete: SetNull)
```

Keep the app-level org assertion. This single change resolves blockers 2 and 3 below, and the `disconnect` question in §7.

### 2. The migration will fail on `prisma migrate deploy`

`20260722175922_add_custom_field_group/migration.sql` contains a bare `COMMIT;`, then `CREATE INDEX CONCURRENTLY`, then `BEGIN; … COMMIT;`.

Prisma runs each migration inside a transaction, and `CREATE INDEX CONCURRENTLY` cannot run inside one. We document this verbatim in our own most recent index migration:

> *"Not CONCURRENTLY: Prisma runs each migration in a transaction, and CREATE INDEX CONCURRENTLY cannot run inside one."*
> — `packages/database/prisma/migrations/20260814171814_add_asset_custom_field_value_searchable_trgm_idx/migration.sql`

No migration in this repo issues `CONCURRENTLY`. Worse, because the `COMMIT;` on line 33 has already committed the `ALTER TABLE` / `CREATE TABLE` / FK statements, a failure at the `CREATE INDEX` leaves the database half-migrated with the migration marked failed — every subsequent deploy then aborts until someone manually runs `prisma migrate resolve`.

The `NOT VALID` + `VALIDATE CONSTRAINT` split is also unnecessary here: `groupId` is a brand-new, all-`NULL` column, so validation is trivially satisfied, and `CustomField` holds field *definitions* (tens of rows per org) — nowhere near the scale that warrants online-DDL mitigations.

Please regenerate this as a plain migration: `ALTER TABLE`, `CREATE TABLE`, a normal `CREATE INDEX`, and a normal FK. No `COMMIT`, no `BEGIN`, no `CONCURRENTLY`.

### 3. `deleteCustomFieldGroup` is not transactional

```ts
await db.customField.updateMany({ where: { groupId: id, organizationId }, data: { groupId: null } });
return await db.customFieldGroup.delete({ where: { id, organizationId } });
```

Two separate statements. If the `delete` fails — concurrent delete (`P2025`), dropped connection, FK check — every custom field in that group has already permanently lost its assignment while the group still exists. The user sees an error, retries, and there's no record of which fields belonged where.

With `onDelete: SetNull` from §1 this function becomes a single `delete` and the problem disappears. If you'd rather keep the manual version, wrap both calls in `db.$transaction`.

### 4. The delete-group dialog never closes after a successful delete

```ts
if (fetcher.state === "idle" && fetcher.data && !("error" in fetcher.data)) {
  setOpen(false);
}
```

`payload()` returns `{ error: null, ...data }` ([`app/utils/http.server.ts:298`](apps/webapp/app/utils/http.server.ts#L298)), so the `error` key is *always* present — the condition is permanently false and `setOpen(false)` never runs. After a successful delete the modal stays open over a list that has already lost the group, inviting a second delete that 404s.

Two things compound it: the dialog's buttons take `disabled` from the parent's `useDisabled()` and reorder-fetcher state, neither of which changes during this dialog's own fetcher submission — so the Delete button never disables (double-click → two POSTs) and the `{disabled ? "Deleting..." : "Delete"}` label is stuck on "Delete".

```ts
// in DeleteCustomFieldGroupDialog
const disabled = useDisabled(fetcher);
useEffect(() => {
  if (fetcher.state === "idle" && fetcher.data && !fetcher.data.error) setOpen(false);
}, [fetcher.state, fetcher.data]);
```

---

## 🟠 Should fix before merge

### 5. The cross-org group guard's error message is swallowed

In both `createCustomField` and `updateCustomField` the new guard throws inside the `try`, and the `catch` re-wraps everything with `maybeUniqueConstraintViolation` — which unconditionally builds a **new** `ShelfError` with the generic *"We could not create or update this Custom field. Please try again or contact support."* whenever the cause isn't `P2002` ([`app/utils/error.ts:792`](apps/webapp/app/utils/error.ts#L792)). It also carries no status, so it surfaces as a 500.

Net effect: your carefully-worded message never reaches the user, and they get a "please try again" on a request that can never succeed. We have a helper for exactly this — call it first in the catch:

```ts
} catch (cause) {
  rethrowIfClientError(cause); // app/utils/error.ts:220
  throw maybeUniqueConstraintViolation(cause, "Custom field", { … });
}
```

Two smaller things on the same guard: `status: 403` should be `400` (or `404`) for an unknown/foreign id, and it needs `shouldBeCaptured: false` so it doesn't page us for user input.

### 6. `position` can never produce a validation error

```ts
position: z.string().optional().transform((val) => {
  if (!val) return 0;
  const parsed = parseInt(val, 10);
  return Number.isNaN(parsed) ? 0 : parsed;
})
```

This never calls `ctx.addIssue`, so `zo.errors.position()` and `validationErrors?.position` are permanently `undefined` — the `error` slot you render on that Input is dead. A user who clears the field or types `-3` gets `0` / `-3` written with no feedback, silently reordering their form. That's contrary to the Form Validation Pattern in `CLAUDE.md`.

```ts
position: z.coerce.number().int().min(0, "Position must be 0 or greater").optional().default(0)
```

…and add `min={0}` to the input.

### 7. Clearing a field's group uses `disconnect: true`

```ts
group: groupId === undefined ? undefined
  : groupId ? { connect: { … } }
  : { disconnect: true },
```

On the current composite relation, `disconnect` nulls the referencing scalars — including `organizationId`, which is `NOT NULL`. Please verify this path against a real database (edit a field that has a group → pick "None (No Group)" → Save); I suspect it throws.

If you take §1 this becomes safe as written. Either way, the unchecked-update form is unambiguous and I'd prefer it:

```ts
groupId: groupId === undefined ? undefined : (groupId ?? null),
```

### 8. `getActiveCustomFields` has a much wider blast radius than the asset form

Adding `include: { group: true }` and a six-key relational `orderBy` changes every caller, not just the asset form:

- `app/utils/csv.server.ts:508` — CSV export headers
- `app/routes/api+/mobile+/asset.create.ts:146`, `asset.update.ts:160` — mobile API
- `app/routes/api+/asset-index-settings.ts:75`
- `app/modules/asset/data.server.ts:608`, `app/modules/asset/service.server.ts:4061`
- asset overview, asset new/edit

Two consequences: the full `group` row (including `createdAt`/`updatedAt`/`organizationId`) now ships in payloads that never asked for it, and the new ordering **silently reorders CSV export columns and asset-index custom-field columns**. That's a user-visible change nobody opted into.

Please either take the group + ordering in a dedicated call for the form, or gate it behind an opt-in flag (`includeGroup?: boolean`). And per [`.claude/rules/code-bearing-entity-list-consistency.md`](.claude/rules/code-bearing-entity-list-consistency.md), prefer a tight `select` over `include: true`:

```ts
group: { select: { id: true, name: true, position: true } }
```

(`{ group: { id: "asc" } }` is also a redundant tie-break given `@@unique([organizationId, name])`.)

### 9. Grouping is all-or-nothing across the whole workspace

```ts
const hasGroups = useMemo(() => customFields.some((field) => field.group), [customFields]);
```

The moment **one** field anywhere in the org is assigned to a group, the Required Fields / Optional Fields split disappears from every asset form for every user. A workspace with 40 fields (10 required) that groups a single one gets: one group section, plus an "Other Fields" section holding the remaining 39 with required and optional intermixed and no longer surfaced first.

There's no incremental-adoption path — touching the feature at all removes the existing organisation for everything else. Could we keep required-first ordering inside each section, or render Required/Optional as a sub-split within "Other Fields"? Happy to discuss the UX here; a screenshot (see top) would help a lot.

### 10. Reorder and delete failures are invisible

`handleMove` fires `fetcher.submit` and nothing ever reads `fetcher.data`, while the page's only error banner renders `useActionData` — which a fetcher submission never populates. So if two admins have the page open and one deletes a group while the other clicks Move Up, the second request fails the completeness check, returns a 400, and the UI just… snaps back. Same for a `P2025` inside the transaction. Please render `fetcher.data?.error?.message`.

### 11. Dead code

`reorderCustomFields` and `updateCustomFieldGroup` are defined, documented, and never called from anywhere.

- `reorderCustomFields` has no UI — field ordering happens through the raw `position` number input, which is precisely the interaction this function looks like it was meant to replace.
- `updateCustomFieldGroup` has no route, so a typo'd group name can only be fixed by deleting and recreating the group.

Either wire them up (a rename affordance on the groups page would be a genuine gap-filler) or drop them from this PR.

Related: the description says groups are reordered "via drag-and-drop", but the implementation is up/down arrow buttons. Please update the description — arrows are honestly fine here.

### 12. `CustomFieldForm` reaches into `useLoaderData` instead of taking a prop

```ts
type FormLoaderData = { groups?: Array<{ id: string; name: string }> };
const loaderData = (useLoaderData<FormLoaderData>() || {}) as FormLoaderData;
```

That's a hand-written assertion with nothing connecting it to either route's real loader type. If a third route ever renders this form — or one loader's `groups` key gets renamed — `groups` silently resolves to `[]`, the select renders with only "None (No Group)", and **every existing assignment is wiped on save**, with typecheck green throughout.

Both routes already thread `categories`, `groupId` and `position` in explicitly. Please pass `groups` the same way. (The `|| {}` is also dead — `useLoaderData` never returns undefined inside a route.)

### 13. Type casts in the row renderer

```ts
fieldTypeToCompMap[field.type as CustomFieldType]?.(field as unknown as CustomField)
```

These exist to paper over the loader type now carrying `group`. `CLAUDE.md`'s TypeScript rule is aimed at exactly this — please type the map's parameter to the shape it actually receives rather than casting through `unknown`.

---

## 🔵 Nits

- **Lint and formatting don't pass.** `settings.custom-fields.groups.tsx` has `import/order` errors on lines 12–13 (`~/utils/http` and `~/utils/http.server` before `~/components/…`), and `prettier --check` fails the file (the 90-char `react-router` import on line 10, among others). `pnpm webapp:validate` before pushing will fix both automatically.
- `DeleteCustomFieldGroupDialog` is exported from a route file — it belongs next to its sibling in `app/components/custom-fields/` (there's already a `delete-dialog.tsx` there).
- `action({ context, request }: LoaderFunctionArgs)` should be `ActionFunctionArgs`.
- `parseData(await request.formData(), …)` runs before `requirePermission`. Not exploitable — the mutations are still gated — but it inverts our usual order and returns a 400 where an unauthorised user should see a 403.
- `getValidationErrors<typeof GroupActionSchema>` over a discriminated union: a missing/invalid `intent` produces an error keyed at `intent`, which the page never renders.
- The "Create New Group" input isn't cleared after a successful create.
- `createCustomFieldGroup`'s JSDoc says it "Calculates the next position atomically", but it's a read-then-write with no transaction — two concurrent creates can land on the same position. Harmless given the name tie-break, but please either drop the word or wrap it in `db.$transaction`. Per [`.claude/rules/comments-describe-code-not-history.md`](.claude/rules/comments-describe-code-not-history.md), comments need to describe what the code actually does.
- `<p>` nested inside `AlertDialogDescription` (Radix renders it as a `<p>`) triggers a DOM-nesting warning. You inherited this from our own `custom-fields/delete-dialog.tsx`, so it's not something you introduced — `components/kits/booking-removal-notice.tsx:176` shows the pattern that avoids it.
- No tests for any of the new behaviour — the grouping/ordering logic in the component, the group service functions, and the reorder payload validation are all good candidates. The only test change here is fixture fields.

---

## Two open questions for us, not blockers for you

**Read view.** Grouping currently only affects the create/edit form. `assets.$assetId.overview.tsx` builds its list via `buildAssetOverviewCustomFields` and is untouched, so a user groups their fields, sees the groups while editing, and a flat list while viewing. Same for asset-index columns and CSV headers. Is form-only the intended v1? I think it's a defensible scope, but let's say so explicitly rather than leave it implicit.

**Field-level ordering UX.** A raw integer "Position" input on each field's edit page is workable but crude — two fields can share a position and the tie-break is name-ascending. Given `reorderCustomFields` already exists, drag-to-reorder (or arrows, matching the groups page) on the custom-fields index list would be a lot nicer. Fine as a follow-up.

---

## One housekeeping note

CI has never actually run on this branch — every `🧪 Test` and `🩺 React Doctor` run shows `action_required` (fork PRs need maintainer approval) or cancelled. I'll approve the workflow run so we get a real typecheck/test/lint signal on the next push.

---

Genuinely appreciate the work here — the feature is wanted and the settings page is well put together. The schema change in §1 is the big one; most of the rest falls out of it or is quick. Shout if any of it is unclear or you'd like to talk through the grouping UX before rewriting.

PR #2744, opened 2026-07-24
fix: allow reducing an oversubscribed QUANTITY_TRACKED booking

### LINKED ISSUE #2725: QUANTITY_TRACKED: an over-reserved booking cannot be reduced through the UI (adjust dialog rejects every value, including 1)
Split out of #2724 because it is fixable immediately and does not depend on the availability-semantics decision in that issue.

Once a `QUANTITY_TRACKED` asset's pool is oversubscribed (`Σ BookingAsset.quantity > Asset.quantity`, so `available` is negative), **there is no way for the workspace to reduce a booking's quantity back down through the UI**. Every recovery affordance is either blocked or circular.

Audited against `origin/main` @ `8838d02a8`.

### 1. The per-row Adjust dialog rejects reductions as hard as increases

`api+/bookings.$bookingId.adjust-asset-quantity.ts:181` compares the submitted **absolute** quantity against `computeBookingAvailableQuantity(...).available`:

```ts
if (quantity > availability.available) { /* 400 */ }
```

`available` is unclamped (`consumption-log/service.server.ts:385`) and the helper excludes only the *current* booking (`:315`). So with `total = 10`, `custody = 0` and ~17 units reserved on other bookings, `available ≈ -7`, and **every positive submission including `1` satisfies `quantity > available`** and throws a 400 reading "Only -7 available".

There is no number the user can enter that is accepted. `0` is not an escape either: the Zod schema requires a strictly positive integer (`:41-44`), so it is rejected before the guard is reached.

The guard is correct in intent. The defect is that it is written as an absolute comparison when a **reduction can never oversubscribe the pool**.

### 2. "Complete the booking" is circular

`checkinBooking` does set `COMPLETE` and would release the units (`booking/service.server.ts:3960-3962`), but the check-in affordance is gated on `bookingStatus?.isOngoing || bookingStatus?.isOverdue` (`components/booking/forms/edit-booking-form.tsx:420`). Reaching `ONGOING` requires a check-out, and check-out is exactly what the insufficient-stock guard blocks (`service.server.ts:2055-2074`).

### 3. "Archive the booking" is gated the wrong way for this case

`archiveBooking` releases the units but is rejected unless the booking is `COMPLETE`, or `RESERVED` with a `to` date already in the past (`booking/helpers.ts:406-422`; `service.server.ts:7905-7911`; the UI dropdown uses the same gate at `components/booking/actions-dropdown.tsx:108`). A **future-dated RESERVED booking**, which is the normal shape of an over-reservation, cannot be archived.

### 4. The Manage-assets picker is not a fallback either

Any QUANTITY_TRACKED asset whose date-scoped `availableQuantity` is `<= 0` is filtered out of the picker's returned items (`manage-assets.tsx:314-321`), and there is no re-add path for assets already on the booking.

## What still works

Two escapes exist, and neither is discoverable from the error message:

- **Remove the asset from the booking.** `removeAssets` (`booking/service.server.ts:9290`) deletes the `BookingAsset` rows inside a transaction with no availability check at all, so it releases units regardless of how oversubscribed the pool is. Wired to the row action at `components/booking/asset-row-actions-dropdown.tsx:191`.
- **Cancel the booking.** Permitted from RESERVED/ONGOING/OVERDUE (`:8024-8028`), sets `CANCELLED` (`:8069`) without deleting rows, and every reserved reader filters to the three active statuses so the units drop out everywhere at once.

Both are heavier than "type a smaller number", and a user hitting "Only -7 available" has no reason to guess either.

## Proposed fix

Make the guard **directional**:

- At `adjust-asset-quantity.ts:181`, always permit a submission that is `<=` the row's current `BookingAsset.quantity`. Only measure an **increase** against availability.
- Either relax the schema at `:41-44` to allow `0` (meaning "remove the slice"), or surface the row's existing Remove action from the same dialog.
- Improve the error copy so a negative pool explains itself and points at Remove / Cancel.

⚠️ Known limitation of this fix, worth recording rather than solving here: it narrows the guard but does not close the TOCTOU window. The availability read deliberately runs on the global `db` client outside the transaction (the file's own comment at `:165-172`), and the row lock at `:174` only serialises writers that also take it, which `updateBookingAssets`' raw upserts (`booking/service.server.ts:7652-7656`) do not. This is acceptable for the directional change specifically, because the newly-permitted operations are reductions, which cannot oversubscribe.

Background on how a pool gets oversubscribed in the first place, and the design question behind it, is in #2724.

cc @DonKoko

iuryeng:
Fixes #2725

## Problem

Once a `QUANTITY_TRACKED` asset's pool is over-reserved by other bookings, `computeBookingAvailableQuantity` returns a negative `available` (`total - inCustody - reserved`, unclamped). The adjust-quantity guard compared the submitted quantity against that unconditionally:

```ts
if (quantity > availability.available) { /* 400 */ }
```

With `available = -7`, every positive submission satisfies `quantity > -7`, including a reduction all the way down to `1`. There was no number a user could enter to recover — confirmed against the exact repro in the issue.

## Root cause

`apps/webapp/app/routes/api+/bookings.$bookingId.adjust-asset-quantity.ts:181` measured **every** submission against availability, regardless of direction. A reduction of this booking's own reservation can never oversubscribe the pool — every other booking's reservation, the custody count, and the total stock are untouched by this booking's own delta — so only an *increase* needs to be checked.

## Fix

- Only measure an **increase** (relative to the current booked quantity) against remaining availability. A reduction always proceeds.
- The "current" quantity is **re-read inside the transaction, under the already-held asset lock** (`tx.bookingAsset.findUniqueOrThrow`), rather than trusting the pre-transaction snapshot — closes a narrow TOCTOU window and matches the existing pattern in `modules/asset/service.server.ts:2628`.
- The rejection message for a still-blocked increase is clamped to never show a negative "available" count, and explains when the pool is already over-committed by other bookings.

## Testing

Added 4 regression tests to the existing mocked-route test (`test/routes-tests/api+/bookings.$bookingId.adjust-asset-quantity.test.ts`, same style as the pre-existing ownership-guard suite — no real DB needed):

- Reduces a booking's quantity even when the pool is oversubscribed (the reported bug) — and asserts `computeBookingAvailableQuantity` is **not even called** for a reduction.
- Still rejects an increase beyond remaining availability, with the clamped/explanatory message.
- Allows an increase that fits within remaining availability.
- Treats a resubmitted, unchanged quantity as a pass-through.

All 11 tests (7 existing + 4 new) pass. `eslint`, `prettier`, and `typecheck` are clean. This PR also went through the repo's Claude-powered pre-commit security reviewer three times across iterations — final verdict: **Approve with notes**, Low risk. Two low-severity notes worth being upfront about:

1. The over-commitment count in the error message is same-org-only info, exposed to a caller who already passed the route's permission gate — flagged as informational, not a blocker.
2. `computeBookingAvailableQuantity` still reads on the ambient `db` client rather than `tx` — **pre-existing**, not introduced by this diff; the asset-level lock still serializes the common case. Noted for a possible follow-up, out of scope here.

The reviewer also asked (twice) to confirm the reduction branch can't be reached with `quantity <= 0` now that it skips the availability check — confirmed: the Zod schema (`AdjustBookingAssetQuantitySchema`, above this hunk) already enforces `z.coerce.number().int().positive(...)`, so `0`/negative values are rejected before this code runs. No change needed there.

## Scope note: a related, narrower frontend gap (not fixed here)

While tracing the consuming dialog (`components/booking/adjust-booking-asset-quantity-dialog.tsx`), I noticed its `maxQuantity` prop is documented as "available + currently booked" but its only caller (`asset-row-actions-dropdown.tsx:231`) passes `asset.quantity` (total workspace stock) instead. For the numbers in this issue that doesn't block anything (a reduction target is well under total stock), which is why the reported symptom is fully explained — and fully fixed — by the backend change alone. But it means a target reduction *larger than total stock* (e.g. total=2, reduce a 17-unit overbooking down to 5) would still be incorrectly blocked client-side before ever reaching this now-fixed backend guard. Filed as #2755 rather than folded into this PR.

## Update after self-review

Two follow-ups on my own diff, pushed as `d8c3cc14e`:

- **`findUniqueOrThrow` → `findUnique` + explicit 404.** The under-lock re-read closes a TOCTOU window, but it opened a narrower one: the row can be removed from the booking by a concurrent request between the pre-transaction lookup and the lock acquisition, and Prisma's P2025 surfaced as an unclassified error rather than the 404 the pre-transaction lookup already returns for the identical condition. Now both paths return the same 404.
- **A regression test for that path** (12 tests in the file now, all passing), plus a note on the shared mock's seeded value — `vi.clearAllMocks()` clears calls but preserves implementations, so the value in `vi.hoisted` is a bootstrap rather than a per-test default.

Branch is also up to date with `main` as of this push.

## What I didn't do

No full local E2E (real Postgres/Supabase + browser) — validated via the route-level test that exercises the real `action()` function with only the DB layer mocked, matching this repo's own established convention for this exact route (the pre-existing 7 ownership-guard tests use the identical technique). Happy to do a fuller manual pass if useful before merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tracked asset quantity adjustments now validate increases against a fresh, lock-protected quantity snapshot.
  * Availability guards apply only to quantity increases; reductions are permitted even when availability is negative.
  * Rejections include improved “over-committed by X” context when applicable, and activity notes now use the correct prior quantity.

* **Tests**
  * Expanded route coverage for increase, reduction, unchanged submissions, over-committed availability, and a 404 case when the locked row is missing.


carlosvirreira:
Heads-up from a parallel effort, so nobody's work gets lost: #2769 reworks quantity availability broadly and includes the same reduce-an-oversubscribed-booking fix you made here.

We ran **your test file unmodified** against that branch — 12/12 pass, including both asserted substrings in the error copy. We're deliberately **not** closing this PR; whichever merges first, the other rebases. Your diagnosis of the dead-end was correct and independently confirmed. 🙏

iuryeng:
@carlosvirreira Thank you — this is a genuinely considerate way to handle an overlap, and running my test file unmodified against your branch *before* saying anything is more diligence than the situation asked for.

For what it's worth, I think #2769 is the better change of the two. Mine fixes the single dead-end reported in #2725; yours reworks the availability formula so that whole class of bug stops recurring, and the ledgered stock edits cover ground my patch does not touch at all. I would rather see the formula fixed once than see my narrower patch land and leave the next instance of this to be found by another user.

So no precedence claim from me, and the rebase offer runs both ways: **if #2769 merges first I will close this one myself** — nothing for you to rebase. My only ask would be that the two regression tests survive in some form, since they pin the exact error substrings a user sees when reducing an over-committed booking, and you have already shown they pass unchanged.

@DonKoko — ordering is entirely your call. The only practical difference today is that this PR is narrow (scoped to #2725) and currently mergeable, while #2769 is the broader rework and shows conflicts against main. Taking the narrow fix now and letting the rework rebase onto it works; so does dropping mine entirely in favour of the rework. I have no preference beyond the users on #2725 getting unblocked.

Thanks again for the care taken here.

DonKoko:
@iuryeng I haven't checked your PR in detail yet, however I have a branch in the works to solve the quantity related bugs already in progress since last week. 
I think the best would be that once its merged, you take a look to see what is overlapping and adjust your PR where needed.

DonKoko:
@iuryeng this was fixed in the PRs mentioned above #2770 and #2780. Can you verify its all resolved now and close your PR. If something is still not handled, please do let me know.

iuryeng:
@DonKoko I verified the current `main` implementation against the scope of this PR.

#2770 covers the directional availability behavior and the related client-side reduction case from #2755. #2780 adds the under-lock quantity re-read that closes the TOCTOU gap handled here. The concurrent-removal 404 and equivalent regression coverage are also present.

I couldn't find any remaining behavior unique to this PR, so it is fully superseded by #2770 and #2780. Closing it now — thanks for pointing me to the merged work.

PR #2750, opened 2026-07-27
perf(assets): move two conditional writes off the /assets index read path

DonKoko:
The GET /assets.data loader fans out ~30 concurrent DB queries per request, so a burst of concurrent advanced-index loads exhausts the 30-connection Prisma pool (P2024). Two conditional WRITES still sat on that synchronous read path; move both off it. No change to what the response renders.

- refreshExpiredAssetImages: keep re-signing expired image URLs synchronously (response still carries fresh URLs), but defer the db.asset.update persistence and backoff bumps into a single fire-and-forget background flush, batched at <=3 concurrent so it can't recreate a write burst. A failed background write is harmless: the URL is already served and the next load re-signs (idempotent).
- fixTeamMembersNames: invoke it fire-and-forget at its 3 loader call sites (it only writes DB rows, never mutates the returned array; user-linked members display via resolveUserDisplayName regardless) and skip blank-named NRMs.

First of a layered fix; the loader fan-out cap and scheduler/connection sizing are tracked separately.

## Summary by CodeRabbit

* **Performance**
  * Asset image refreshes now respond quickly while persistence happens asynchronously in the background, with concurrency limiting to keep writes efficient.
  * Team member name cleanup now runs as fire-and-forget background work to keep load/form reads fast.

* **Reliability**
  * Background persistence issues are logged and no longer affect refreshed image URLs or signing flow.
  * Team member name repair targets only user-linked blank/whitespace names, safely handling missing data and DB failures.

* **Tests**
  * Expanded coverage for deferred background writes, including the background write concurrency limiter.

### DIFF
--- a/apps/webapp/app/modules/asset/service.server.test.ts
+++ b/apps/webapp/app/modules/asset/service.server.test.ts
@@ -503,12 +503,14 @@ describe("refreshExpiredAssetImages", () => {
     await vi.waitFor(() =>
       expect(mockUpdateMany).toHaveBeenCalledWith(
         expect.objectContaining({
-          // Guarded on the original mainImage so a concurrent image replace
-          // can't be clobbered by this deferred write.
+          // Guarded on the original mainImage AND thumbnailImage (a thumbnail is
+          // written here) so a concurrent replace of either can't be clobbered
+          // by this deferred write.
           where: {
             id: "asset-1",
             organizationId: "org-1",
             mainImage: "https://old-signed-url.com",
+            thumbnailImage: "https://old-thumbnail-url.com",
           },
           data: expect.objectContaining({
             mainImage: "https://new-main-url.com",
--- a/apps/webapp/app/modules/asset/service.server.test.ts
+++ b/apps/webapp/app/modules/asset/service.server.test.ts
@@ -490,16 +490,23 @@ describe("refreshExpiredAssetImages", () => {
 
     const result = await refreshExpiredAssetImages(assets);
 
+    // The return value carries the fresh URLs synchronously — re-signing stays
+    // awaited, only the DB persist is deferred.
     expect(result[0].mainImage).toBe("https://new-main-url.com");
     expect(result[0].thumbnailImage).toBe("https://new-thumbnail-url.com");
-    expect(mockUpdate).toHaveBeenCalledWith(
-      expect.objectContaining({
-        where: { id: "asset-1", organizationId: "org-1" },
-        data: expect.objectContaining({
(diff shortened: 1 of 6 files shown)


DonKoko:
Thanks for the review — addressed in the latest commit on this branch. Summary of dispositions:

**Fixed**

- **Stale-image race (Codex P1)** — the deferred persist could overwrite a newer image if `updateAssetMainImage` ran while the write was queued. The buffered write is now a **guarded `updateMany`** whose `where` includes the original `mainImage`; if the image changed between the index read and the flush, it matches 0 rows and the stale re-signed URL is skipped instead of clobbering the new one. This also makes the previous `P2025` handling moot (`updateMany` returns `count: 0` rather than throwing), so the dead inline `P2025` branch and the flush's `P2025` catch are removed.
- **Per-request cap wasn't process-wide (Codex P1)** — replaced the per-request `PERSIST_CONCURRENCY` loop with a **module-level (per-process) semaphore**. Since the Prisma pool is per-process, N concurrent `/assets` loads can no longer each start their own batch; at most `MAX_CONCURRENT_BACKGROUND_PERSISTS` background writes hold a connection at any instant.
- **PII in logs (CodeRabbit, security)** — `fixTeamMembersNames` now logs `teamMemberIds` + `count`, not the full user records.
- **Unbounded `Promise.all` in `fixTeamMembersNames` (CodeRabbit)** — chunked at 3, mirroring the asset flush.
- **Test factory usage (Codex P1)** — the `fixTeamMembersNames` tests are rebuilt on the shared `createTeamMember` factory instead of hand-cast records; the asset tests assert the guarded `updateMany` (including the `mainImage` guard).

**Declined, with reasons**

- **Add a `Promise.race` timeout to each background write (CodeRabbit).** Prisma has no query cancellation, so a race-timeout would free the *logical* semaphore slot while the DB connection stayed held — which **breaks** the ≤`MAX_CONCURRENT_BACKGROUND_PERSISTS` connection cap the semaphore now guarantees. The limiter is the stronger protection: even if a write hangs, held connections stay capped and reads keep the rest of the pool. Adding the timeout would regress that invariant, so it's intentionally omitted.
- **Move `PERSIST_CONCURRENCY` (now `MAX_CONCURRENT_BACKGROUND_PERSISTS`) into `shelf.config.ts`.** That file holds product/feature configuration (feature flags, branding, geocoding) — it contains no pool/concurrency/timeout/scheduler tuning constants today. A documented module-level constant next to the code it governs is the right home for an internal DB-write concurrency knob; the broader connection-sizing work is tracked separately as noted in the PR description.

PR #2751, opened 2026-07-27
fix(working-hours): race-safe first-access create + retry-wiring regression guard

DonKoko:
Two P2 follow-ups that close out the connection-pool-exhaustion hardening.

- Loader-path write race (working hours): getWorkingHoursForOrganization, reached from the highest-traffic authenticated _layout loader, created default working hours with a bare `create` on first access. Two concurrent first-hits for one organization race a unique-constraint violation on WorkingHours.organizationId — the same class already hardened for booking settings. Convert the shared createDefaultWorkingHours helper to an idempotent `upsert` (update:{}), which fixes all three find-then-create call sites (loader, override creator, admin dashboard) at once. Adds a service test mirroring the booking-settings coverage.
- Retry-wiring regression guard: the transient-retry `$extends` was silently dropped once during the monorepo migration. Add a test asserting the client factory still routes operations through withPrismaRetry. Prisma's client is a Proxy (its `$extends` can't be intercepted) and injecting a fake base client perturbs the inferred client type enough to break downstream typechecks, so this is a source-level guard — the cheap regression pattern the repo already uses for un-mockable seams.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved working-hours initialization to be idempotent and safe under concurrent first-time requests, avoiding duplicate-record errors.
  * Improved booking settings and notification settings initialization by recovering from concurrent-create unique-constraint failures via a re-read, while preserving existing error wrapping.
* **New Features**
  * Added a shared notification-only projection for booking notification settings to keep reads consistent.
* **Tests**
  * Expanded unit tests for working-hours (existing, default creation, and failure handling).
  * Expanded unit tests for booking settings/notification settings concurrency recovery.
  * Added a regression test to ensure the database client keeps its retry wiring.

### DIFF
--- a/apps/webapp/app/modules/booking-settings/service.server.test.ts
+++ b/apps/webapp/app/modules/booking-settings/service.server.test.ts
@@ -3,6 +3,7 @@ import { ShelfError } from "~/utils/error";
 
 import {
   BOOKING_SETTINGS_SELECT,
+  getBookingNotificationSettingsForOrg,
   getBookingSettingsForOrganization,
   updateBookingSettings,
 } from "./service.server";
@@ -15,6 +16,7 @@ vitest.mock("~/database/db.server", () => ({
   db: {
     bookingSettings: {
       findUnique: vitest.fn(),
+      findUniqueOrThrow: vitest.fn(),
       upsert: vitest.fn(),
       update: vitest.fn(),
     },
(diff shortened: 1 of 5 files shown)


DonKoko:
Thanks for the review — addressed in the latest commit. Dispositions:

**Fixed**

- **Codex P2 — the upsert didn't actually close the race.** Correct, and a real bug in the first commit: returning a nested relation (`include`) makes Prisma *emulate* the upsert (separate read + create) instead of a native atomic `INSERT … ON CONFLICT`, so two concurrent first-hits could still throw `P2002`. Now the helper catches `P2002` and re-reads the row the winning request created, closing the race regardless of which path Prisma takes. Added a real behavior test for the recovery path (+ a non-`P2002` rethrow test).
- **CodeRabbit (Major) — source guard sliced to EOF.** Fixed: the retry-wiring guard now bounds its slice to the factory body (next top-level `export` or EOF), so a `$extends` / `withPrismaRetry(() => query(args))` reference elsewhere in the file can't keep it green after the wiring is dropped.
- **CodeRabbit — "add a real concurrency test or narrow the claim."** The `P2002`-recovery test above is a genuine behavior test of the race handling; the arg-shape tests were renamed to describe the query contract they actually verify.

**Declined, with reasons**

- **Type the mocks instead of `@ts-expect-error`.** These mirror the accepted `booking-settings/service.server.test.ts` pattern. `vitest.mocked(...).mockResolvedValue(fixture)` reintroduces the exact type friction the suppression sidesteps — Prisma's delegate return types (e.g. `weeklySchedule: JsonValue`) reject the plain fixture object. `@ts-expect-error` is also self-guarding here: it errors if the line stops having a type error, so it can't silently mask signature drift.
- **Use factories for test data.** Same rationale — mirrors the booking-settings test's module-level fixtures, and there is no `WorkingHours` factory in `test/factories/` today. Adding one is reasonable but out of scope for this fix.

**Sibling worth noting:** `getBookingSettingsForOrganization` (the merged read-first fix) has the *identical* emulated-upsert `P2002` race — its `upsert` returns `BOOKING_SETTINGS_SELECT`, which includes a nested relation (`alwaysNotifyTeamMembers → user`), so Prisma emulates it the same way. Codex only reviewed this diff so didn't flag it. Happy to fold the same `P2002` guard into this PR or send a quick follow-up.

PR #2764, opened 2026-07-29
fix(assets): stamp UTC on advanced-index timestamps so non-UTC users see the right time

carlosvirreira:
## The bug

In a UTC-6 workspace, the assets index showed **Created at `27/7/2026, 21:50`** while the same asset's **Activity tab showed `27/7/2026, 15:50`**. Exactly 6 hours apart. Late-evening UTC timestamps also rolled over to the **wrong day**.

## Root cause

Prisma maps `DateTime` to Postgres `TIMESTAMP(3)` — *without* time zone — storing UTC instants. A top-level `$queryRaw` column is fine: the driver knows the column type and decodes it into a real JS `Date`. But the advanced asset index builds its payload with `jsonb_build_object`, and there the driver only sees an opaque blob — Postgres has already rendered the value with **no zone designator**.

Verified against a real database:

```sql
SELECT jsonb_build_object('createdAt', a."createdAt") FROM "Asset" a;
-- {"createdAt": "2026-07-27T19:42:46.459"}     <-- no "Z", no offset
```

Per ECMA-262, a date-time string with no offset is parsed as **local** time. So `new Date(...)` in the browser reinterprets the UTC instant as the viewer's wall clock — the raw UTC clock, shown as if it were local. The Activity tab goes through the normal Prisma path (`Date` → ISO with `Z`), which is why the two surfaces disagreed.

Anyone on `TZ=UTC` sees no difference, which is why this survived so long.

## The fix

Wrap the four affected fields in `to_char(col, 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"')`:

| Field | Column type | Impact |
| --- | --- | --- |
| `createdAt` | `timestamp` | the reported symptom |
| `updatedAt` | `timestamp` | same |
| `mainImageExpiration` | `timestamp` | gates signed-image refresh — skew made URLs re-sign early or late |
| `upcomingReminder.alertDateTime` | `timestamp` | reminder time in the index and CSV |

### Why `to_char` and not `AT TIME ZONE 'UTC'`

The terser cast yields a `timestamptz`, which `to_jsonb` renders in the **session** TimeZone:

```
SET TIME ZONE 'UTC';                -> "2026-07-27T19:42:46.459+00:00"
SET TIME ZONE 'America/Costa_Rica'; -> "2026-07-27T13:42:46.459-06:00"
```

Both denote the same instant and both parse correctly — but the payload shape would depend on ambient server config. `to_char` on the bare column is session-independent. `NULL` still maps to JSON `null`.

### What is deliberately *not* wrapped

`Booking.from` / `to` are `@db.Timestamptz(3)`. Postgres already emits an offset for them, so they are correct today — wrapping them would *introduce* the session-TZ dependency. A blanket fix here would have been a regression. Sort keys also stay typed timestamps so ordering is unaffected (formatting them would make the comparison lexicographic).

## Verification

- **Real DB, real query builder, before vs after**, rendered through the real formatting path with `TZ=America/Costa_Rica`:

  | DB truth (UTC) | before | shown | after | shown |
  | --- | --- | --- | --- | --- |
  | `2026-04-13T06:17:40.494Z` | `…494` | `13/4 6:17 a.m.` ❌ | `…494Z` | `13/4 12:17 a.m.` ✅ |
  | `2026-04-12T05:31:40.572Z` | `…572` | `12/4 5:31 a.m.` ❌ | `…572Z` | `11/4 11:31 p.m.` ✅ |

- **Real `DateS` component** rendered with `es-CR` / `America/Costa_Rica` hints reproduces the exact reported symptom before the change and the correct value after.
- **Negative control**: reverting only the SQL fails the 4 new assertions; the guard tests (timestamptz untouched, sort keys unformatted, aliases pinned) hold either way.
- `pnpm webapp:validate`: **3676 tests pass**, 0 lint errors.

## Notes for the reviewer

- `utcJsonTimestamp` uses `Prisma.raw`, so the `column` parameter is a **closed literal union** of the four column names — routing user input in is a compile error. Please keep it closed.
- The helper calls are **hoisted out of the template interpolation** on purpose, matching the existing `rankOrderBy` note in `buildAdvancedAssetsQuery`: building a nested SQL fragment inline previously tripped esbuild into silently dropping the enclosing function from the production bundle.
- Tests pin the `aq` / `ar` table aliases the wrapper hardcodes, since `Prisma.raw` puts them beyond typecheck's reach.

## Follow-ups (not in this PR)

Two adjacent issues surfaced during the audit, both in CSV export and both needing a product call rather than a mechanical fix:

1. `csv.server.ts` formats DATE **custom fields** with `format(new Date(valueDate), …)`, which uses the *server process* timezone — wrong day on any non-UTC host. Using the already-stored `raw` value (as the UI does) would fix it.
2. The same export mixes conventions: `createdAt` / `updatedAt` are emitted as raw UTC ISO, while `alertDateTime` is emitted in the *viewer's* timezone. Worth picking one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed asset timestamps in JSON responses to consistently use explicit UTC ISO-8601 formatting.
  * Corrected formatting for creation, update, image expiration, and reminder alert timestamps.
  * Preserved accurate booking timestamps and timestamp-based sorting behavior.


DonKoko (changes requested):
We have a component that transforms the dates client side to the correct format based on the user's browser prefs. This doesn't seem necessary to me. Also keep in mind that the date and time format and timezone configuration PR is almost ready so that will resolve it as well.

carlosvirreira:
@DonKoko short version: this PR and the formatting work in #2654 fix different layers, and #2654 cannot close this one on its own.

**It is the payload, not the formatter.** The advanced index is the only surface that builds its rows with `jsonb_build_object`. Prisma maps `DateTime` to `TIMESTAMP(3)` without time zone, and Postgres renders that inside JSON with no zone designator:

```sql
SELECT jsonb_build_object('createdAt', a."createdAt") FROM "Asset" a;
-- {"createdAt": "2026-07-27T19:42:46.459"}     <-- no "Z"
```

ECMA-262 parses that as local time, so `new Date()` has already picked the wrong instant by the time `DateS` runs, and `DateS` then formats that wrong instant faithfully. Rendering the real component with `America/Costa_Rica` hints gives `6:17 a.m.` where the truth is `12:17 a.m.`

Every other surface (Activity tab, asset detail, bookings) goes through Prisma, receives a real `Date`, serializes with a `Z`, and is correct today. That gap between the list and the Activity tab is the reported behaviour.

**Why #2654 does not close it.** I ran its `formatDate` against this payload:

| input | #2654 renders | correct |
| --- | --- | --- |
| `"2026-07-27T19:42:46.459"` (today) | `27/7/2026, 7:42 p. m.` | `1:42 p. m.` |
| `"2026-07-27T19:42:46.459Z"` (this PR) | `27/7/2026, 1:42 p. m.` | `1:42 p. m.` |

Its date-only guard is `/^\d{4}-\d{2}-\d{2}$/`, which matches a bare calendar date only, so a datetime string falls through to `new Date(value)`. Row 2 is #2654 working correctly once the payload carries its zone. The two are complementary.

**Scope.** Four columns wrapped, all `timestamp without time zone`: `createdAt`, `updatedAt`, `mainImageExpiration`, `upcomingReminder.alertDateTime`. `Booking.from` and `Booking.to` are `@db.Timestamptz(3)`, already carry an offset, and are deliberately left bare.

Happy to move this onto `feat/configurable-date-format` instead if you would rather it ships as one change.

DonKoko (approved):

Issue #2724, opened 2026-07-21
QUANTITY_TRACKED: four divergent availability formulas; over-reserved pool renders negative Available and blocks check-out


carlosvirreira:
Reported from support. A `QUANTITY_TRACKED` asset with `Asset.quantity = 10`, zero custody, renders **Available -11** / **Reserved (bookings) 21** on the asset overview, every booking row shows the red **Insufficient stock** badge, and check-out is refused. The workspace has no way to recover through the UI (see #2725).

Audited against `origin/main` @ `8838d02a8`. All line references are full-body reads at that commit.

---

## 1. There are four independent QT availability formulas, and they disagree by construction

This is the headline. There is no shared primitive. Each surface computes "available units for a quantity-tracked asset" its own way:

| # | Surface | Expression | Clamped? | Excludes current booking? | `inKits`? | Dispositions? | Date-scoped? |
|---|---|---|---|---|---|---|---|
| 1 | Asset overview sidebar<br>`assets.$assetId.overview.tsx:362` | `total - inKits - operatorCustody - reserved - checkedOut` | **no** | no | **yes** | no | **no** |
| 2 | Insufficient-stock badge map<br>`bookings.$bookingId.overview.tsx:957-960` | `max(0, total - inCustody - reserved - checkedOut)` | yes | yes | no | no | **no** |
| 3 | Check-out guard `computeBookingAvailableQuantity`<br>`consumption-log/service.server.ts:385` | `total - inCustody - reserved` | **no** | yes | no | **yes** | **no** |
| 4 | Manage-assets picker<br>`manage-assets.tsx:314-315` | `quantity - inKits - inCustody - reserved` | n/a (filtered ≤ 0) | yes | yes | no | **yes** |

Consequences worth calling out:

- **The `-11` is #1, not #3.** The asset-overview loader never calls `computeBookingAvailableQuantity`, directly or transitively. It computes the number inline from two `db.bookingAsset.aggregate` calls (`:294-300` RESERVED, `:304-313` ONGOING/OVERDUE) plus `computeCheckedOutForAsset` at `:317`, combined at `:324-332`. `computeBookingAvailableQuantity` has four call sites (`booking/service.server.ts:2043`, `adjust-asset-quantity.ts:176`, `add-to-existing-booking.tsx:100` and `:157`) and the overview loader is not one of them. **A fix applied to that helper would not change the rendered number at all.**
[...]

Issue #2767, opened 2026-07-29
UX: asset index shows total stock without availability context (quantity-tracked assets)


carlosvirreira:
Found during device testing of the quantity-trust work (PR pending).

**What happens:** the assets index (web AND companion) shows a quantity-tracked asset's TOTAL stock — e.g. `5 pcs`. Opening the detail can reveal `0 available` (units in custody + bookings reserving more). A field user reads the index number as 'available to grab'.

**Not a regression** — this ambiguity predates the quantity-trust work; that work just makes it visible by finally surfacing over-commitment honestly on the detail screen.

**Proposed fix (~1h, both platforms):** add availability context to the index row for quantity-tracked assets — e.g. `5 pcs · 0 available`, or a warning dot when over-committed. Availability is now cheap to compute via the canonical `getBookingPoolAvailability` batch fetcher introduced by the quantity-trust PR (one call per page), so no N+1.

Reported by Carlos, 2026-07-29, on a real device.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2755, opened 2026-07-28
[Bug]: adjust-quantity dialog caps at workspace stock, blocking reductions on an over-committed booking


iuryeng:
### What happened?

`AdjustBookingAssetQuantityDialog` documents its `maxQuantity` prop as:

```ts
/** Maximum quantity the user can set (available + currently booked) */
maxQuantity?: number;
```

Its booking call site passes workspace stock instead:

```tsx
// apps/webapp/app/components/booking/asset-row-actions-dropdown.tsx:231
<AdjustBookingAssetQuantityDialog
  currentQuantity={asset.bookedQuantity ?? 1}   // correct: BookingAsset.quantity
  maxQuantity={asset.quantity ?? undefined}     // ← Asset.quantity = workspace stock
  ...
/>
```

The dialog enforces that value client-side before the request is ever sent:

```ts
// adjust-booking-asset-quantity-dialog.tsx:128
if (maxQuantity != null && qty > maxQuantity) {
  setError(`Only ${maxQuantity} ${unitLabel} available. Please reduce the quantity.`);
}
```

So a booking that reserves **more than total workspace stock** — which is exactly what an over-committed `QUANTITY_TRACKED` pool looks like — cannot be edited from this dialog at all. With `Asset.quantity = 2` and a booking reserving 17 units, every value from 3 upward is rejected client-side, including a *reduction* from 17 to 5.

This is the same class of defect the repo's own `quantity-semantics-per-surface` rule describes: a booking surface reaching for `Asset.quantity` when the row's real multiplier is `BookingAsset.quantity`.

**Relationship to #2725 / #2744:** #2744 fixes the backend guard, which was rejecting reductions outright once availability went negative. That is the reported symptom and it is fully fixed there — for the numbers in #2725 a reduction target sits well under total stock, so this client-side cap never engages. This issue is the narrower, separate case where the target itself exceeds total stock. Filed on its own rather than folded into #2744 to keep that PR scoped to the backend guard.

**Suggested fix:** pass `available + bookedQuantity` (what the prop documents) rather than `asset.quantity`. The booking row loader would need to surface the availability figure, or the prop could be left undefined on this surface and the now-corrected backend guard treated as the single source of truth.

### What is the expected behaviour?

[...]

PR #2769, opened 2026-07-29
fix(assets): one availability formula, ledgered stock edits, honest over-commitment

carlosvirreira:
## Why this exists

Quantity tracking — a flagship we shipped recently — could contradict itself. A product-discovery sweep flagged it, and adversarial verification confirmed every claim against current code and open issues (#2724 Priority-High, #2725, #2755, #2677 — all open, nothing merged). Three concrete failures:

1. **Seven expressions** computed "available" across the app, and they disagreed (some clamped, some not; some subtracted kits, some didn't; one was date-windowed).
2. An oversubscribed pool rendered as a bare **`-3 available`** and blocked the very edit that would fix it — the user's only escape was undiscoverable.
3. Editing an asset's quantity wrote **no ledger entry**. Nobody could answer who changed 40 gloves to 25, when, or why. The code carried a TODO admitting it.

Carlos's framing was "the app must never lie about inputs and outputs." That's the bar this PR is measured against.

## What changed, and the intent behind each part

**One formula, stated per surface.** A canonical module (pure core + batch fetcher) with *named* variants — `custodyScope`, `includeKitSlices`, `window`, `dispositionAware`. Every migrated call site now declares its semantics instead of re-deriving them, which is what `quantity-semantics-per-surface` has been asking for. Each of the seven migrations is behaviour-preserving; the flags exist precisely so a reviewer can see where surfaces legitimately differ. The **custody family is deliberately NOT merged** — it answers a different question (physically present vs bookable) — and both sides now carry cross-referencing JSDoc so the split reads as a decision, not an oversight.

**Recovery instead of dead ends.** Reducing an oversubscribed booking now succeeds; increasing is still blocked. This aligns with the semantics of the still-open community PR #2744 (@iuryeng) — we ran their test file unmodified against this branch and it passes 12/12. **We are not closing or superseding their PR; whoever merges second rebases.** Guard errors now clamp the printed number and name the recovery paths instead of stating a bare failure.

**Honesty in the callout.** This one is worth reading closely, because it's the thesis in miniature. The over-commitment message originally read *"bookings have reserved more units than are in stock."* With 5 total / 5 in custody / 5 reserved, that sentence is **literally false** — bookings reserved exactly stock; the *combination* over-commits — and "release custody", the cure that would actually work, went unmentioned. The copy now names only the commitments actually consuming the pool and only the cures that apply, and it's property-tested. The same class of defect was swept from the two sibling surfaces (the booking-add panel can't know the cause, so it no longer claims one).

**Audit trail.** Quantity/minQuantity edits (form *and* CSV) write `ConsumptionLog` + `ActivityEvent` inside the mutation transaction. `ASSET_QUANTITY_CHANGED` is also emitted by the quick-adjust dialog and both check-in decrement paths — without that, the new report row would have shown only form edits while silently omitting the majority of real stock movement, which would have been a new lie in place of the old one.

**Card reconciliation.** The overview card's `Reserved` / `Checked out` rows now derive from the same disposition-aware row as `Available`. Previously, after a CONSUME/LOSS/DAMAGE check-in, the rows summed to *more than total stock* — on the single screen whose entire job is being trustworthy.

## ⚠️ Deployment requirement

**This PR ships a database migration** (two additive `ActivityAction` enum values). Deploying the code without it makes every quantity edit fail. `fly.toml`'s `release_command` runs `prisma migrate deploy` before machines roll, so the standard path is safe — flagging it because we hit exactly this locally: the app 500'd until the migration was applied.

## Test plan

- **3,743 webapp tests green**, 284 files, `validate` exit 0; companion `tsc`/lint clean.
- New suites: canonical module (incl. a property test that all variants agree on sign and that the ledger identity holds after a disposition), the callout copy (5 cases incl. custody-only and the mixed case the old copy got wrong), CSV batching, quick-adjust and check-in event emission.
- PR #2744's tests run unmodified against this branch: 12/12.
- **Manually verified against a real database**, browser-driven: consistent numbers across overview/booking/picker/badge; deliberate over-commitment shows `0 available` + a truthful callout; reducing the oversubscribed booking succeeds; the edit wrote a ConsumptionLog row and an activity event.
- **On-device** (iPhone, local server): the app shows the same story as web, clamped, with the callout.

## Known residuals, stated rather than hidden

- The mobile booking picker inherits a **pre-existing double-count** of kit-held custody units (`custodyScope: "all"` + `includeKitSlices`). Documented in the module. It's pre-existing behaviour, but this is a *new* surface, so it deserves naming — happy to fix here if you'd rather not ship a known-wrong number on a new screen.
- Reducing a booking to **0** is still rejected (schema requires positive); removing the asset is the path. Pre-existing, unchanged.
- `updateBookingAssets` still adds QT slices with no availability guard, so over-commitment remains *reachable* (#2724) — this PR makes it **recoverable and visible**, not impossible.
- Follow-ups already filed: #2767 (index shows total without availability context).

## Reviewer starting points

`modules/asset/availability.ts` + `.server.ts` (the contract), then the migration list in the module JSDoc, then `quantity-overview-card.tsx` for the copy logic. The security-sensitive hunk is the org-ownership proof added **before** the row lock in `updateAsset` — a foreign asset id previously took a lock and could be distinguished from a nonexistent one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Quantity-tracked assets now show accurate availability across booking, asset, scanner, and mobile views.
  * Over-committed inventory displays clear warnings, affected commitment categories, and recovery guidance.
  * Mobile booking pickers include available quantity, total quantity, and units of measure.
  * Low-stock alerts are triggered after stock reductions and consolidated during bulk imports.
* **Bug Fixes**
  * Adjusting booked quantities now correctly allows reductions, even when inventory is over-committed.
  * Availability and error messages no longer display negative quantities.
* **Activity History**
  * Quantity and low-stock threshold changes now appear in asset activity reports.


carlosvirreira:
Ready for merge from our side — validate green, migration included (must deploy with the code).

Could you send over the OTA signing key so we can publish updates ourselves? Happy to take publishing off your plate going forward.

DonKoko:
Closing in favor of #2770 (merged), which fixes #2724 / #2725 at the root.

The core change here centralizes the availability calculation but keeps the same global/overlap sum, so it doesn't fix the actual cause of #2724 — bookings that don't overlap in time still get summed together, which is what pushes a pool negative and blocks checkout when nothing is really double-booked. #2770 replaces that formula with a windowed, peak-concurrency model (only demand overlapping at the same instant counts), plus the directional guard for the #2725 reduce-to-recover case. Reconciling the two would mean gutting this core anyway, so it's cleaner to close here.

The parts worth keeping, which we'll pull into a follow-up PR:
- **Quantity audit trail** — quantity / min-quantity changes logged as activity events + consumption-log entries (form, CSV, quick-adjust, check-in). Not in #2770.
- **Two hardening fixes** we're missing: the org check before the row lock, and a re-read under lock on the adjust route.
- **Companion-app improvements** (coordinated with the mobile side).

PR #2775, opened 2026-07-31
fix(bookings): let an in-flight booking check out assets a later reservation holds

DonKoko:
An ONGOING/OVERDUE booking could never check out its remaining assets once any overlapping RESERVED booking also held them. That reservation was itself allowed precisely because an ONGOING booking does not "occupy" an asset it has not physically checked out yet, so the priority inverted: a reservation created AFTER the booking started blocked it permanently, with no in-product way out. The asset row still rendered "Available" (the already-booked badge is suppressed on ONGOING bookings), so the action was offered right up until it failed with "Some assets are already booked or checked out elsewhere".

`hasAssetBookingConflicts` now accepts an `ignoreReservedConflicts` option, and `partialCheckoutBooking` passes it only when the booking is already in flight (new `outranksReservations` helper). Everything else is unchanged: a RESERVED-vs-RESERVED clash still blocks, and an asset physically CHECKED_OUT on another in-flight booking still blocks. The losing reservation is told at its own check-out, exactly as before.

Both bookings get a system note naming the clash so the override is never silent - the reservation's owner would otherwise not discover it until their own check-out failed. User-supplied booking names and asset titles go through the markdoc wrappers, with a regression test asserting a name containing a Markdoc tag cannot inject one.

## Summary by CodeRabbit

* **New Features**
  * Progressive checkouts can proceed with individually tracked assets that overlap reservations.
  * Booking notes identify overridden reservations, affected assets, and required follow-up.
  * Notes safely format booking and asset names, including special characters.

* **Bug Fixes**
  * Improved conflict handling for ongoing and overdue bookings.
  * Active bookings and quantity-tracked asset limits continue to block unavailable conflicts.
  * Checkout protections prevent assets concurrently checked out or placed into custody from being reassigned.

### DIFF
--- a/apps/webapp/app/modules/booking-note/service.server.test.ts
+++ b/apps/webapp/app/modules/booking-note/service.server.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, beforeEach, vi } from "vitest";
 import {
   createBookingNote,
   createSystemBookingNote,
+  createSystemBookingNotes,
   deleteBookingNote,
   getBookingNotes,
 } from "./service.server";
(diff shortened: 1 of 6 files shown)


DonKoko:
@coderabbitai full review

PR #2781, opened 2026-07-31
fix(kits): keep the custodian filter's value a TeamMember id

carlosvirreira:
## Problem

On **/kits**, filtering by custodian returns **"No kits found"** whenever you type to search before picking — even when that person demonstrably holds kits. Picking the same person straight from the list works.

Same defect on the QR "link to existing kit" screen, which uses an identical picker.

## Root cause

Both pickers rewrite the id of the selected item:

```ts
transformItem={(item) => ({
  ...item,
  id: item.metadata?.userId ? item.metadata.userId : item.id,
})}
```

`metadata` exists only on records from `/api/model-filters`. `useModelFilters` renders the **route loader's** records until a search query is entered, then switches to that endpoint — and the loader's `db.teamMember.findMany` results have no `metadata` field. So the two paths submit different things:

| Path | `item.metadata` | Value submitted |
|---|---|---|
| Pick from the list | absent | `TeamMember.id` ✅ |
| Type, then pick | present | `TeamMember.userId` ❌ |

`getPaginatedAndFilterableKits` consumes it as `custody: { custodianId: teamMember }`, and `KitCustody.custodianId` is a **TeamMember** FK, so the user-id form matches nothing.

## Why it stayed hidden

Two conditions have to line up:

1. **More than 12 team members** in the workspace — below that the picker has all rows loaded and filters client-side, so it never calls the endpoint.
2. **A custodian linked to a user account** — NRMs have `userId: null`, so the `? :` fallback happens to yield the correct id for them.

## Fix

Drop the rewrite. The value must stay the `TeamMember.id` the query expects, which is also what the assets advanced-filter custodian picker already does (`transformItem: (item) => item`).

## Verification

Reproduced and fixed on a local dev server against a workspace seeded to 14 team members with a kit held by a user-linked custodian.

**Before** — typed "Carlitos", selected the custodian who holds the kit:
- Result: `No kits found`
- URL: `?teamMember=694aa6ec-…` (the **user** id)

**After** — identical steps:
- Result: `1 kit`, the correct row
- URL: `?teamMember=cmn8vu17z0022ooi12w6n8aax` (the **TeamMember** id)

Picking from the list without searching behaves the same before and after. `pnpm turbo typecheck` and ESLint clean.

## Related

Same loader-vs-search shape divergence as #2778, on a different surface. That PR fixes the booking pickers, where the mismatch made rows render as nothing; this one fixes the kit custodian filter, where it makes the filter match nothing. The two changes touch different files and are independent.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved custodian filtering to consistently retain the selected team member across list selections and typed searches.
  - Ensured filtering behaves consistently across kit listings and kit detail views.
  - Resolved mismatches that could prevent results from appearing for the selected custodian.


carlosvirreira:
@codex review

Issue #13, opened 2023-03-08
Implement tailwind theme


DonKoko:
To ensure a consistent style around the website, we need to implement a out tailwind theming based on the figma designs.
We will start with some basic elements that are required for building the UI and build on top of that.

- [x] Colors - including all variations. Follow untitled UI
- [x] Buttons
- [x] Spacing
- [x] Typography
- [x] Shadows

DonKoko:
Based on our discussion yesterday I have implemented the following sizes for headings:

```
Display 2xl = H1
Display xl = H2
Display lg = H3
Display md = H4
Display sm = H5
Display xs = H6
```

This however seems inconsistent with the design of the app. 
If I look at the biggest heading on the page, based on basic HTML semantics and structure that should be an H1. However when I look at it it uses `display-sm` which above is defined for H5. 

![Image](https://user-images.githubusercontent.com/7840007/224270777-640d6241-3281-447d-8987-566a0a6e0c87.png)

At this point I am confused what to do. I will override the default styles we defined, however I am concerned that in the future that will cause a domino effect when updating the UI.

DonKoko:
@carlosvirreira I have implemented the primary button color based on the new design you made on friday. What I am missing or I simply dont know how to see are hover/focus states of the buttons.
Also it would be good to know what the transitions are on the hover states.

carlosvirreira:
Hi @DonKoko 

Our buttons have 4 states.

1) Default 
2) Hover 
3) Focused 
4) Disabled

![Screenshot 2023-03-13 at 09 57 17](https://user-images.githubusercontent.com/12449255/224653440-35d2c30c-93b7-4354-89b4-fdee9d363b57.png)

For the example above: 

1) Default  -  background = base/white | border (inside) 1px - gray 300 | (drop shadow) Shadow / XS (blur = 2 / Y = 1) 101828 (5% opacity)
2) Hover - background = gray 50| border (inside) 1px - gray 300 | (drop shadow) Shadow / XS (blur = 2 / Y = 1) 101828 (5% opacity)
3) Focused -   base/white | border (inside) 1px - gray 300 | (drop shadow) Shadow / XS focused 4px - gray 100 (blur = 2 / Y = 1) 101828 (5% opacity) + F2F4F7 - spread 4 (199% opacity)
4) Disabled -

DonKoko:
This is done for now.

PR #2820, opened 2026-08-06
fix(bookings): make model reservations countable, fulfillable and correctly logged

carlosvirreira:
## What is in this PR

Four fixes, all on booking model reservations. They were found in sequence: fixing the first exposed the second, and verifying the second exposed the last two.

1. **Counts** — the bookings index and the booking page gave two different numbers for the same booking. This is what the customer reported.
2. **Stuck bookings** — a booking could reserve "3 of this model", and adding a matching asset from **Manage assets** did not tick it off. Unfulfilled reservations hard-block check-out, so the booking could not go out.
3. **Duplicate activity note** — adding one asset wrote the same log line twice.
4. **Missing activity note** — adding a kit from the kit page wrote no log line at all.

**This PR contains a database migration.** It adds one nullable column, one partial index, and one foreign key. It is additive and safe to run before the app deploys.

---

### The report

A customer flagged that the bookings index showed **7 assets** for a booking whose overview said **11 items**.

Both numbers were correct. Neither was labelled:

- Index counted concrete assets → 7
- Overview counted asset/kit rows **plus** outstanding model reservations, and called the total "items" → 7 + 4 = 11

## Why the mismatch wasn't the real bug

Model reservations are units nobody has matched to a physical asset yet, so a booking carrying them is not ready to leave. Blending them into a count hid exactly that:

- A half-assigned booking looked identical to a finished one in the list.
- A pure book-by-model booking read as **"0 assets"** with no indication anything was on it.

Correcting the count to 11 everywhere would have left that intact. The customer would have trusted the number and still been short when loading out.

## What changed

**1. Reservations move out of the assets table into their own titled section above it**, matching what the index assets drawer already did. The overview was the surface that diverged.

**2. Every count now says what it counts.** "Items" was the root problem: a noun that names nothing can't be reconciled with anything.

| | Before | After |
|---|---|---|
| Reservations | interleaved as table rows | own section, `18 units across 4 models` |
| Assets & Kits header | `20 items` | `18 assets and 2 kits` |
| Same, no kits | `11 items` | `7 assets` — matches the index verbatim |
| Index | `7 assets` | `7 assets` + amber `4 units unassigned` |

A kit is one row holding several assets, so a row count can never equal the index's asset count. Naming both kinds makes the remainder self-evident, and each kit row already states its own member count. Nothing to compute, nothing to look up in another panel.

Each reservation row states its outstanding work in words (`3 units to assign`, `1 of 2 units still to assign`) rather than a bare `× N` in a Qty column that means *booked* units on asset rows and *remaining* units on reservations.

### Also fixed

- **`totalItems` no longer has the reservation count added to it.** The `clientLoader` re-derived it *without* the addend, so re-sorting the list silently changed the header count while the rows stayed put.
- **`ListTitle` counted selectable entities, not rows.** `itemsGetter` flattens each kit row into its members plus the kit object, so a 10-row page with two kits read `17 items out of 20`.
- **One shared component** replaces two hand-maintained copies of the section that had already diverged.
- **Outstanding units counted in one place**, in units not request rows, so a 3-unit reservation counts as 3 things to go and find.
- `ListTitle` rendered `0 item` for an empty list.

## On the "count kits as 1 item" setting

Raised during triage as a possible cause of the 11. Ruled out, two independent ways:

1. `countKitsAsSingleUnit` feeds only `partialCheckinProgress` and `calculateBookingLifecycleProgress`, both progress-bar computations. `shapeBookingAssets`, which produces `totalPaginationItems`, takes no settings parameter at all.
2. The setting only changes how *kits* collapse. The reported booking has **Kits 0**, so every branch it gates returns the same thing there.

## Testing

- `booking-model-requests.test.ts` — 8 tests on the counting helper (units vs rows, partial fulfilment, over-fulfilment clamping, serialised dates)
- `booking-model-reservations-section.test.tsx` — 8 tests on the section
- `booking-assets-column.describe-rows.test.ts` — 5 tests on the header phrasing
- Full suite **332 files / 4320 tests** green. Typecheck clean across all 4 packages. Lint at the 5 pre-existing warnings.
- Fulfilment exercised against a **live database with the real services** across 8 shapes: picker, scanner, partial, several units in one call, re-save, over-delivery, no reservation, and no actor threaded through.
- Every new guard is mutation-tested — removed it, confirmed the test fails, restored it.
- Both activity-log fixes driven through the **real UI** in a browser, with the resulting rows read back from the database.

Exercised locally against seeded data on four shapes:

1. 7 assets + 4 outstanding units (the reported shape)
2. 0 assets + 5 units (pure book-by-model — previously looked empty)
3. 2 kits + 2 quantity-tracked + 18 standalone + 5 reservations incl. one fully fulfilled (correctly hidden)
4. The same, paginated at 10/page and under an active search

---

## 2. A booking could reserve units it had no way to fulfil

Found while testing the section above.

A reservation means "this booking needs N units of this model, any N". Until this PR, exactly ONE code path turned that promise into real assets: the **scanner**. `materializeModelRequestForAsset` had a single production caller.

So adding the exact asset the booking had reserved, from **Manage assets**, left the reservation outstanding. And `checkoutBookingWritesWithinTx` hard-blocks check-out while any reservation is outstanding. The error message even said *"Scan matching assets to fulfil the reservation"*, which was literally true and a dead end.

Verified against a live database with the real services. Same asset, same 1-unit reservation:

| Path | Reservation after adding the reserved asset | Check-out |
|---|---|---|
| Manage assets (`updateBookingAssets`) | `0/1` — untouched | **hard blocked** |
| Scanner (`addScannedAssetsToBooking`) | `1/1` fulfilled | allowed |

A workspace with no working scanner could create a reservation and then had two options: delete the reservation it had correctly made, or leave the booking stuck.

### The fix

Fulfilment is now a property of **an asset landing on the booking**, not of the device it arrived with. Every add-assets path routes through one new chokepoint, `fulfilModelRequestsForAssets`:

- Manage assets (`updateBookingAssets`)
- Web scanner (`addScannedAssetsToBooking`)
- Asset index bulk add, `api/assets.add-to-booking`
- Mobile `api/mobile/bookings.add-scanned-assets`, and the companion app through it

Web, mobile and companion now behave identically, and a future surface inherits it by construction rather than by remembering to call it.

I deliberately did **not** build a separate "pick assets" screen. That would have been a third path bolted alongside, while the path operators use every day stayed broken.

### Two things that would have shipped wrong

**Re-saving the dialog would have double-counted.** The manage-assets dialog reposts the operator's full selection on every save, and the service derives `addedAssetIds` from *everything the call touched*, not from what is new. Keying fulfilment off that lets a plain re-save decrement the reservation again — a 3-unit reservation reaching `3/3` with only two physical assets behind it, with nothing in the system to flag it. Only genuinely new rows may discharge a reservation now, pinned by a test that fails when the guard is removed.

**The schema.** My first attempt declared `bookingModelRequestId` as a Prisma `@relation` and pushed the extended client's generics past TS's recursion limit (TS2321) in two services. `assetKitId`, two fields above it in the same model, documents that exact failure and its fix. It is now a plain column with a DB-enforced FK, matching that precedent.

### New column: `BookingAsset.bookingModelRequestId`

Records which reservation each asset discharged. Before, that link existed only as a sentence in an activity note, so *"which assets fulfilled the LED reservation?"* was not answerable from data. `ON DELETE SET NULL`, so a vanished reservation costs a row its provenance and never its place in the booking.

---

## 3 & 4. The activity log was wrong in two places

Found while reading the log to verify fix 2.

**Adding one asset wrote the same line twice.** Two writers fire in one request: `updateBookingAssets` writes a booking note as a side effect, and the manage-assets route writes its own summary. For a single INDIVIDUAL asset the strings are byte-identical, because `formatUnitCount` returns null off `QUANTITY_TRACKED` and the service's wrapper then collapses to the same bare link the route emits. A reader cannot tell whether one asset was added or two.

The service's only opt-out was a non-empty `kitIds`. That flag was standing in for *"the caller writes its own note"*, which is why the kit routes were fine and the asset route was not — a caller with no kits had no way to say so. Replaced with an explicit `skipBookingNote`.

Applied at both manage-assets call sites. The second is a pure quantity edit on assets **already** on the booking, so it was logging a phantom "added X to the booking" for an asset that never arrived.

**Adding a kit from the kit page wrote nothing.** Same root fault, opposite symptom: `kits.$kitId.assets.add-to-existing-booking` forwards `kitIds` (suppressing the service's note) and never wrote its own, so the booking's feed was silent. The same action from the booking's page logged it correctly. Fixed, with kit names fetched so both doors produce an identical note. That route gets its first test file.

Proved in one feed, before and after:

```
13:31  added ZZ UI-PROOF Projector A1 to the booking.   <- twice
14:06  added ZZ NOTE-DUP-PROOF A2 to the booking.       <- once
```

---

## Known, deliberately out of scope

Model reservations write **booking notes** (via `createSystemBookingNote`) but no **`ActivityEvent`** rows, and `ActivityAction` has no members for them. So the Activity tab shows these changes, but reports — which read `ActivityEvent` — cannot count or aggregate them. Needs an enum migration, so it is being handled on a separate branch.

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated section for outstanding model reservations, with assignment actions controlled by booking permissions.
  * Added badges showing unassigned model units in booking lists.
  * Improved booking row counts and labels for assets and kits across desktop, mobile, and overview screens.
  * Updated empty states to distinguish missing booking dates from bookings without concrete assets.
  * PDF previews now consistently include outstanding model reservations.

* **Bug Fixes**
  * Fulfilled reservations are excluded from displays and counts.
  * Corrected handling of partial and over-fulfilled reservations.


carlosvirreira:
## Manual test guide

@DonKoko — checklist for testing this by hand. About 15 minutes for the core path.

<details>
<summary>Click to expand</summary>

# Manual test guide — PR #2820

**Model reservations: counts, fulfilment, activity log**

Time needed: about 15 minutes for the core path, 25 minutes for everything.

---

## What you are testing

A booking can reserve **units of an asset model** without naming which physical
items. For example "3 of PT-DZ21K projector, any 3". The booking then owes 3
units, and Shelf blocks check-out until all of them are matched to real assets.

This PR changes four things about that.

| # | Before | After |
|---|---|---|
| 1 | Bookings list said `7 assets`, the booking page said `11 items` | Both agree. Reserved units sit in their own section. |
| 2 | Only the camera scanner could match a reserved unit | Any way of adding the asset matches it |
| 3 | Adding one asset wrote the activity line twice | Writes once |
| 4 | Adding a kit from the kit page wrote no activity line | Writes one |

**Fix 2 is the one that matters most.** Before this PR, a workspace with no
working scanner could create a reservation and then had no way to check the
booking out. The only escape was to delete the reservation.

---

## Before you start

You need, in one workspace:

- An **asset model** with at least **4 INDIVIDUAL assets** attached to it, all
  Available (not in custody, not booked over your test dates).
- A second asset model with at least 1 asset, for the multi-model check.
- One **kit** with at least one asset in it.

> Model reservations only work with `INDIVIDUAL` assets. Quantity-tracked
> assets are not part of this flow.

If you need to create a model: **Assets → Asset models → New**, then open an
asset and set its model.

---

## 1. The count the customer reported

**Setup**

1. Create a booking. Give it dates in the future. Do not add anything yet.
2. Open **Manage assets → Models** tab.
3. Reserve **3** of your model. Save.
4. Go back into **Manage assets → Assets** and add **2 ordinary assets** that
   are NOT of that model.

**Check the booking page**

- [ ] A section titled **"Unassigned model reservations"** sits **above** the
      Assets & Kits list.
- [ ] Its subtitle reads **`3 of 3 units still to assign, across 1 model`**.
- [ ] The row underneath reads **`3 of 3 units still to assign`**.
- [ ] The **Assets & Kits** header below reads **`2 assets`**, and there are
      exactly 2 rows.
- [ ] In **Booking statistics** on the right, **Unassigned model units** shows
      an amber **3**.

The point: every number describes the rows directly beneath it. Nothing to
reconcile.

**Check the bookings list**

- [ ] Go to **Bookings**. That booking shows **`2 assets`** and an amber
      **`3 units unassigned`** pill.
- [ ] The `2 assets` here matches the `2 assets` on the booking page exactly.

**Check the drawer**

- [ ] In the bookings list, open the assets drawer for that booking (click the
      asset count).
- [ ] It shows the same reservations section and the same wording as the
      booking page.

---

## 2. The main fix: matching a reserved unit without a scanner

This is the part that was broken.

1. Stay on the same booking.
2. Open **Manage assets → Assets**.
3. Tick **one asset that belongs to your reserved model**.
4. **Confirm**.

**Check**

- [ ] The reservations section now reads **`2 of 3 units still to assign`**.
- [ ] The header reads **`2 of 3 units still to assign, across 1 model`**.
- [ ] **Assets & Kits** now says **`3 assets`**.
- [ ] **Booking statistics** shows **Unassigned model units: 2**.

> On `main` this number would still say **3**. The asset would be on the
> booking, but the reservation would be untouched.

**Now finish it**

5. Add the remaining **2 assets** of that model the same way.

**Check**

- [ ] The **"Unassigned model reservations" section disappears completely.**
- [ ] The **Unassigned model units** row disappears from Booking statistics.
- [ ] The bookings list no longer shows the amber pill for this booking.
- [ ] **Reserve** the booking, then **Check out**. It goes through with no
      error.

That last step is the whole point. Before this PR you could not reach it
without a scanner.

---

## 3. The entry point

1. On a booking that still has an outstanding reservation, click the **⋮** menu
   on a reservation row.

**Check**

- [ ] The menu offers **"Select assets to assign"** first, then
      **"Scan to assign"**.
- [ ] "Select assets to assign" opens **Manage assets**.
- [ ] "Scan to assign" opens the scanner as before.

---

## 4. The scanner still works and agrees

If you have a scanner or a phone handy.

1. New booking, reserve **2** of your model.
2. Scan one matching asset in **Scan to assign**.

**Check**

- [ ] Reservation drops to **`1 of 2 units still to assign`**.
- [ ] The result is identical to what the picker produced in step 2.

Both doors, same outcome. That is the design.

---

## 5. Activity log: no duplicates, nothing missing

**Duplicate check**

1. On any booking, add **one single asset** through **Manage assets**.
2. Open the **Activity** tab.

- [ ] There is exactly **ONE** line saying `<you> added <asset> to the booking.`
- [ ] On `main` there would be **two identical lines**.

**Quantity edit check**

3. On a booking with a quantity-tracked asset, change only its **quantity**
   through Manage assets.

- [ ] The log records the **quantity change**.
- [ ] It does **NOT** say "added ... to the booking" for an asset that was
      already there.

**Missing note check**

4. Go to a **kit's page → Assets → Add to existing booking**, and pick a
   booking.
5. Open that booking's **Activity** tab.

- [ ] There is a line naming the **kit by name**, e.g.
      `<you> added <Kit name> to the booking.`
- [ ] On `main` there would be **nothing at all** here.

**Reservation match check**

6. On a booking where you matched a reserved unit in step 2, open **Activity**.

- [ ] There is a line like
      `<you> assigned <asset> (<model>) to this booking — 2 × <model> remaining.`
- [ ] The countdown in that line matches what the page shows.

---

## 6. Edge cases worth a minute

**Over-adding**

1. Reserve **1** unit of a model, then add **3** assets of that model at once.

- [ ] Reservation shows fulfilled and the section disappears.
- [ ] All **3** assets are on the booking. The extras are ordinary assets.
- [ ] Nothing shows a negative or a count above the reserved amount.

**Re-saving without changing anything**

2. On a booking with a **partly** matched reservation (say `1 of 3` matched),
   open Manage assets and press **Confirm** without changing the selection.

- [ ] The reservation count does **not** move.

> This is the one that would corrupt data silently if it were wrong. The dialog
> resends the whole selection each time, so a careless fix would count the same
> assets again.

**Finished bookings**

3. Find or make a booking with a reservation, then **Cancel** it.

- [ ] The reservation row turns **grey**, not amber.
- [ ] Wording changes to **`never assigned`**, not "still to assign".
- [ ] The bookings list shows **no amber pill** for it.
- [ ] The **⋮** menu offers no assign actions.

Same for **Complete** and **Archived**.

**Two models at once**

4. Reserve **3** of model A and **2** of model B on one booking.

- [ ] Header reads **`5 of 5 units still to assign, across 2 models`**.
- [ ] Match one asset of model A. Header becomes **`4 of 5 ...`**, model A's row
      becomes `2 of 3`, model B's row stays `2 of 2`.
- [ ] The row numbers always add up to the header numbers.

---

## 7. Other places the numbers appear

- [ ] **PDF**: booking page → **Actions → Generate PDF**. Outstanding
      reservations are listed and the counts match the screen.
- [ ] **Mobile / narrow window**: shrink the browser to phone width. The
      reservations section is readable and the counts still match.
- [ ] **Booking update email**: if you trigger one, outstanding reservations
      appear with the same numbers.

---

## 8. Permissions

Worth one pass as a non-owner.

1. Sign in as a **BASE** or **SELF_SERVICE** user who can see the booking.

- [ ] They can **see** the reservations section and the counts.
- [ ] They do **not** get assign or remove actions they are not allowed to use.
- [ ] Nothing renders blank or broken for them.

---

## Deploy notes

**This PR contains a database migration.**

```
20260810120000_add_booking_asset_model_request_provenance
```

It adds one nullable column to `BookingAsset`, one partial index, and one
foreign key with `ON DELETE SET NULL`. It only adds. It changes and removes
nothing, and existing rows get `NULL`, which is the correct value for every
asset that was added directly.

Fly runs migrations automatically on release, so no manual step is needed.

The column records **which reservation each asset filled**. Before, that link
only existed as a sentence in the activity log, so "which assets filled the
projector reservation?" could not be answered from the data.

---

## If something looks wrong

Useful things to capture:

- The booking id from the URL.
- A screenshot of the reservations section **and** the Booking statistics panel
  together, so the two counts can be compared.
- The Activity tab for that booking.

The number to distrust first is any count that does not match the rows directly
under it. That was the original bug and it is the shape this PR is built to
prevent.

</details>

DonKoko (changes requested):
Reviewed at `4bc4403` with a multi-agent pass (6 finder angles, 50 candidates, each verified independently; 8 were refuted and dropped). 15 findings survived — 10 confirmed, 5 plausible.

**The design is right.** One `fulfilModelRequestsForAssets` chokepoint that every add-assets surface routes through is the correct answer to "the scanner was the only path that discharged a reservation", and it does fix the stuck-booking dead-end. The write-up is excellent — the `preExistingAssetIds` re-save guard and the TS2321 schema note both saved me time.

What blocks merge is that most findings cluster on that chokepoint's **edges**, and they are the same class of bug the PR set out to eliminate: a booking that checks out with fewer physical assets than it promised, or one that stays blocked with no way out.

---

## Blocking — the chokepoint's four edges

**1. Rollback on removal ignores the provenance column this PR adds.** `modules/booking/service.server.ts:9937-9946`

`removeAssets` builds `removalsByModel` from every removed asset that merely *has* an `assetModelId`. It never reads `bookingModelRequestId`.

Reserve 2 × Dell Latitude, tick 3 matching assets in Manage assets. Two discharge the reservation; the third is an ordinary add carrying no provenance. Untick that third asset — `fulfilledQuantity` drops 2 → 1, `fulfilledAt` clears, and check-out is hard-blocked again while **both discharging assets are still on the booking**. The only escape is deleting the reservation the operator correctly made.

This was near-unreachable before (scanner-only fulfilment). Routing Manage assets through the chokepoint makes it routine. The new column is written but never read — reading it here is the fix.

**2. The scan path can discharge the same reservation twice.** `modules/booking/service.server.ts:12107`

`updateBookingAssets` computes `preExistingAssetIds` (L8161) and passes only genuinely-new rows, because discharging is the consequence of a unit *arriving*. `addScannedAssetsToBookingWithinTx` passes `allScannedAssetIds` verbatim.

Reserve 2 × Canon C300. Scan a kit whose member CAM-7 is a C300 → 1/2. Then scan CAM-7's own QR (or let mobile `bookings.add-scanned-assets` post it — that endpoint never filters either). The two partial uniques let the standalone row coexist with the kit row, so it decrements again to 2/2, `fulfilledAt` stamped. Guard at L2128 passes; the booking leaves with one camera where two were promised.

**3. The pre-existing check misses kit-driven rows.** `modules/booking/service.server.ts:8158`

The `preExistingAssetIds` query has no `assetKitId: null`. An asset present *only* as a kit slice reads as pre-existing, so when the operator picks it to discharge a reservation, the brand-new standalone row (upsert at L8174) is filtered out at L8246 and nothing discharges. Booking now holds the asset twice and is still hard-blocked — the exact dead-end this PR removes.

**4. Kit members get stamped with reservation provenance.** `modules/booking/service.server.ts:8246`

`assetTypeRows` is built from `addedAssetIds`, which unions standalone ids with `kitSlices[].assetId`. But the schema comment on `BookingAsset.bookingModelRequestId` says the column is "`null` for assets added directly — picked, scanned onto the booking, **or pulled in via a kit**".

Reserve 2 × C300 as loose units, then add a kit containing one. The section silently drops to "1 of 2 units still to assign", and the raw UPDATE at L8258 stamps the kit-driven row (the `ORDER BY ("assetKitId" IS NOT NULL)` tiebreak falls through when no standalone row exists). The operator packs one loose camera instead of two, and "which assets discharged this reservation?" answers with a camera committed to the kit.

Related: `.claude/rules/kit-members-via-kit-slices.md` — this bug class travels in packs; worth grepping the siblings.

---

## Surface parity

**5. "Select assets to assign" opens an unfiltered picker.** `components/booking/model-request-row-actions-dropdown.tsx:139`

The link is a bare `/bookings/:id/overview/manage-assets` — no `bookingFrom`/`bookingTo`/`hideUnavailable`/`unhideAssetsBookigIds`. Every other entry point sends all four (`booking-assets-column.tsx:107-112`).

Without `hideUnavailable`, `getPaginatedAndFilterableAssets` skips `availableToBook`, the custody exclusion, the overlapping-booking exclusions and `assetKits: { none: {} }` (`asset/service.server.ts:864-1010`). The picker lists assets under custody, kit members, and assets already checked out on an overlapping booking. Ticking one either fails the save or double-books a unit that the "Add assets" button on the *same page* would have hidden.

**6. The unassigned-units pill never renders outside `/bookings`.** `components/booking/list-bookings-content.tsx:299`

`item.modelRequests` comes only from the bookings-index `extraInclude`. `assets.$assetId.bookings.tsx` renders the same row component via `<BookingsIndexPage />` with `extraInclude: { tags: … }` only — so `countUnassignedModelUnits(undefined)` returns 0 and the pill silently disappears. A booking with 4 unassigned units shows the amber pill on `/bookings` and nothing at all on the asset's Bookings tab, where an operator concludes it's ready to go out. `hasStockConflict` was wired for both surfaces; this needs the same.

**7. `getAssignableAssetsForModel` has zero callers.** `modules/booking-model-request/service.server.ts:122`

Exported with a JSDoc calling itself "the pick-from-list counterpart to scanning" for #2831, but nothing references it. The menu item it was written for points at the generic picker instead — which has no model column, no model filter, no assignable-units hint. Asset titles like "Projector — Studio B" don't name the model, so an operator with a 3 × PT-DZ21K reservation can't tell which rows discharge it. The dead helper also duplicates `getAssetModelAvailability`'s exclusion clauses that its own JSDoc says must be kept in step. Either wire it up or drop it.

---

## Audit trail

**8. `skipBookingNote` doesn't suppress the ActivityEvent.** `modules/booking/service.server.ts:8289`

The flag suppresses the human-readable note, but `recordEvents` still writes `BOOKING_ASSETS_ADDED` for every id in `addedAssetIds`. The pure-quantity-edit path the flag was added for (`manage-assets.tsx:821`) still emits a phantom "added" into the machine-readable feed reports aggregate. `preExistingAssetIds` is computed 130 lines earlier and passed only to the fulfilment helper — it's the filter `recordEvents` needs too.

**9. One countdown note per asset on bulk fulfilment.** `modules/booking-model-request/service.server.ts:991`

Tick 20 matching assets and save once → 20 near-identical notes ("… 19 × remaining.", "… 18 × remaining.") plus the route's summary, all INSERTed inside the interactive transaction. That wording was designed for the scanner's one-at-a-time volume; generalising the helper to bulk callers needs an aggregated note.

**10. The booking note is now lost if a per-asset note fails.** `routes/_layout+/bookings.$bookingId.overview.manage-assets.tsx:669` *(plausible)*

`skipBookingNote: true` removes the service's independently-guarded note, but the replacement is sequenced *after* the per-asset `createNotes` in the same try/catch. One rejecting `Promise.all` member (asset deleted concurrently) jumps to `catch` at L750 and `createSystemBookingNote` at L743 never runs — assets committed, feed silent. Previously the service's own best-effort try/catch survived that.

---

## Lower tier

**11. Per-asset round-trips inside the transaction.** `modules/booking-model-request/service.server.ts:1086` — the loop runs one sequential `findUnique` per model-tagged asset with no "does this booking have reservations at all" pre-check. Assets-index select-all → `api/assets.add-to-booking` with ~800 model-tagged assets means 800 serial round-trips inside a `$transaction` with no `timeout` override (Prisma default 5s) → P2028, full rollback, on an operation that worked before. One `count` up front short-circuits the common case.

**12. Read-then-absolute-write race.** `modules/booking-model-request/service.server.ts:971` *(plausible)* — no row lock, no `{ increment: 1 }`. Two operators (web picker + companion) both read `fulfilledQuantity = 1`, both write the literal `2`. Two assets land, the request reads 2/3, check-out stays blocked with no unassigned unit left to find. The comment reasons about the intra-transaction race but not the cross-transaction one, which this PR newly makes reachable.

**13. Fulfilment doesn't check `asset.type`.** `modules/booking-model-request/service.server.ts:1086` *(plausible)* — `getAssetModelAvailability` and `getAssignableAssetsForModel` both restrict the pool to `INDIVIDUAL`; the fulfilment loop accepts anything with a matching `assetModelId`. The UI producers are closed (form hides the selector for QT, `bulkUpdateAssetModel` skips it) — but CSV import, direct API writes and INDIVIDUAL → QT conversion aren't. Adding `type: AssetType.INDIVIDUAL` makes both halves agree on what a model unit is.

**14. UI "outstanding" is now narrower than the checkout guard.** `utils/booking-model-requests.ts:98` *(plausible)* — the helper adds `fulfilledQuantity < quantity` on top of `fulfilledAt === null`, but the guard at `booking/service.server.ts:2128` still blocks on `fulfilledAt: null` alone. A row in that gap is hidden from the overview, drawer, PDF, statistics and pill while check-out still hard-fails — nothing left to remove or edit. I couldn't find a live producer, so this is a defensive-guard removal rather than a reachable path, but the old code rendered the row and this hides it.

**15. `ListTitle` pluralises on the wrong variable.** `components/list/list-title.tsx:172` — `{totalItems} {rowCount === 1 ? singular : plural}`. The noun agrees with `items.length`, the number printed is `totalItems`. Since this change is specifically about the header lying, both branches should derive the noun from the number they actually print.

---

## Checked and not flagging

- **The migration is sound.** Nullable column + partial index + FK, all idempotent, additive. The plain-column-with-DB-FK approach matches the existing `assetKitId` precedent in `20260525131507_bookingasset_kit_discriminator` — consistent with repo practice, not a workaround. `ON DELETE SET NULL` is the right choice and the reasoning in the migration header is correct.
- **CodeRabbit's functional items are addressed** — the empty state branches on dates, and `hasItems` accounts for reservations. Its remaining comments are test-factory nitpicks.
- **A tempting false positive:** three verifiers killed the claim that the asset form and bulk dialog let QT assets carry a model. Both do restrict it. That's why #13 is plausible rather than confirmed.

Findings 1–4 are what I'd fix before merge; the rest can be triaged.

carlosvirreira:
@DonKoko — addressed at `5a12029`. Thank you for this; the four blocking findings were all real and all in the same place.

## The root cause behind blockers 1–4

The PR never said **which row** discharges a promise. A reservation is for LOOSE units from the free pool, so the rule is now explicit:

> A reservation is discharged by a **new standalone** `BookingAsset` row, and by nothing else.

Findings 2, 3 and 4 fall out of that single rule rather than needing three separate patches.

| # | Fix |
|---|---|
| 1 | Rollback now reads `bookingModelRequestId` instead of grouping by `assetModelId`. You were right that the column was written and never read — reading it here is the fix. |
| 2 | The scan path now takes the same pre-existing-standalone filter `updateBookingAssets` had. |
| 3 | `preExistingAssetIds` is scoped to `assetKitId: null`, so a kit-only asset is no longer mistaken for pre-existing. |
| 4 | Kit rows neither discharge nor get stamped. The stamping SQL is now `WHERE "assetKitId" IS NULL` — you were right that `ORDER BY` is a preference, not a constraint, and it fell through exactly as you described. |

Each of your four scenarios reproduced against a live database, then re-run after the fix:

```
PASS  F1: removing an extra asset leaves the reservation fulfilled   before 2/2, after 2/2, blocked=false
PASS  F2: kit scan + standalone scan discharges at most one unit     after kit 0/2, after standalone 1/2
PASS  F3: kit-slice-only asset discharges when picked standalone     0/1 -> 1/1, blocked=false
PASS  F4: adding a kit discharges nothing, stamps no kit row         0/2 fulfilled, 0 kit rows stamped
```

## Surface parity and audit trail

- **5** — the menu item now receives the booking-window-filtered picker URL from the parent, as a required prop rather than a defaulted one.
- **6** — this was worse than reported. **Five** surfaces render the shared bookings row: `/bookings`, the asset tab, the **kit tab**, the team-member tab and `/me/bookings`. Only the first had `modelRequests`. All five now do.
- **7** — dropped `getAssignableAssetsForModel`. Dead code that must be kept in step with a sibling is worse than no code.
- **8** — `skipBookingNote` now filters `recordEvents` too, using the `preExistingStandaloneAssetIds` set that was already computed.
- **10** — you flagged this as plausible and it was right. Per-asset notes now have their own guard, so one rejecting note cannot skip the booking-side note the route is now solely responsible for.

## Lower tier

- **11** — short-circuits on an outstanding-reservation count. A select-all bulk add was one serial round-trip per model-tagged asset inside a 5s interactive transaction.
- **12** — now `{ increment: 1 }`. You were right that the comment reasoned about the intra-transaction race and not the cross-transaction one this PR makes reachable.

## Not addressed

**Finding 9** — one countdown note per asset on bulk fulfilment. Real, and I have left it. Aggregating that wording changes note semantics for the scanner path too, and I would rather that be its own reviewed change than folded into an already large PR. Happy to do it here if you would prefer it landed together.

344 test files / 4533 tests green, typecheck clean across all 4 packages.

DonKoko (changes requested):
Re-reviewed at `5a12029`, checking each of the 15 findings against the code rather than the commit message. **11 fixed, 1 deferred with good reason, 3 untouched, 1 new problem.**

Blocking again for one interaction — details below — but the framing in `5a12029` is better than my finding list was. Identifying that findings 2, 3 and 4 shared one root cause (the PR never defined *which* row discharges a promise) and making the rule explicit — only a new **standalone** `BookingAsset` row does — is the right level to fix it at, and all three fall out together. Replacing the stamping SQL's `ORDER BY ("assetKitId" IS NOT NULL)` with a hard `WHERE "assetKitId" IS NULL` is the right instinct too: a preference falls through, a constraint refuses. And on finding 6 you found four sibling surfaces where I'd only spotted one.

---

## Blocking — findings 12 and 14 now interact

`fulfilledQuantity: { increment: 1 }` (`booking-model-request/service.server.ts:885`) closes the lost update on that column. But `alreadyFulfilled` (L858) and `justCompleted` (L867) are still computed from the **pre-write** read, so the write is relative while the guard and the stamp are absolute.

Reservation for 2 × Dell, `fulfilledQuantity = 0`, web picker and companion concurrently:

```
T1  findUnique      → 0        T2  findUnique      → 0
T1  justCompleted = (1 === 2) = false
T2  justCompleted = (1 === 2) = false
T1  update: increment → 1, no stamp, commits
T2  update: blocks on row lock, then increments committed 1 → 2, no stamp
```

Final row: `fulfilledQuantity = 2`, `quantity = 2`, `fulfilledAt = NULL`.

That state is unreachable from the UI in both directions:

- `getOutstandingModelRequests` filters it out — `2 < 2` is false — so it's invisible on the overview, drawer, PDF, statistics and pill.
- The checkout guard (`booking/service.server.ts:2131`) still matches on `fulfilledAt: null` alone and hard-blocks with "2 × Dell still unassigned".
- There's no visible row to edit or delete, and `removeBookingModelRequest` refuses while `fulfilledQuantity > 0`.

**This is finding 14's scenario.** I marked it *plausible* in the first review specifically because I couldn't find a producer; this commit supplies one. Before the change the same race produced a visible, recoverable under-count (1/2, section still showing work to do). Now it produces an unrecoverable one — so the fix moves the failure mode in the wrong direction, in the exact race it was written to close.

The stale `alreadyFulfilled` guard also permits over-fulfilment that the previous absolute write capped: quantity 1, two concurrent fulfilments, both read 0, both pass the guard, both increment → `2/1`.

The added test asserts the payload shape only (`toEqual({ increment: 1 })`), so nothing exercises the race.

### Suggested fix

Cheapest durable option closes 12 and 14 together: make `fulfilledQuantity >= quantity` the single source of truth for **both** the checkout guard and `getOutstandingModelRequests`, and treat `fulfilledAt` as a cosmetic timestamp. Finding 14 *is* the two-sources-of-truth split; removing the split removes the class rather than this one instance of it.

If you'd rather keep `fulfilledAt` load-bearing, make the guard atomic instead of advisory:

```ts
const { count } = await tx.bookingModelRequest.updateMany({
  where: { id: existing.id, fulfilledQuantity: { lt: existing.quantity } },
  data: { fulfilledQuantity: { increment: 1 } },
});
if (count === 0) return { matched: false }; // another tx took the last unit
// stamp from the post-write value, not the pre-write snapshot
```

Either way it needs a test that drives two fulfilments against one request row.

---

## Not addressed, and not mentioned

Neither appears in the commit message, so I can't tell whether these were declined or missed.

**13 — `materializeModelRequestForAsset` still has no `asset.type` check.** Confirmed: between the function entry and the update there is no reference to `asset.type` or `AssetType`. The `AssetType.INDIVIDUAL` occurrences at L125/134/146 are all in `getAssetModelAvailability`. So the availability math counts INDIVIDUAL only while fulfilment accepts anything with a matching `assetModelId`. The UI producers are closed, but CSV import, direct API writes and INDIVIDUAL → QT conversion aren't. One clause in the existing guard.

**15 — `ListTitle` still pluralises on the wrong variable.** `list-title.tsx:172` is unchanged: `` `${totalItems} ${rowCount === 1 ? singular : plural}` ``. Narrow, since the branch only runs when `perPage >= totalItems` — but this change was specifically about the header telling the truth, and both branches should derive the noun from the number they actually print.

---

## Residual on finding 8

The fix works for the reported scenario — a pure quantity edit on a standalone asset no longer emits a phantom `BOOKING_ASSETS_ADDED`. One thing to tidy:

```ts
newlyStandaloneAssetIds.has(assetId) || !preExistingStandaloneAssetIds.has(assetId)
```

`newlyStandaloneAssetIds` is built as `standaloneAssetIds.filter(id => !preExistingStandaloneAssetIds.has(id))`, so the first operand is a strict subset of the second and the expression reduces to the second alone.

The comment above it says an asset counts as added if this call "created its standalone row **or any of its kit-driven rows**" — but nothing checks whether a kit row was created, so an asset present only as a kit slice always passes. Low impact in practice (the kit insert is `ON CONFLICT DO NOTHING` and the picker filters already-added kits), so this is about the comment and the code disagreeing rather than a live bug. Either implement the kit half or narrow the comment to what the predicate actually says.

---

## Accepted

**9 — one countdown note per asset on bulk fulfilment.** Agreed, and the deferral is documented rather than silent. Aggregating that wording does change note semantics and belongs in its own review.

---

## Verified fixed

Checked against the code, not the message:

| # | What I verified |
|---|---|
| 1 | `rowsBeingDeleted` selects `bookingModelRequestId`; rollback groups by request id, skips null rows, re-checks `request.bookingId !== id` |
| 2 | `preExistingStandaloneScannedIds` computed; deduped standalone scans only; mobile endpoint reaches the same code |
| 3 | Query scoped with `assetKitId: null`, renamed to match |
| 4 | Candidates from `newlyStandaloneAssetIds`; stamping SQL scoped `AND "assetKitId" IS NULL`; scan path hard-codes `bookingModelRequestId: null` on kit slices |
| 5 | `manageAssetsUrl` a required prop — required rather than defaulted is the right call, a silent fallback is how the two entry points drifted |
| 6 | `modelRequests` included on the asset, kit, my-bookings and team-member bookings loaders |
| 7 | `getAssignableAssetsForModel` deleted; duplicated exclusion clauses go with it |
| 10 | Per-asset notes and the booking-side note in separate try/catch blocks |
| 11 | Early return on empty list plus an indexed `count` short-circuit before the loop |

Two notes on scope. The branch merged `main`, so the PR diff now carries ~120 unrelated files from already-merged work — I reviewed only `5a12029`. And the four blocking fixes are the kind that need a live database rather than a read; you say all four were reproduced before and after, which I'm taking at your word.

Once 12/14 are closed together and 13/15 get a yes-or-no, this is good to go.

carlosvirreira:
@DonKoko — addressed at `6a7191d`. You were right, and the framing was right: my fix moved the failure mode in the wrong direction, in the exact race it was written to close.

## The 12/14 interaction

`{ increment: 1 }` closed the lost update on the column and left the capacity check and the stamp reading a pre-write snapshot. Relative write, absolute guard. That is how `2/2` with `fulfilledAt = NULL` becomes reachable, and your read of the consequence is exactly right — invisible to the UI, blocking check-out, and `removeBookingModelRequest` refuses to delete it. Nothing to click.

I took **both** of your suggestions, because they close different halves.

**The claim is now one conditional statement.** Capacity check, increment and stamp move together:

```sql
UPDATE "BookingModelRequest"
SET "fulfilledQuantity" = "fulfilledQuantity" + 1,
    "fulfilledAt" = CASE WHEN "fulfilledQuantity" + 1 >= "quantity"
                      THEN COALESCE("fulfilledAt", NOW()) ELSE "fulfilledAt" END
WHERE "id" = $1 AND "fulfilledQuantity" < "quantity"
RETURNING "fulfilledQuantity", "quantity"
```

Zero rows back means another transaction took the last unit, so that asset lands as an ordinary add. `remaining` is derived from what committed, not from the snapshot. This also closes the over-fill you spotted (`2/1`), which the stale guard permitted.

**And the split is gone.** The checkout guard now reads outstanding through the same `getOutstandingModelRequests` the overview, drawer, PDF, statistics and pill use. Your point that finding 14 *is* the two-sources-of-truth problem is the right level: with one predicate the gap has nowhere to exist. Prisma cannot compare two columns in a `where`, so a SQL predicate here could only ever approximate the helper — fetching the booking's requests and filtering in JS keeps them in lockstep by construction.

Proven against a live database with genuinely concurrent transactions, not a simulation:

```
PASS  cannot land fully-delivered-but-unstamped AND blocking   2/2, fulfilledAt=set, UI work=false, blocks=false
PASS  completion is stamped when the last unit lands           2/2, fulfilledAt=set
PASS  never over-fills past the reserved quantity              1/1 with 3 concurrent claims
```

You were also right that nothing exercised the race. Added two unit tests that drive two claims at one row, plus an assertion that the capacity predicate is *in* the statement — asserting the payload shape is precisely how a fix for one race shipped while opening a worse one. Mutation-tested: removing `AND "fulfilledQuantity" < "quantity"` fails the suite.

## 13 and 15 — both yes, neither was declined

Both were misses, not decisions. My commit message said nothing about them, which is worse than declining them.

- **13** — `materializeModelRequestForAsset` now requires `AssetType.INDIVIDUAL`. Your reasoning holds: the UI producers are closed but CSV import, direct API writes and an INDIVIDUAL → QT conversion are not, and the two halves should agree on what a model unit is regardless of how the row got its model.
- **15** — `list-title.tsx` now pluralises on the number it prints. Narrow, but this change is about the header telling the truth.

## Finding 8 residual

Correct on both counts. The predicate reduced to its second operand, and the comment described a kit check that was not there. It now genuinely accounts for new kit slices — the pre-existing read selects `assetKitId`, so a re-submitted kit slice (which the `ON CONFLICT DO NOTHING` insert creates nothing for) no longer counts as an add.

## On scope

Fair point about the merge noise — `main` was 101 commits ahead and CI had only ever tested this branch in isolation, so I merged rather than leave that untested. Reviewing `5a12029` alone was the right call.

And on taking the live-database claims at my word: that is a reasonable thing to want stronger. The scripts are throwaway by design, but if you would rather they were committed as integration tests I will do that — say the word.

344 test files / 4535 tests green, typecheck clean, lint at the 6 pre-existing warnings.

DonKoko (approved):
Approving. Re-checked `6a7191d` against the code; all four outstanding items are fixed.

**The 12/14 interaction** is closed at the right level. Making the claim one conditional statement — capacity check in the `WHERE`, increment, and the stamp computed from the post-write value — puts all three on the same side of the write, which was the whole problem. I specifically checked the parts that fail silently: `$queryRaw` rather than `$executeRaw` (needed for `RETURNING`), the id parameterised, no `@map`/`@@map` on `BookingModelRequest` so the literal column names are correct, and `COALESCE` preserving an existing stamp rather than moving it. The loser path is right too — under READ COMMITTED the second transaction re-evaluates the `WHERE` after the first commits, gets zero rows, and its asset lands as an ordinary add. Same predicate closes the `2/1` over-fill.

Taking both suggestions rather than picking one was the better call. The atomic statement fixes the race; routing the checkout guard through `getOutstandingModelRequests` removes the split that made the gap state reachable at all. Filtering in JS there is correct — Prisma can't compare two columns in a `where`, so a SQL predicate could only ever approximate the helper.

**13, 15 and the finding-8 residual** are all clean. On 8 the two operands are now genuinely independent, backed by widening `preExistingRows` to carry `assetKitId`.

---

**One thing for a follow-up, not a blocker.** The two new concurrency tests exercise the right paths, but the concurrency is simulated — `installClaimSimulator` re-implements the `WHERE`/`CASE` in JS against a mutable object, so they pin the caller's handling of both outcomes rather than that Postgres behaves as claimed. The string assertions on the capacity predicate and `COALESCE` hedge that, and they're what `.claude/rules/raw-sql-respects-prisma-map.md` asks for as a minimum.

So yes — please do commit those live-DB scripts as integration tests. A fix for this race has now shipped twice with a gap in it, and both times the unit tests were green. That's the signal that this particular statement deserves a test that actually hits Postgres. Happy for it to land separately.

Also fine as-is, just noting it: the perf short-circuit at `booking-model-request/service.server.ts:1055` still counts on `fulfilledAt: null` rather than the shared helper. It's a strict superset of outstanding so it can never skip work that needs doing, and it's moot now the gap state can't occur.

Finding 9 (one countdown note per asset on bulk fulfilment) stays deferred as agreed.

Nice work on this one — the "a reservation is discharged by a new standalone row, and by nothing else" framing is what turned four separate-looking bugs into one rule, and it's worth carrying into the follow-up.

PR #2841, opened 2026-08-12
fix(bookings): stop the unavailable-assets badge firing on partial quantity-tracked custody

carlosvirreira:
## The report

A customer (2026-08-07):

> "I have set up a **'tracked by quantity'** asset and **assigned some of the items to a custodian**. I **can book the unassigned items as expected**, however when the booking is created it says it **'Includes unavailable assets'** — do you know how to fix this?"

He had not set anything up wrong. The badge is simply wrong.

## Reproduced on production

ACME, booking **`ZZ YORK REPRO - partial custody QT (safe to delete)`** — left in place so you can see it yourself:

- `Sandbag (Counterweight) 24×45cm`, QUANTITY_TRACKED, **12 units**
- **4 assigned** to John Doe → status `Partial custody`, Available **8**
- Booking takes **1 free unit**. The picker offered it, the service accepted it.
- The list shows the amber **"Includes unavailable assets"** badge.

Read from the DOM: `badgePresent: true`. Every unit in that booking is free.

## Cause

`list-bookings-content.tsx` asked:

```ts
!ba.asset.availableToBook || hasCustody(ba.asset.custody)
```

`hasCustody` is only `custody.length > 0` — no notion of **how many** units are held.

That is the right question for an `INDIVIDUAL` asset, which is one indivisible thing: if someone holds it, it is not free. It is meaningless for a quantity-tracked one, where 4 of 12 assigned leaves 8 bookable.

**This is a divergence, not a missing feature.** The server already draws the line:

- `getBookingFlags` (`booking/service.server.ts`) excludes `QUANTITY_TRACKED` from `hasAssetsInCustody`, with a comment saying partial custody leaves the rest bookable.
- `availability-label.tsx:90` and `:407` both guard with `!isQuantityTracked`.

The bookings-list badge was the **last inline copy** that never got the same fix.

## The fix

The predicate moves to `bookingIncludesUnavailableAssets` in `~/utils/booking-assets`, so the component and its tests read one definition instead of the component holding a second opinion.

Per-unit availability for QT assets is still enforced at the service layer by `computeBookingAvailableQuantity()` when assets are added, so nothing is lost by dropping them from this cosmetic signal.

## Testing

8 tests. **Two fail against the current production logic**, so they encode the bug rather than only the fix:

| Case | Old logic | With fix |
|---|---|---|
| QT asset, some units in custody | ❌ fails | ✅ |
| QT asset, several custodians | ❌ fails | ✅ |
| INDIVIDUAL in custody → **still warns** | ✅ | ✅ |
| `availableToBook: false`, either type → **still warns** | ✅ | ✅ |
| Mixed booking, one genuinely unavailable → **still warns** | ✅ | ✅ |
| Terminal bookings stay quiet | ✅ | ✅ |
| Missing custody relation | ✅ | ✅ |
| Empty booking | ✅ | ✅ |

The control cases are the point: this removes a false positive without blinding the badge.

Also verified by running the **exact production row shape** through both predicates:

```
OLD predicate (live on prod today) -> badge shown: true
NEW predicate (the fix)            -> badge shown: false
CONTROL: INDIVIDUAL asset in custody -> badge shown: true
```

Full suite green. Typecheck clean. Lint at the pre-existing warnings.

## Why it matters more than a cosmetic badge

A badge that cries wolf is worse than no badge: it trains people to ignore the cases it exists for. This customer's reaction was to assume he had configured something incorrectly.

## Not in scope

The **Available count** reading 9 when the truth is 4 is a separate defect, fixed in #2830. Both were reported by the same customer. Shipping only one leaves him with half of what he was told was coming.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved unavailable-asset warnings on booking lists.
  - Warnings now accurately identify individual assets that are unavailable or in custody.
  - Quantity-tracked assets in custody no longer trigger an incorrect warning.
  - Warnings are hidden for completed or otherwise terminal bookings.
  - Bookings with missing asset details are handled more reliably.

### DIFF
--- a/apps/webapp/app/components/booking/list-bookings-content.tsx
+++ b/apps/webapp/app/components/booking/list-bookings-content.tsx
@@ -15,9 +15,9 @@
  */
 import type { Prisma } from "@prisma/client";
 import { useLoaderData } from "react-router";
-import { hasCustody } from "~/modules/custody/utils";
 import type { BookingsIndexLoaderData } from "~/routes/_layout+/bookings._index";
 import { BADGE_COLORS } from "~/utils/badge-colors";
+import { bookingIncludesUnavailableAssets } from "~/utils/booking-assets";
 import {
   canAssignModelUnits,
   countUnassignedModelUnits,
@@ -216,10 +216,11 @@ export default function ListBookingsContent({
     bookingAssets: rawItem.bookingAssets ?? [],
   };
 
-  const hasUnavaiableAssets =
-    item.bookingAssets.some(
-      (ba) => !ba.asset.availableToBook || hasCustody(ba.asset.custody)
-    ) && !["COMPLETE", "CANCELLED", "ARCHIVED"].includes(item.status);
+  // Shared, tested predicate — see `bookingIncludesUnavailableAssets`.
+  const hasUnavaiableAssets = bookingIncludesUnavailableAssets(
+    item.bookingAssets.map((ba) => ba.asset),
+    item.status
+  );
 
   /**
    * Pull this booking's slice of the page-wide dispositioned-quantity
--- a/apps/webapp/app/utils/booking-assets.ts
+++ b/apps/webapp/app/utils/booking-assets.ts
@@ -1,4 +1,4 @@
-import { AssetStatus, KitStatus } from "@prisma/client";
+import { AssetStatus, AssetType, KitStatus } from "@prisma/client";
 import type { PartialCheckinDetailsType } from "~/modules/booking/service.server";
 
 /**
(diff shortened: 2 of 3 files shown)


carlosvirreira:
Closing as a duplicate of #2830, which already fixes this and does it better.

My mistake: I searched open PRs by title for availability keywords, and #2830's title is about custody writes, so its "Also in here: the bookings-list badge" section never surfaced. I read `main`, saw the bug live, and concluded no fix existed. I never opened #2830's diff.

#2830 already has it, and its version is the one to keep:

```ts
!ba.asset.availableToBook ||
(!isQuantityTracked(ba.asset) && hasCustody(ba.asset.custody))
```

Logically identical to what I wrote, but it uses the shared `isQuantityTracked` helper, which matches the sibling guards in `availability-label.tsx:90` and `:407`. Mine hand-rolled `asset.type !== AssetType.QUANTITY_TRACKED`, which is the same check spelled differently — one more place to drift.

It also already carries the control case I cared about, in `list-bookings-content.test.tsx`:

- does NOT flag a quantity-tracked asset that merely has units on custody
- **DOES flag an individual asset in custody**
- DOES flag a quantity-tracked asset marked as not available to book

Nothing here is worth a competing diff on the same file.

One thing from this branch that may still be useful: a live reproduction on ACME, left in place deliberately — booking `ZZ YORK REPRO - partial custody QT (safe to delete)`, with `Sandbag (Counterweight) 24×45cm` at 12 units, 4 assigned to John Doe, and 1 free unit booked. The badge fires on it today. Handy for eyeballing #2830 before merge, and I will clear it once that lands.

PR #2885, opened 2026-08-18
fix(security): org-scope every note and override deletion

DonKoko:
## What

Four cross-organization IDOR findings with one root cause: a delete keyed on an
id the caller supplies, with no organization in the `where` clause.

detail.dev findings **D019, D062, D074, D079**. Independent of #2882/#2883/#2884
— different files entirely, so it can merge in any order relative to them.

## The pack

None of these models has an `organizationId` column, so each has to be scoped
through its parent:

| Model | Scoped through | Finding |
| --- | --- | --- |
| `Note` (asset) | `asset.organizationId` | D019 |
| `LocationNote` | `location.organizationId` | D062 |
| `AuditNote` | `auditSession.organizationId` | D074 |
| `WorkingHoursOverride` | `workingHours.organizationId` | D079 |

`deleteBookingNote` and `deleteTeamMemberNote` already scoped correctly, so this
makes the family consistent rather than inventing a shape.

## Severity is not uniform — and one was understated

Three scoped by `userId`, so the exposure was an author reaching their own note
in an organization they had since **left**.

The audit scan-details route scoped by **neither** user nor organization:

```ts
await db.auditNote.delete({ where: { id: noteId } });
```

Any authenticated user who could reach that route could delete **any** audit
note in **any** organization. Its type-check lookup (`findUnique({ where: { id }})`)
was equally unscoped.

## What the required param found that the findings didn't

`organizationId` is a **required typed param**, per
`.claude/rules/org-scope-user-supplied-ids.md`, so the compiler enumerated the
call sites instead of a grep. That surfaced two extra holes:

- **`updateAuditAssetNote`** has the identical unscoped lookup — the same bug on
  the *update* path. Fixed here.
- The sibling sweep found **`updateWorkingHoursOverride` with no scoping at
  all** — not even a preceding lookup. Any authenticated caller could edit
  another organization's override by id. Never reported by the scan.

## Implementation notes

Prisma's `delete`/`update` require a unique `where`, so relation-scoped
mutations become `deleteMany`/`updateMany` plus a count check. Still one query —
a relation filter, not an extra round trip.

A miss now throws **403** rather than returning `count: 0`. Silently reporting
"nothing deleted" let a refused cross-org delete read as an ordinary no-op. The
wrapping catches re-throw `ShelfError` so that 403 is not repackaged as a
generic 500 — one of those (`deleteWorkingHoursOverride`) I had missed and the
pre-commit reviewer caught.

## Checked rather than assumed

- **Parent FKs are all non-nullable** (`Note.assetId`, `LocationNote.locationId`,
  `WorkingHoursOverride.workingHoursId` are `String`, not `String?`), so scoping
  through the parent cannot strand a row as permanently undeletable.
- **`updateAuditAssetNote`, `deleteAuditAssetNote` and both working-hours
  override functions currently have no callers.** The fixes are still correct as
  exported API surface, but it means D074's *live* exposure was the two route
  handlers, not the service. Worth knowing before weighting the severity.

## Left deliberate

The audit scan-details delete is scoped by organization but **not** by author,
unlike its sibling route `audits.$auditId.note.tsx`. Deleting an audit note
there reads as a role-gated, org-wide capability — an auditor tidying a session
shouldn't be blocked by who typed the note. Cross-**org** deletion was the bug,
and the `auditSession` filter closes it. There's a `// why:` at the call site so
the asymmetry reads as a decision rather than an oversight.

## Tests

145 pass across the four modules. New coverage pins the actual vulnerability:
the delete is asserted to carry the relation filter (`asset: { organizationId }`
etc.), a cross-org note is refused with the sink never reached, and a miss
raises 403 instead of a silent zero-count.

## Summary by CodeRabbit

* **Bug Fixes**
  * Note updates and deletions are now securely restricted to the correct organization.
  * Prevented cross-organization access to audit, asset, location, and general notes.
  * Working-hours override updates and deletions now enforce organization-level permissions.
  * Unauthorized or unavailable records return clear permission errors.
  * Success notifications appear only after deletions complete successfully.
  * Improved error handling preserves meaningful access-denied responses instead of generic failures.

### DIFF
--- a/apps/webapp/app/modules/audit/asset-details.service.server.test.ts
+++ b/apps/webapp/app/modules/audit/asset-details.service.server.test.ts
@@ -1,3 +1,4 @@
+import type { Prisma } from "@prisma/client";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
 // Mock the database
@@ -130,13 +131,30 @@ describe("audit asset details service", () => {
         updatedAt: new Date(),
       };
 
-      const updatedNote = {
+      // Typed against the query rather than `as any`: the re-read selects
+      // `displayName`, and a loose fixture silently omitted it.
+      const updatedNote: Prisma.AuditNoteGetPayload<{
+        include: {
+          user: {
+            select: {
+              id: true;
+              firstName: true;
+              lastName: true;
+              displayName: true;
+              email: true;
+              profilePicture: true;
+            };
+          };
+        };
+      }> = {
         ...existingNote,
+        type: "COMMENT",
         content: "Updated content",
         user: {
           id: "user-1",
           firstName: "John",
           lastName: "Doe",
+          displayName: "John Doe",
           email: "john@example.com",
           profilePicture: null,
         },
@@ -144,9 +162,7 @@ describe("audit asset details service", () => {
 
       vi.mocked(db.auditNote.findFirst).mockResolvedValue(existingNote as any);
       vi.mocked(db.auditNote.updateMany).mockResolvedValue({ count: 1 });
-      vi.mocked(db.auditNote.findFirstOrThrow).mockResolvedValue(
-        updatedNote as any
-      );
+      vi.mocked(db.auditNote.findFirstOrThrow).mockResolvedValue(updatedNote);
 
       const result = await updateAuditAssetNote({
         noteId: "note-1",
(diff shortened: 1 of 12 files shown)


DonKoko:
Both findings from this review are fixed in `16349568f`, and the out-of-diff one was the important one.

### Apply the authorization predicate to both mutations — fixed

You're right, and this was my misjudgement rather than an oversight. I saw exactly this pattern during the sibling sweep and talked myself out of it on the grounds that the preceding `findFirst` had already verified ownership. That reasoning is wrong: the lookup authorizes a *past* state, and `update({ where: { id: noteId } })` then writes to whatever the row is now. Every other mutation in this PR carries its predicate; there was no reason for these two to be the exception.

Both now do:

```ts
const updated = await db.auditNote.updateMany({
  where: {
    id: noteId,
    userId,
    auditAssetId: { not: null },
    auditSession: { organizationId },
  },
  data: { content },
});
if (updated.count === 0) { /* 404 */ }
```

Two consequences worth naming, since they change the contract:

- `updateMany` can't `include` a relation, so the note-with-author the caller needs comes from a **re-read carrying the same scope** — not an unscoped `findUnique`, which would have reopened the same gap on the read.
- `deleteAuditAssetNote` now returns a count rather than the deleted row. Echoing back a row that no longer exists was inventing a value.

### Send the success notification only after deletion succeeds — fixed

Also correct, and this one was a defect my own change introduced: those deletes used to be unconditional, and they're now org-scoped and can throw 403. So the notification firing first would have told the user it worked while the response said otherwise. Moved after the mutation in both `assets.$assetId.note.tsx` and `audits.$auditId.note.tsx`.

### Use factories for the authorization cases — declining

These inputs are three bare scalars (`"note-1"`, `"user-1"`, `"org-1"`) passed straight to the function under test. A factory would add a layer of indirection without preventing anything, and it would push the ownership boundary — the thing each test is actually about — one level away from the assertion.

For consistency: I *accepted* a very similar suggestion on #2882 for Stripe price fixtures. The distinction is how much shape the fixture carries. A price has four properties the validator inspects, and adding one requirement broke 13 hand-written literals at once — that's where a factory earns its keep. Three scalars don't reach that bar.

Happy to revisit if these grow.

PR #2893, opened 2026-08-18
fix(bookings): pass the user's timezone to the assets-index booking dialog

carlosvirreira:
## The bug

On the assets index, **Book selection → Create new booking** refuses to create a booking that starts soon, for any user **west of UTC**, with:

> Start date must be in the future

The form pre-fills the start time with **now + 10 minutes**, so its own default value fails its own validation. The user types nothing and is blocked on the first click. They then have to push the start time out by the length of their UTC offset before it is accepted.

## Cause

`CreateBookingForSelectedAssetsDialog` built its schema without `hints`:

```tsx
BookingFormSchema({
  action: "new",
  workingHours,
  bookingSettings,
  isAdminOrOwner: isAdministratorOrOwner,
})
```

`hints` was optional, and `coerceLocalDate` falls back to UTC when it has no zone:

```ts
const zone = timeZone ?? "UTC";
```

So that dialog read the typed wall-clock as UTC. For a user at UTC-5, a typed 15:00 became 15:00Z, which is 10:00 their time, which is in the past.

`validateFutureDate` compares against a real `new Date()`, so the rejection window is exactly the user's UTC offset.

## Blast radius

| Zone | Blocked window (summer) |
|---|---|
| America/Sao_Paulo (UTC-3) | 3 hours |
| America/New_York (UTC-4) | 4 hours |
| America/Chicago (UTC-5) | 5 hours |
| America/Denver (UTC-6) | 6 hours |
| America/Los_Angeles (UTC-7) | 7 hours |

Each gets one hour worse in winter. Users east of UTC are not blocked by the future check, but with working hours enabled they can get spurious "Time must be between X and Y" errors from the same dialog.

Every other booking entry point is correct: the new-booking page, the edit form, the duplicate dialog, and the three mobile API routes all pass the resolved zone.

**No data was stored wrong.** The server re-parses in the resolved preference zone (`resolveUserFormatPrefsById`), so this only ever blocked people. Nothing to repair, no migration.

## When it started

`00f675b78` (2026-05-14) moved the schema from `z.coerce.date()` to `coerceLocalDate`. A browser reads `new Date("2026-08-18T15:00")` as **local** time, so this dialog was correct before that commit and wrong after it.

`3ca4ef1a6` (2026-07-17) routed the new-booking page, the edit form and the duplicate dialog through the resolved preference zone. This caller was missed because `hints` was optional, so nothing failed at build time.

## The fix

1. The dialog now passes `{ ...hints, timeZone: prefs.timeZone }`, identical to the new-booking and edit forms.
2. **`hints` is now required** on `BookingFormSchemaParams`. An omitted zone is a compile error instead of a silent UTC fallback.

Verified the guardrail bites: reverting only the dialog change makes `tsc` fail on the exact line that shipped the bug.

```
create-booking-for-selected-assets-dialog.tsx(42,23): error TS2345:
  Property 'hints' is missing in type '{ action: "new"; ... }'
  but required in type 'BookingFormSchemaParams'.
```

3. The 18 test call sites that were relying on the implicit UTC now state their zone explicitly.

## Tests

New regression suite in `forms-schema.test.ts`, clock pinned to 13:00 America/Chicago:

- a 15:00 start is accepted when the user's own zone is passed
- it resolves to `2026-08-18T20:00:00.000Z`, not `15:00Z`
- the same start is rejected when the zone is lost to UTC
- the rejection window is exactly the UTC offset (14:00/17:00/18:00 rejected, 19:00 accepted)

`pnpm webapp:validate`: 380 files, 5063 tests pass, 0 lint errors, typecheck clean.

## Verification note

The diagnosis and the cliff were reproduced on production before the fix: with the account time zone set to `America/Chicago`, the same start time was **accepted** by Bookings → New booking and **rejected** by this dialog, and the dialog only started accepting once the start was more than 5 hours out.

The fixed dialog was **not** clicked through in a local browser: the shared dev database is currently drifted from `main` (`column a.stockStatus does not exist`) so the assets index does not load, and it carries migrations from several in-flight branches, so I did not want to migrate it. The change makes this dialog use the exact expression already proven correct on the new-booking page in the same render context, and the guardrail is enforced by the compiler. Worth one click on staging before merge.

## Reproduce (2 minutes, any location)

1. Account details → General → Language & region → Time zone → `America/Chicago`. Save. Note the "Dates will look like" clock, call it T.
2. Assets → tick one asset → Book selection → Create new booking.
3. Start = today at T + 1 hour, End = today at T + 3 hours, any custodian → Confirm.
4. Before: "Start date must be in the future". After: the booking is created.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved booking date validation across time zones.
  * Corrected handling of near-future bookings in regions west of UTC.
  * Added clearer validation when timezone information is unavailable.
  * Preserved working-hours checks and timezone-aware date parsing.

### DIFF
--- a/apps/webapp/app/components/assets/assets-index/create-booking-for-selected-assets-dialog.tsx
+++ b/apps/webapp/app/components/assets/assets-index/create-booking-for-selected-assets-dialog.tsx
@@ -15,11 +15,13 @@ import { Button } from "~/components/shared/button";
 import { Card } from "~/components/shared/card";
 import { TagsAutocomplete } from "~/components/tag/tags-autocomplete";
 import { useBookingSettings } from "~/hooks/use-booking-settings";
+import { useFormatPrefs } from "~/hooks/use-format-prefs";
 import { useUserData } from "~/hooks/use-user-data";
 import { useWorkingHours } from "~/hooks/use-working-hours";
 import { useUserRoleHelper } from "~/hooks/user-user-role-helper";
 import { getBookingDefaultStartEndTimes } from "~/modules/working-hours/utils";
 import type { AssetIndexLoaderData } from "~/routes/_layout+/assets._index";
+import { useHints } from "~/utils/client-hints";
 import { getValidationErrors } from "~/utils/http";
 import { userCanViewSpecificCustody } from "~/utils/permissions/custody-and-bookings-permissions.validator.client";
 
@@ -36,10 +38,15 @@ export default function CreateBookingForSelectedAssetsDialog() {
   const bookingSettings = useBookingSettings();
   const { isBaseOrSelfService, roles, isAdministratorOrOwner } =
     useUserRoleHelper();
+  const hints = useHints();
+  // TIMEZONE FIX: client-side date validation uses the RESOLVED pref zone
+  // (matches display + the server parse), not the browser hint.
+  const prefs = useFormatPrefs();
 
   const zo = useZorm(
     "CreateBookingWithAssets",
     BookingFormSchema({
+      hints: { ...hints, timeZone: prefs.timeZone },
       action: "new",
       workingHours,
       bookingSettings,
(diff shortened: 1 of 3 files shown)


DonKoko:
This approach has some flaws because it uses hints instead of prefs properly. This led to me finding some other issues with those booking forms, so It needs further refactoring. Closing this PR and will open a new with the solution.

PR #2899, opened 2026-08-19
feat(audits): snapshot asset title and expectedness on each scan

carlosvirreira:
## ⚠️ Contains a migration

Two **nullable** columns on `AuditScan`. Catalog-only in Postgres: no table rewrite, no long lock, nothing for the Fly `release_command` to block on. **No backfill** — see below.

## The problem

An `AuditScan` stored its descriptive facts **by reference**. The asset's name and whether it was expected lived in `Asset` and `AuditAsset`, both of which vanish when the asset is deleted (`Cascade` on `AuditAsset.asset`, `SetNull` on `AuditScan.asset`).

So deleting an asset silently rewrote history in three places at once:

- the scan row survived pointing at nothing — **blank name**, and badged **"Unexpected"** when the activity feed still said it was an *expected* asset
- the web asset table dropped it entirely ("No assets" above a counter reading "1 of 1 found")
- the session's denormalised counters kept claiming it

That is not three bugs. It is one property: **an audit is a historical record, and its entries could silently lose their meaning.**

It also bypasses the product's own rule. `removeAssetsFromAudit` throws *"Can only remove assets from pending audits"* — the asset set is deliberately frozen once an audit starts. Deleting the asset does at the database level exactly what the API refuses to do.

## The fix

Record the two descriptive facts **by value** at scan time:

- `assetTitle` — written at create, from the already org-verified fetch
- `wasExpected` — written on the existing update, where expectedness is derived from `AuditAsset` (never trusted from the request)

Readers **prefer the live asset** and fall back to the snapshot. A rename should show the current name; the snapshot's job is to survive deletion, not to freeze naming. A new `assetDeleted` flag lets clients distinguish "the asset is gone" from "this row predates the columns".

The phone now shows `Test (deleted)` with the scanned code instead of a blank mystery row.

## Why no backfill

The rows that would benefit are precisely the ones whose asset is already deleted — there is nothing left to backfill *from*. Everything else still resolves through the live asset. A large `UPDATE` inside a migration that auto-runs on deploy buys nothing and risks the release.

## What I deliberately did NOT do

**Recompute the stale session counters.** A trigger would be the tempting fix and it is wrong: it would do by database trigger exactly what `removeAssetsFromAudit` refuses to do by request — silently rewrite a completed audit so it reports finding fewer assets than it found. A stale number is better than a rewritten record.

## Verification

Both apps typecheck and lint clean; the full route-test suite (54 files, 443 tests) passes.

## Summary by CodeRabbit

- **New Features**
  - Audit scans retain asset names and expected-status details after assets are deleted.
  - Audit records identify scans associated with deleted assets.
  - Scanned QR codes appear on asset cards.
  - Deleted and unexpected scans display clearer, more reliable labels.
- **Bug Fixes**
  - Improved handling of deleted and legacy scan records to prevent missing or misleading audit information.
  - Audit details now apply more consistent access controls for assigned and privileged users.


DonKoko (changes requested):
Really nice piece of work — the framing ("an audit is a historical record") is the
right one, and it shows in the details. A few things I checked rather than took on
trust, all of which hold up:

- the migration really is catalog-only — both columns nullable with no default, so
  no rewrite and nothing for the `release_command` to sit on
- `assetTitle` comes from the org-verified fetch (the fetch-then-verify at
  `service.server.ts:1294`), so the snapshot can't be poisoned cross-org
- `wasExpected` is always written: there's no early `return` or `throw` between
  the create and the update, so the pair can't come apart
- `assetDeleted: scan.assetId === null` is sound — every creation path sets
  `assetId` (the one service site and the seeder), so a null genuinely means the
  `SetNull` fired rather than "never linked"

The `deleted:${code}` id in the companion is a good catch; every deleted row would
otherwise share `""` and collide in the `keyExtractor`. And declining to recompute
the counters is the right instinct, for exactly the reason you give.

Two things I'd like resolved before this goes in.

---

### 1. The description opens with three symptoms and the diff addresses one

You fix the scan row, and you explicitly park the counters. The middle one — *"the
web asset table dropped it entirely ('No assets' above a counter reading 1 of 1
found)"* — isn't addressed and isn't mentioned.

It also can't be reached from here: `expectedAssets` is built from `session.assets`
filtered on `auditAsset.asset` (`service.server.ts:767`), and `AuditAsset.asset` is
`onDelete: Cascade`. The row is destroyed along with the asset, so snapshotting
`AuditScan` can't bring it back to that table.

I don't think it has to be fixed in this PR. But the body reads as though all three
symptoms are covered, and the next person to hit it will reasonably believe they're
looking at a regression. Could it either get the fix or a line under "What I
deliberately did NOT do"?

### 2. The two surfaces now describe the same row differently

`getAuditScans` has two consumers — `audits.$auditId.scan.tsx:317` and
`api+/mobile+/audits.$auditId.ts` — and only the mobile route forwards
`assetDeleted`.

The web page does pick up the title and expectedness fallbacks for free, so the data
corruption is fixed there too. What it can't do is *say* the asset is deleted: it
renders a snapshot title with no indication anything is gone, while the phone shows
`Test (deleted)`. Same row, two stories, and the web one is the one that looks like
a normal asset that has quietly lost its badge.

---

### Notes, not blockers

**The name can visibly change at deletion.** While the asset lives a rename shows
the current name; once it's deleted the row reverts to the scan-time snapshot. I
think that's the right trade and your comment argues it well — worth a sentence in
the schema doc so the next reader doesn't file it as a bug.

**The seeder doesn't write the new columns.**
`seed-reporting-demo/phases/audits.ts:274` creates scans without `assetTitle` /
`wasExpected`, so demo data only ever exercises the live-asset path and never the
fallback you added.

**For anyone else reading:** CodeRabbit's summary says *"Audit details now apply
more consistent access controls for assigned and privileged users."* There are no
access-control changes anywhere in this diff — that line is wrong and worth
ignoring.

carlosvirreira:
Both out-of-diff findings from the last two reviews are addressed in 75900ad83.

**Deleted-scan row keys (`audits/[id].tsx`) — fixed.** Confirmed: the key was `deleted:${scan.code || scan.scannedAt}`, and `AuditScan.code` is `String?` with no unique constraint (schema.prisma) and is flattened to `""` on the wire, so a code-less deleted scan fell through to `scannedAt`, whose milliseconds two scans can share. The scan **row id** is the identity that actually survives asset deletion, so it now rides `AuditScanData` end to end — added to the type and the `getAuditScans` map in `audit/service.server.ts`, to the `existingScans` mapper in `api+/mobile+/audits.$auditId.ts`, and to the companion wire type — and the list key uses it. Two notes on the shape: the `findMany` already used `include`, so `scan.id` was in hand and no query changed; and the companion field is optional with the old `code || scannedAt` fallback retained, because a companion build in the store today talks to servers that do not send it yet.

**Test factories and mock comments (`audit/service.server.test.ts`) — half taken.** The `// why:` half is right and is fixed: the `beforeEach` session mock had no comment at all (out of step with this file's own convention — the equivalents at lines 569, 685, 879, 1045 and 1249 all carry one), and the comment above the deleted-scan mock carried real rationale without the `// why:` prefix. Both now conform, and the mocked rows gained ids with an assertion pinning the passthrough.

The factory half is **not** being applied, for two reasons: there is no `AuditScan` factory to use — `test/factories/audit.ts` exports only `createAuditSession` and `createAuditAsset`, and no `createAuditScan` exists anywhere under `test/` — and these fixtures are not plain Prisma rows but the `getAuditScans` `findMany` projection with nested `asset`, `auditAsset` and `_count`, so a row factory would not drop in. Inventing one for three fixtures in a single suite would be more indirection than it removes.

`pnpm --filter @shelf/webapp test -- --run app/modules/audit/service.server.test.ts` passes 108/108, and both apps typecheck clean.

PR #2925, opened 2026-08-24
fix(exports): serve every CSV download as UTF-8

carlosvirreira:
## Problem

CSV exports were served as `text/csv` with no charset and no UTF-8 byte order mark.

Spreadsheet applications pick a CSV's encoding from its first bytes. With no mark, Excel decodes the file with the machine's locale codepage, so any non-Latin content opens as mojibake: Arabic, Cyrillic, Greek, CJK, and accented Latin. Reported against the Asset Distribution report export, where Arabic category and location names came out unreadable.

The bytes we send were always valid UTF-8. The file just never said so.

## Fix

New `apps/webapp/app/utils/csv-utf8.ts`:

- `withUtf8Bom(csv)` prefixes the mark, idempotently.
- `csvResponse(csv, init)` builds the download response: adds the mark and sets `text/csv; charset=utf-8`, merging whatever headers the route supplies (`content-disposition`, caching).

Every CSV loader now returns through `csvResponse`, so the encoding is a property of the download rather than a per-route decision a new column or a new export can miss:

| Surface | Route |
| --- | --- |
| Report exports (all 11) | `reports.export.$fileName[.csv].tsx` |
| Asset index / backup / import-ready export | `assets.export.$fileName[.csv].tsx` |
| Bookings export | `bookings.export.$fileName[.csv].tsx` |
| Non-registered members export | `settings.export-nrm.$fileName[.csv].ts` |
| Admin org-assets backup | `admin.export-org-assets.$organizationId.$fileName[.csv].tsx` |
| Asset / booking / audit / location activity notes | four `*.activity[.csv].ts` routes |

The bulk-update result report is built client-side, so it gets the same treatment on its `Blob`.

## Re-import safety

An exported file stays re-importable. `parseCsv` runs `csv-parse` with `bom: true` behind `chardet` detection, and the bulk-update client parser strips a leading `﻿` — both already handled the mark, because files coming back from Excel carry one.

## Tests

- `app/utils/csv-utf8.test.ts` — the mark is emitted, applied once, and lands as the three bytes `EF BB BF`; caller headers survive; a caller-supplied content type is overridden.
- `test/routes-tests/_layout+/reports.export.$fileName[.csv].test.ts` — new. Drives the real loader for the distribution report with Arabic group names and asserts the response bytes, the charset, and the row content.
- `test/routes-tests/_layout+/assets.$assetId.activity[.csv].test.ts` — asserts the mark on the wire.

Assertions read the body as bytes rather than through `response.text()`: UTF-8 decode drops a leading mark by spec, so a decoded body cannot show whether the file written to disk has one.

Full webapp suite green: 406 files, 5385 tests.

## Summary by CodeRabbit

* **New Features**
  * CSV downloads now use UTF-8 encoding for improved spreadsheet compatibility.
  * Exported CSV files include the correct UTF-8 content type and encoding marker.
  * Multilingual content, including Arabic characters, is preserved in exports.

* **Bug Fixes**
  * Standardized CSV downloads across asset, booking, audit, location, report, and settings exports.
  * Improved CSV re-import compatibility while preserving filenames and download behavior.

### DIFF
--- a/apps/webapp/app/components/assets/bulk-update/results-display.tsx
+++ b/apps/webapp/app/components/assets/bulk-update/results-display.tsx
@@ -5,6 +5,7 @@
  *
  * @see {@link file://./form.tsx} Parent orchestration component
  */
+import { CSV_CONTENT_TYPE, withUtf8Bom } from "~/utils/csv-utf8";
 import type { BulkUpdateResult } from "~/utils/import-update.server";
 import { escapeCsvValue } from "./helpers";
 import { SummaryPill } from "./shared";
@@ -62,7 +63,7 @@ export function ResultsDisplay({
     }
 
     const csv = lines.join("\n");
-    const blob = new Blob([csv], { type: "text/csv" });
+    const blob = new Blob([withUtf8Bom(csv)], { type: CSV_CONTENT_TYPE });
     const url = URL.createObjectURL(blob);
     const a = document.createElement("a");
     a.href = url;
--- a/apps/webapp/app/routes/_layout+/assets.$assetId.activity[.csv].ts
+++ b/apps/webapp/app/routes/_layout+/assets.$assetId.activity[.csv].ts
@@ -1,6 +1,7 @@
 import { data, type LoaderFunctionArgs } from "react-router";
 import { z } from "zod";
 import { db } from "~/database/db.server";
+import { csvResponse } from "~/utils/csv-utf8";
 import { exportAssetNotesToCsv } from "~/utils/csv.server";
 import { makeShelfError } from "~/utils/error";
 import { buildContentDisposition, error, getParams } from "~/utils/http.server";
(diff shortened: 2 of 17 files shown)


carlosvirreira:
CodeRabbit nitpick triage:

**1. Historical JSDoc on `buildCsv` — fixed** (7bc606c). The block narrated how escaping used to be applied per field and what it missed. Rewritten as the standing contract: every cell passes through `escapeCsvField`, and callers pass raw values because a pre-escaped cell is escaped twice. Per `.claude/rules/comments-describe-code-not-history.md`.

**2. Test-data factories for the CSV fixtures — declining.** Each fixture is used once, in the file that declares it, and nothing varies between scenarios. A factory here would add a layer of indirection to produce the same constant every call, which makes the Arabic content harder to see, not easier to maintain. The tests are readable as-is.

DonKoko (approved):

PR #2958, opened 2026-08-27
docs(companion): record where the 1.5.0 builds are

carlosvirreira:
Updates the release-status file now the builds exist.

| | Build | State |
| --- | --- | --- |
| iOS | 36 | Uploaded to App Store Connect, in TestFlight. Not sent to review. |
| Android | 16 | Built, not submitted — needs `google-service-account.json`. |

Both from `27dbeeda7`. Also records what has actually been exercised on a device — private-server sign-in has been, by DonKoko; the camera paths have not, since every other check was a simulator or manual code entry — and the App Review prerequisite for sending iOS to review.

## Summary by CodeRabbit

* **Documentation**
  * Updated companion release status for version 1.5.0.
  * Added iOS and Android build details, including submission status.
  * Documented device and simulator verification coverage.
  * Added App Review requirements for the demo domain and credentials.

### DIFF
--- a/COMPANION-RELEASE-STATUS.md
+++ b/COMPANION-RELEASE-STATUS.md
@@ -25,7 +25,20 @@ Fill each one in from the store listing once that build is actually live.
 
 ## In flight
 
-**1.5.0 — cut, not yet submitted.**
+**1.5.0 — built; iOS in TestFlight, Android not submitted.**
+
+|         | Build | State                                                             |
+| ------- | ----- | ----------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


carlosvirreira:
Superseded by #2959, which deletes the file instead of updating it.

You are right on all three points. It restates what App Store Connect and the Play Console already own, so it can only go stale; naming the demo domain and saying App Review credentials exist on it does not belong in a public repo; and the monorepo root was the wrong home for a companion-only file.

The one durable line it carried is now a rule in `apps/companion/STORE-RELEASE.md` §2 — write release notes from what is an ancestor of the build commit — with the command to check it.

PR #3008, opened 2026-09-08
fix(webapp): add Content-Disposition header to list/bulk CSV export routes

### LINKED ISSUE #2585: Add Content-Disposition to the 4 list/bulk CSV export routes (helper exists, already used by 5 others)
## Summary

Four CSV export routes return `content-type: text/csv` but **no `Content-Disposition` header**, so the downloaded filename is set only client-side by the export button. Hitting the export URL directly (or any download path that doesn't go through the button) yields a file with no name/extension — the OS can't open it.

This is an **inconsistency**, not a missing capability: a dedicated, tested helper already exists and is used by 5 sibling exports. These 4 just weren't wired up.

## Evidence (verified on `main` @ current HEAD)

A shared helper exists and is unit-tested (12 cases, incl. RFC 5987 non-ASCII):
- `apps/webapp/app/utils/http.server.ts:329` — `buildContentDisposition(name, { fallback, suffix })`
- Tests: `apps/webapp/app/utils/http.server.test.ts:395+`

**Already using it (5):** all activity exports + reports export
- `assets.$assetId.activity[.csv].ts`, `bookings.$bookingId.activity[.csv].ts`, `locations.$locationId.activity[.csv].ts`, `audits.$auditId.activity[.csv].ts`, `reports.export.$fileName[.csv].tsx`

**Missing it (4):**
| Route | File |
|---|---|
| Bookings export | `app/routes/_layout+/bookings.export.$fileName[.csv].tsx:52` |
| Assets export | `app/routes/_layout+/assets.export.$fileName[.csv].tsx:63` |
| NRM export | `app/routes/api+/settings.export-nrm.$fileName[.csv].ts:39` |
| Admin org-assets export | `app/routes/api+/admin.export-org-assets.$organizationId.$fileName[.csv].tsx:25` |

Their buttons currently compensate with `link.setAttribute("download", ...)` (e.g. `export-bookings-button.tsx:55`, `export-assets-button.tsx:53`, `export-nrm-button.tsx:46`), so the **normal UI flow is unaffected** — this only bites non-button access.

## Impact

- **Low/medium.** Real but narrow: the in-app Export buttons work today. The gap shows up for direct-URL hits, API/automation consumers, or any future caller that doesn't set the filename client-side.
- **Not a security issue**; purely download ergonomics + consistency.

## Suggested fix

Add the header to each of the 4 routes using the existing helper — the route already has `$fileName` in its path, so use that (or the entity name) as the source:

```ts
return new Response(csvString, {
  status: 200,
  headers: {
    "content-type": "text/csv",
    "content-disposition": buildContentDisposition(fileName, { fallback: "export" }),
  },
});
```

No new code, no migration — reuse what's tested. Could also drop the now-redundant client-side `download` attribute, but that's optional and can stay for belt-and-suspenders.

## Why this is filed separately

Spotted while verifying PR #2579 (per-asset check-in columns in the bookings export). It's pre-existing and out of scope for that PR; filing as a standalone follow-up rather than scope-creeping the feature.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Nofates:
## Summary

Fixes #2585 — four CSV export routes return `text/csv` without a `Content-Disposition` header, so hitting the export URL directly yields a file with no name/extension.

This wires the existing, already-tested `buildContentDisposition` helper (used by 5 sibling exports) into the 4 routes that were missing it.

## Changes

| Route | Fallback filename |
|---|---|
| `bookings.export` | `bookings-export-<ts>.csv` |
| `assets.export` | `assets-export-<ts>.csv` |
| `settings.export-nrm` | `nrm-export-<ts>.csv` |
| `admin.export-org-assets` | `assets-org-export-<ts>.csv` |

## Verification

- `tsc --noEmit` passes (needs `NODE_OPTIONS=--max-old-space-size=6144` on constrained VMs)
- `eslint` clean on the 4 touched files
- Webapp unit test suite passes (booking partial-scan tests etc.)

## Summary by CodeRabbit

- **Improvements**
  - CSV exports for assets, organization assets, and NRM data now use the requested filename when provided.
  - When no filename is specified, downloads use a clear data-type-based fallback name.
  - CSV downloads retain the `.csv` extension and are easier to identify.
- **Documentation**
  - Added documentation for CSV export routes and download behavior.
- **Tests**
  - Expanded coverage for filename handling, header safety, and CSV download requirements.

### DIFF
--- a/apps/webapp/app/routes/_layout+/assets.export.$fileName[.csv].tsx
+++ b/apps/webapp/app/routes/_layout+/assets.export.$fileName[.csv].tsx
@@ -8,7 +8,11 @@ import {
   exportAssetsFromIndexToCsv,
 } from "~/utils/csv.server";
 import { makeShelfError } from "~/utils/error";
-import { error, getCurrentSearchParams } from "~/utils/http.server";
+import {
+  buildContentDisposition,
+  error,
+  getCurrentSearchParams,
+} from "~/utils/http.server";
 import {
   PermissionAction,
   PermissionEntity,
@@ -87,7 +91,14 @@ export const loader = async ({ context, request }: LoaderFunctionArgs) => {
       csvString = await exportAssetsBackupToCsv({ organizationId });
     }
 
-    return csvResponse(csvString);
+    return csvResponse(csvString, {
+      headers: {
+        "content-disposition": buildContentDisposition(null, {
+          fallback: "assets",
+          suffix: "-export",
+        }),
+      },
+    });
   } catch (cause) {
     const reason = makeShelfError(cause, { userId });
     return data(error(reason), { status: reason.status });
--- a/apps/webapp/app/routes/_layout+/bookings.export.$fileName[.csv].tsx
+++ b/apps/webapp/app/routes/_layout+/bookings.export.$fileName[.csv].tsx
@@ -3,7 +3,11 @@ import { z } from "zod";
 import { csvResponse } from "~/utils/csv-utf8";
 import { exportBookingsFromIndexToCsv } from "~/utils/csv.server";
 import { makeShelfError, ShelfError } from "~/utils/error";
-import { error, getCurrentSearchParams } from "~/utils/http.server";
+import {
+  buildContentDisposition,
+  error,
+  getCurrentSearchParams,
+} from "~/utils/http.server";
 import {
   PermissionAction,
   PermissionEntity,
@@ -50,7 +54,14 @@ export const loader = async ({ context, request }: LoaderFunctionArgs) => {
       organizationId,
     });
 
-    return csvResponse(csvString);
+    return csvResponse(csvString, {
+      headers: {
+        "content-disposition": buildContentDisposition(null, {
+          fallback: "bookings",
+          suffix: "-export",
+        }),
+      },
+    });
   } catch (cause) {
     const reason = makeShelfError(cause, { userId });
     return data(error(reason), { status: reason.status });
--- a/apps/webapp/app/routes/api+/admin.export-org-assets.$organizationId.$fileName[.csv].tsx
+++ b/apps/webapp/app/routes/api+/admin.export-org-assets.$organizationId.$fileName[.csv].tsx
@@ -3,7 +3,7 @@ import { z } from "zod";
 import { csvResponse } from "~/utils/csv-utf8";
 import { exportAssetsBackupToCsv } from "~/utils/csv.server";
 import { makeShelfError } from "~/utils/error";
-import { error, getParams } from "~/utils/http.server";
+import { buildContentDisposition, error, getParams } from "~/utils/http.server";
 import { requireAdmin } from "~/utils/roles.server";
 
 export async function loader({ context, params }: LoaderFunctionArgs) {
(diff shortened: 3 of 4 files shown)


Nofates:
Implements #2585 — adds `Content-Disposition` (via the existing `buildContentDisposition` helper) to the four list/bulk CSV export routes: `bookings.export`, `assets.export`, `settings.export-nrm`, and `admin.export-org-assets`. Typecheck and ESLint are clean on the touched files.

apps/webapp/app/routes/_layout+/assets.export.$fileName[.csv].tsx
```diff
@@ -87,7 +91,14 @@ export const loader = async ({ context, request }: LoaderFunctionArgs) => {
       csvString = await exportAssetsBackupToCsv({ organizationId });
     }
 
-    return csvResponse(csvString);
+    return csvResponse(csvString, {
+      headers: {
+        "content-disposition": buildContentDisposition(null, {
```
Comment (DonKoko) [inline apps/webapp/app/routes/_layout+/assets.export.$fileName[.csv].tsx:96]: Passing `null` means the server always picks its own name, and for direct-link downloads the header wins over the `download` attribute. That changes flows that work today:

- Settings → General "Download CSV" (`export-backup-button.tsx`) goes from `assets-2026-09-14-<ms>.csv` to `assets-export-2026-09-14T1230.csv`. The timestamp only goes down to the minute, so two backups in the same minute get the same name. The admin backup link has the same problem.
- Import-ready exports lose their `assets-import-ready-…` name.
- The in-app buttons (fetch + blob) and a direct hit on the URL now give different names for the same export.

Using `params.fileName` here, and in the other three routes, fixes all of these. `buildContentDisposition` always adds a suffix, a timestamp and `.csv`, so it needs an option to take a finished filename as-is.

DonKoko (changes requested):
Thanks for picking this up! The direction is right, but the new header renames some downloads that work today, so two changes before merge:

1. **Use the route's `$fileName` param as the filename** (details inline). That keeps the header in sync with the name the client already builds.
2. **Add a guard so this can't regress.** `test/routes-tests/csv-download-contract.test.ts` already checks that every `[.csv]` route uses `csvResponse`. Extending it to also require a `content-disposition` header would cover these four routes and any future export.

apps/webapp/app/routes/_layout+/assets.export.$fileName[.csv].tsx
Comment (Nofates) [reply apps/webapp/app/routes/_layout+/assets.export.$fileName[.csv].tsx:102]: Addressed in f18b6f8c9: the four export routes now use the route $fileName as the finished download name (with .csv restored), so existing backup/import-ready names and millisecond timestamps are preserved. buildContentDisposition now supports an explicit finished filename, and the CSV route contract test also requires Content-Disposition.

Nofates:
Implemented the requested review changes in `f18b6f8c9`:

- preserve each route's `$fileName` exactly (restoring `.csv`) instead of generating a new minute-resolution name; this keeps backup/import-ready/direct-download filenames consistent
- extend `buildContentDisposition` with an explicit finished-filename path, while keeping the existing generated-name behavior unchanged
- extend `csv-download-contract.test.ts` so every `[.csv]` route must set `Content-Disposition`

Validation: targeted Vitest `66 passed`; staged-file ESLint/Prettier pass; webapp typecheck passes after Prisma client generation; `git diff --check` clean.

@DonKoko re-review when convenient.

Nofates:
Addressed the CodeRabbit follow-up in `4b192a919`: added file-level JSDoc to all touched TypeScript files, documented each exported CSV route loader, and added explicit-filename regression coverage for CR, LF, and quote sanitization. Validation: targeted Vitest 67/67 passed; ESLint/Prettier and webapp typecheck pass; pre-commit hooks pass.

PR #3010, opened 2026-09-09
fix(assets): end the availability bar where the asset was checked in

carlosvirreira:
## What

On `/assets?view=availability`, an asset that was checked in from a live booking (partial check-in) kept that booking's full bar until the booking ended, so the calendar showed it as taken. The booking rules already treat it as free: the picker lists it and a new booking can reserve it. The asset overview also hides the booking. The calendar was the only surface still drawing the old state.

Now the bar of a returned asset ends at its check-in, is drawn with the complete (green) palette, and reads `Returned <time> | booking | custodian` on the bar itself. The popover keeps the booking's real status badge and period and adds a `Returned <time>` line.

## How

- **Signal is the slice.** `BookingAsset.checkedOutAt` / `checkedInAt` per `.claude/rules/booking-checkout-is-recorded-per-slice.md`. Simple mode already returns them (the pivot is an `include`); the advanced-mode subquery adds the two jsonb keys. `AdvancedAssetBooking` and `CalendarExtendedProps` are widened (`returned`, `returnedAt`).
- **Rule** (`resolveReturnedAt` in `use-asset-availability-data.ts`): only ONGOING/OVERDUE bookings; every folded slice of the asset must be back; a check-in is a return only when it is at or after the departure, the same test `isBookingFullyCheckedIn` applies, which keeps rows stamped by the check-in backfill from reading as returned.
- **Bar end** (`returnedBarEnd`): the latest check-in, or one hour after the start when the check-in precedes an unadjusted early-checkout start. FullCalendar drops an end that is not after the start.
- **One source for styling**: `availabilityEventClassNames` in `utils/calendar.ts` is used by the hook and the calendar's `eventClassNames` callback. A returned bar is never treated as a one-day event, so a same-day return keeps its fill. `calendarDisplayStatus` gives the event card its text colour.
- **Hover**: the sweep over a booking's bars keys each bar to its own state (`booking-returned` class), so hovering a returned bar does not paint a still-out sibling green and vice versa.

## Out of scope, stated on purpose

- Kits availability calendar: kit rows keep the full bar; a kit bar needs a per-kit "every member slice returned" fold that its select does not carry (`// why:` at the include).
- Slices on a live booking that never went out (added after departure, or removed and re-added) still draw the planned bar.
- The bookings calendar (`/calendar`) draws one bar per booking and is unchanged.

## Tests

- `use-asset-availability-data.test.ts`: 13 new cases (mixed slices, all returned, RESERVED ignored, QT partial return, stale marker, check-in before start, same-day fill, both loader shapes, latest check-in wins, helper units).
- `calendar-returned-hover.test.ts` (new): `availabilityEventClassNames` and the hover sweep on ONGOING and OVERDUE bookings.
- `event-card.test.tsx`: text colour, `Returned` prefix on returned bars only, popover line with the time, real status badge kept.
- `query.server.test.ts`: the two jsonb keys are projected with `withBookings` and absent without.

`pnpm -F webapp typecheck`, lint on the touched files, and the four touched test files (158 tests) pass.

## Verified in a browser

Local build against the QA database: booking over three days with two individual assets, both checked out, one partially checked in. Month view in both index modes: the returned asset shows the short green `Returned 10:35` bar, the other keeps the full purple bar; the popover shows the ONGOING badge, the booking period and the return time; hovering the purple bar leaves the green one green. A second departure of the returned asset turns its bar purple again. Before/after renders were shared internally.

## Deploy

No migration, no new setting, no companion change. Safe to deploy in any order.

## Summary by CodeRabbit

* **New Features**
  * Availability calendars show returned assets with completed styling and accurate return times.
  * Event cards display a “Returned” indicator and return time when applicable.
  * Calendar hover behavior reflects returned and active booking states.
  * Availability timelines include checkout and check-in details.

* **Bug Fixes**
  * Returned asset bars maintain a visible duration and accurate end time.
  * Booking status colors remain accurate when returned and active bars appear together.

### DIFF
--- a/apps/webapp/app/components/assets/assets-index/use-asset-availability-data.test.ts
+++ b/apps/webapp/app/components/assets/assets-index/use-asset-availability-data.test.ts
@@ -392,8 +392,10 @@ describe("useAssetAvailabilityData — returned slices", () => {
   const BACK = "2026-07-10T15:30:00.000Z";
   // Bar start/end go through the same zone formatter as every bar today
   // (`useHints` is pinned to UTC above), so compare against its output.
+  /** Formats an instant the way the hook formats bar bounds. */
   const iso = (d: string) => toIsoDateTimeToUserTimezone(d, "UTC");
 
+  /** An ONGOING three-day booking row, with per-slice overrides. */
   const ongoing = (overrides: Partial<AdvancedAssetBooking> = {}) =>
     makeBooking({
       id: "b1",
@@ -404,6 +406,7 @@ describe("useAssetAvailabilityData — returned slices", () => {
       ...overrides,
     });
 
+  /** Renders the hook for one asset and returns its single event. */
   const eventFor = (items: unknown) => {
     const { result } = renderHook(() =>
       useAssetAvailabilityData(items as Items)
--- a/apps/webapp/app/components/assets/assets-index/use-asset-availability-data.test.ts
+++ b/apps/webapp/app/components/assets/assets-index/use-asset-availability-data.test.ts
@@ -8,10 +8,9 @@
  *   - Advanced mode (`query.server.ts` raw SQL) aggregates the pivot into a
  *     flat `asset.bookings: AdvancedAssetBooking[]`.
  *
- * The quantities pivot updated the hook + simple loader to the `bookingAssets`
- * shape but left advanced mode on `bookings`, so advanced-mode availability
- * rendered the asset rows but produced zero booking events. These tests lock in
- * that BOTH shapes yield events.
+ * Both loader shapes must yield the same events. These tests lock that in for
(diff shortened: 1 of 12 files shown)


carlosvirreira:
@coderabbitai Both comment items are done in 0555943: `handleEventMouseEnter` / `handleEventMouseLeave` now document that each bar of the booking takes the hover class of its own rendered state (COMPLETE for a bar carrying `RETURNED_EVENT_CLASS`), and the OVERDUE comment states the standing constraint (every surface reads `bookingStatusColorMap`, which maps OVERDUE to the error palette). The same commit adds JSDoc to `getStatusClasses`, `isOneDayEvent`, `useAssetAvailabilityData`, `toMillis` and the three test helpers for the docstring-coverage check.

DonKoko (approved):

PR #3028, opened 2026-09-11
feat(companion): select kits as one unit in booking modes

carlosvirreira:
## What

On the booking screen, "Select to Check Out", "Select to Check In" and "Select to Remove" used to open every kit and give each kit member its own tick box. A kit then read as a long run of tick boxes, and part of a kit could be checked out or removed by accident. The web does the opposite, and the app now mirrors it:

- **Kits stay as the user left them.** Entering a selection mode no longer opens kits, so they stay collapsed unless the user opened one. The header's chevron still opens a kit in select mode, so people can look inside.
- **The kit header tick is the only way to pick a kit.** One tap picks every member the current mode can act on (in check-in mode, only the members still out). A second tap drops them. This is the existing header behaviour, unchanged.
- **Kit members have no tick box and select nothing when tapped** in any selection mode. Outside a selection mode a member still opens its asset. Members of a picked kit keep the "selected" tint, which shows what the kit's tick covers. In a selection mode, inert rows are `disabled`, so a screen reader announces them as dimmed.
- **Standalone assets keep their own tick box.**
- **The floating button names what it acts on**, counting a kit as one thing: `Check Out 1 Kit & 1 Asset`, `Remove 2 Kits`, `Check In 3 Assets`. Its screen reader label uses the same words.

Web references: kits collapsed by default (`booking-assets-column.tsx`), the kit checkbox selects the kit and its members (`kit-row.tsx`), member rows get an empty cell instead of a checkbox (`list-asset-content.tsx`).

## How the button counts

`countSelection` (in `lib/booking-kit-rows.ts`) counts a kit once its header reads "all", meaning every member the mode can act on is picked. Every other selected id counts as an asset. Members are only picked through their header, so in practice these are the standalone rows, and the label always covers everything the submit sends. `describeSelection` turns the counts into Title Case with an ampersand. It shares its "kits first, never empty" phrasing with `describeRemoval` through one private helper, and the removal wording is unchanged.

The server calls are unchanged. A picked kit still travels as its members' asset ids, and `splitRemovalSelection` still decides when a removal may name the kit.

## Known differences from the web

- On the web, a kit member's row menu can remove that one member from the booking. The phone list has no per-row menu, so a member leaves the booking only with its whole kit. This is by design.
- The web can check out part of a kit through its scanner. The phone's scan-to-check-out is #3002; from the phone list a kit goes out whole.
- The check-out and check-in confirm dialogs still count assets ("Check out 5 selected assets?"), because that is what the endpoint receives.
- The remove dialog keeps `describeRemoval`, which names what the removal sends, and `splitRemovalSelection` names a kit only when the booking holds all of it. So a kit the booking holds only in part reads "Remove 1 Kit" on the button, while the dialog counts its assets. If such a kit is picked together with a whole kit, the existing split rule sends both as asset ids, and the dialog counts all of their assets.

## Maestro

- `18-kit-grouping.yaml`: after entering remove mode, it asserts the opened kit stays open and the other stays collapsed. It checks that a member row offers no tick box and that tapping it selects nothing. The kit header tick then reads `Remove 1 Kit`, and adding a standalone asset reads `Remove 1 Kit & 1 Asset`. Header taps are centred, because the list scrolls under the nav bar and the tab bar.
- `07-create-add-reserve-edit-remove.yaml`: reads the new button label (`Remove 1 Asset`). It picks the workspace with a trailing wildcard, since the picker labels the active one "(current)".
- `03b-checkin-skips-unchecked-kit-member.yaml`, `04-checkout-flow.yaml` and `05-checkin-flow.yaml` tick no kit member and are unchanged.

## Tests

- `lib/booking-kit-rows.test.ts` (`node --test`), 12 new cases. Wording: every singular and plural combination, kits-only, assets-only, and the empty form. Counting: a picked kit counts once; a standalone asset beside a kit; two kits and two assets; opening a kit does not change the count; in check-in a kit counts once its outstanding members are picked; members of a partly picked kit count as assets; nothing picked. Five of the seven counting cases fail against the old raw-id count; the other two are cases where both counts agree.
- Companion `typecheck`, `lint` and `test` (131 of 131 pass). React Doctor in diff mode reports no errors.

## Verified on the iOS simulator

Local webapp and Metro from this branch, QA data, an owner account in a team workspace.

- Select to Check Out on a reserved booking with two kits and a standalone asset: both kits stay collapsed with one tick box each. One kit tick plus the standalone asset reads `Check Out 1 Kit & 1 Asset`.
- A kit opened before entering the mode stays open. Its members show no tick box, and tapping one changes nothing. Ticking the header tints the members and reads `Check Out 1 Kit & 1 Asset`. A second tap drops the kit, and the button reads `Check Out 1 Asset`.
- Select to Remove on the same booking and on the fixture booking of flow 18: `Remove 1 Kit`, then `Remove 1 Kit & 1 Asset`.
- Flow 18 passes on the simulator. Flow 07 passes through its create, add and remove steps with the new label. Nothing was submitted from a selection except flow 07's removal of the asset it had just added.

## Ship path

- Client only, in `apps/companion`. No server change, no migration.
- Ships by OTA on runtime 1.5.0. The app reads the same booking payload as before.
- #3002 edits the same screen file for scan-to-check-out, in different regions. Whichever merges second rebases.

## Summary by CodeRabbit

* **New Features**
  * Kit selection is now handled as a single unit from the kit header.
  * Selection actions clearly distinguish kits and individual assets, including combined selections.
  * Eligible users can scan assets to check them out progressively.
* **Bug Fixes**
  * Kit expansion states are preserved when entering selection mode.
  * Kit members can no longer be selected or activated individually during selection.
  * Improved accessibility and interaction behavior for unavailable asset rows.
  * Checkout availability now reflects booking-specific asset quantities and permissions.
  * Updated booking flows to match current workspace and removal confirmation labels.

### DIFF
--- a/apps/companion/.maestro/flows/bookings/07-create-add-reserve-edit-remove.yaml
+++ b/apps/companion/.maestro/flows/bookings/07-create-add-reserve-edit-remove.yaml
@@ -20,7 +20,9 @@ tags:
     visible: ".*Current workspace.*"
     timeout: 8000
 - tapOn: ".*Current workspace.*"
-- tapOn: "GRANULAR WORKSPACE"
+# The picker renders the active workspace as "NAME (current)", so a bare name
+# misses when the session already sits in it.
+- tapOn: "GRANULAR WORKSPACE.*"
 - waitForAnimationToEnd
 
 # ── Create a DRAFT ──
@@ -116,9 +118,9 @@ tags:
 - tapOn: "Select assets to remove"
 - tapOn: ".*3B Scientific.*"
 - extendedWaitUntil:
-    visible: ".*Remove 1 selected.*"
+    visible: "Remove 1 Asset"
     timeout: 10000
-- tapOn: ".*Remove 1 selected.*"
+- tapOn: "Remove 1 Asset"
 - extendedWaitUntil:
     visible: "Remove Selected"
     timeout: 10000
--- a/apps/companion/.maestro/flows/bookings/18-kit-grouping.yaml
+++ b/apps/companion/.maestro/flows/bookings/18-kit-grouping.yaml
@@ -9,6 +9,7 @@ env:
   BOOKING_NAME: "ZZX Stress"
   AUDIO_KIT: "ZZX Audio Package"
   RIG_KIT: "ZZX Projection Rig"
+  LOOSE_ASSET: "ZZX Standalone Fixture 01"
 
 ---
 # A booking that holds kits shows one row per kit, collapsed, with the members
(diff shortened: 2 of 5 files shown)


DonKoko (approved):

carlosvirreira:
@coderabbitai On the outside-diff item "Use kit-aware counts in selection confirmations" (`apps/companion/app/(tabs)/bookings/[id].tsx`, L409-412 and L485-488): verified, both confirms read `selectedAssetIds.size`. This PR keeps that on purpose.

- The floating button names what the user picked and counts a kit as one thing. The confirm is the last step before the partial check-in or check-out call, and it states how many assets that call moves. A kit-aware confirm would read "Check out 1 kit?" for a kit that holds many assets.
- The PR description lists this under "Known differences".

If both steps should name kits, that is a wording follow-up, and it can take its counts from `countSelection` and `describeSelection`.

PR #3035, opened 2026-09-15
fix(assets): skip custody-only booking slices in the availability hook

carlosvirreira:
## Summary

Regression from #3010. The simple-mode assets **table** crashed to the "Oops, something went wrong" boundary whenever a listed asset had been checked back in on a booking that is still ONGOING or OVERDUE. Reported from the field as `Invalid Date` (Safari's wording for `toISOString()` on an invalid `Date`; Chrome says `Invalid time value`).

## Cause

`useAssetAvailabilityData` runs in `AssetsList` on every view, not only the availability calendar. Simple mode ships `bookingAssets` in two shapes:

- the availability include (`data.server.ts`) selects the booking's `from`/`to`;
- the custody-column include (`assetIndexFields`) runs on every simple-mode row that sits on a live booking, table view included, and selects only `id`/`status` and the custodian.

#3010 added the returned-slice rule to the hook. For a slice that is back (`checkedInAt >= checkedOutAt`) on a live booking it calls `returnedBarEnd(returnedAt, booking.from)`. With the custody-column shape `booking.from` is `undefined`, `toMillis` gives `NaN`, and `new Date(NaN).toISOString()` throws inside render. Before #3010 the same shape produced an unused, harmless event.

## Fix

- The hook skips slices whose booking carries no `from`/`to` before any date math (`hasBarPeriod`). Custody data is not a bar.
- `returnedBarEnd` returns the check-in instant when the start cannot be read, so it can no longer throw.

No loader or query changes. No migration.

## Proof

- New tests in `use-asset-availability-data.test.ts` feed the hook the exact custody-column shape. On `main` they fail with `RangeError: Invalid time value` from `Date.toISOString`; with this change 27/27 pass.
- Local dev server, simple mode, one asset checked back in on a booking left ONGOING: `main`'s hook renders the error boundary and the console shows `RangeError: Invalid time value at Date.toISOString` inside `<AssetsList>`; with this change the table renders every row.
- `typecheck`, `lint`, `prettier` clean on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed asset availability views for returned bookings with missing or invalid start times.
  * Prevented errors when displaying custody-only booking records.
  * Ensured bookings without a complete time period no longer generate incorrect availability bars or events.
  * Preserved check-in end times when booking start dates are unavailable or invalid.


carlosvirreira:
Re the two CodeRabbit nitpicks: both valid, both doc-only.

1. `use-asset-availability-data.ts` gets the file-level JSDoc (purpose, both loader shapes, why slices without a period are skipped).
2. The test comment now states the constraint (a custody-only booking has no calendar period, so it never becomes an event) instead of describing the old failure.

The commit is ready locally. Holding the push so the head under review stays stable; it lands on request or with the next touch of this file.

DonKoko (approved):

PR #3040, opened 2026-09-16
fix(api): re-sign expired asset images for the companion

carlosvirreira:
## What

The companion draws an asset photo straight from the signed storage URL the API sends. That URL stops loading once `mainImageExpiration` passes, and the app has no repair flow, so any photo whose link has lapsed shows as an empty tile. #3029 fixed this for kit images. This PR does the same for asset photos on every mobile read path, through the existing shared helper.

Server-side only: no migration and no app update. Phones receive working URLs on their next load.

## Where the photos were blank

| Screen in the app | Read path |
| --- | --- |
| Assets list, My Custody | `api/mobile/assets` |
| Asset detail | `api/mobile/assets/:assetId` |
| Booking detail | `api/mobile/bookings/:bookingId` |
| Add assets to a booking | `api/mobile/bookings/available-assets` |
| Kit detail, member rows | `api/mobile/kits/:kitId` |
| Home, newest assets | `api/mobile/dashboard` |
| Scan card for a QR code, SAM id or barcode | `mobile-code-resolve.server.ts`, `api/mobile/barcode/:value` |
| Asset returned after a quantity or custody change | `getMobileAssetForViewer` |
| Audit rows, phone and web scan tab | `getAuditSessionDetails`, opt-in |

## How

- `refreshExpiredAssetImages` in `modules/asset/service.server.ts` is the one helper every surface calls. It now takes an optional `organizationId` for rows selected without one, and overloads require a workspace from either the rows or the option. It also takes `maxRefreshes` and `timeBudgetMs`. Existing web and report callers pass nothing new and behave as before.
- Every mobile read path and the audit re-sign pass `ASSET_IMAGE_RESIGN_LIMITS`: at most 100 rows, and no new batch once 2.5s have passed, so the worst case is one batch past the budget. Rows not reached keep their stored URL and are repaired on a later load.
- The scanner paths and `getMobileAssetForViewer` share one step, `resignAndShapeMobileAsset` in `mobile-auth.server.ts`. It requires `mainImageExpiration` and re-signs against the workspace that owns the asset.
- `getAuditSessionDetails` re-signs only when a caller passes `refreshExpectedAssetImages: true`, which only the web scan tab and the mobile audit route do, and only for expected rows.
- The booking and kit routes match re-signed rows by index. The kit route signs the kit image and the member photos together.
- The deferred write-back bumps `Asset.updatedAt` on every asset it re-signs. The helper's JSDoc says so.
- No response gains a field: `mainImageExpiration` is dropped again wherever this PR newly selects it.

## Tests

- Asset service: the `organizationId` option, `maxRefreshes`, the time budget with a signer that takes a second per call, and index alignment.
- `mobile-auth.server.test.ts`: the shared step re-signs a lapsed photo against the owning workspace, sends a live one as stored, and `getMobileAssetForViewer` returns a re-signed photo.
- Code resolver: each path hands the step the workspace that owns the code. New `barcode.$value.test.ts` covers the current and a sibling workspace.
- Audit service: `false` signs nothing and `true` signs expected rows only. New `audits.$auditId.photo-signing.test.ts` pins the flag for the layout, overview, activity and scan loaders.
- Booking detail and kit detail payload tests.
- The time budget, the viewer path, the audit flag, the overview flag and the barcode scope were each checked to fail with their fix removed.
- Against the QA database and storage, with a lapsed photo on an expected audit asset: a session-only read served the stored link and left it untouched; a scan-tab read served a working link and renewed it for 72h.
- `pnpm webapp:validate` passes.

## Follow-ups, separate PRs

- `routes/api+/mobile+/asset.refresh-image.$assetId.ts` has its own copy of the re-sign, with a 24h expiry and an unguarded write, and nothing in the companion calls it. Delete it or reduce it to one helper call.
- `mainImageExpiration` on the asset list and detail responses has no reader in the companion.
- A batch signer on Supabase `createSignedUrls` would turn up to 200 signing calls per response into one or two requests, here and on the web lists.
- `barcode.$value.ts` repeats the lookup that `resolveSamShapedBarcode` already does.

## Deploy

Web deploy only. No migration, no OTA.

## Summary by CodeRabbit

- **New Features**
  - Mobile asset photos now automatically refresh when secure URLs expire across asset, booking, kit, dashboard, available-assets, barcode, and QR-code views.
  - Barcode and QR-code results consistently display refreshed asset images, including assets owned by another accessible workspace.
  - Audit scanning refreshes expected asset photos before display.

- **Bug Fixes**
  - Prevented expired image URLs and internal expiration details from appearing in mobile responses.
  - Reduced duplicate photo refreshes and unnecessary refreshes when viewing audit details that do not display expected asset images.


carlosvirreira:
@coderabbitai review

apps/webapp/app/routes/api+/mobile+/assets.ts
```diff
@@ -1,6 +1,7 @@
 import { AssetStatus, type Prisma } from "@prisma/client";
```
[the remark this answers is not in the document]
carlosvirreira: Fixed in dd352aeed: added file-level docblocks to `assets.ts`, `assets.$assetId.ts` and `bookings.$bookingId.ts`, and to `dashboard.ts`, which this PR also touches and had the same gap.

apps/webapp/test/routes-tests/api+/mobile+/bookings.$bookingId.asset-images.test.ts
```diff
@@ -0,0 +1,196 @@
...
+
+// why: booking settings and permission checks feed the action flags, a
+// different contract from the photos under test — stub them to fixed values.
+vi.mock("~/modules/booking-settings/service.server", () => ({
+  getBookingSettingsForOrganization: vi.fn().mockResolvedValue({
+    requireExplicitCheckinForAdmin: false,
+    requireExplicitCheckinForSelfService: false,
+  }),
+}));
+vi.mock("~/utils/permissions/permission.validator.server", () => ({
```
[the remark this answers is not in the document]
carlosvirreira: Fixed in dd352aeed: the permission validator mock now has its own `// why:` comment, and the booking settings mock keeps a separate one.

carlosvirreira:
@coderabbitai the outside-diff finding is addressed in dd352aeed: the asset detail loader docblock and the asset list comment now describe the server-side re-sign instead of a client refresh flow. The booking detail loader docblock, which mentioned only kit images, now covers asset photos too.

carlosvirreira:
@coderabbitai please retry the verification of the outside-diff finding against dd352aeed. The updated comments are the loader docblock in `assets.$assetId.ts`, the comment above the `mainImageExpiration` re-attach in `assets.ts`, and the loader docblock in `bookings.$bookingId.ts`.

apps/webapp/app/modules/audit/service.server.ts
```diff
@@ -872,6 +879,32 @@ export async function getAuditSessionDetails({
...
+    /**
+     * An asset photo is a signed URL that stops loading once
+     * `mainImageExpiration` passes. The web repairs that in the browser, but
+     * the companion audit screen reads this same payload and has no repair
+     * flow, so it draws a lapsed URL as an empty tile. Re-sign server-side
+     * instead, scoped to the workspace that owns the session — which is not
+     * always the caller's own, since a sibling-workspace session resolves here
+     * too.
+     */
+    const refreshedAssetById = new Map(
```
DonKoko: **Blocking: this re-sign runs on web pages that never use it.**

`getAuditSessionDetails` has five callers, and only two read `expectedAssets`:

| Caller | Uses the re-signed photos? |
|---|---|
| `routes/_layout+/audits.$auditId.tsx:271` (layout) | No, `session` only |
| `routes/_layout+/audits.$auditId.overview.tsx:104` | No, `session` only |
| `routes/_layout+/audits.$auditId.activity.tsx:60` | No, `session` only |
| `routes/_layout+/audits.$auditId.scan.tsx:196` | Yes |
| `routes/api+/mobile+/audits.$auditId.ts:73` | Yes |

What that causes:
- **Duplicate signing.** The layout loader runs alongside the child loader, so on the scan tab both sign the same photos (up to 300) at the same time. The write-back is deferred, so neither call sees the other's result.
- **Wasted work before the permission check.** On overview and activity every storage call is thrown away, and it happens before `requireAuditAssigneeForBaseSelfService` (`overview.tsx:124`, `activity.tsx:67`). A BASE or SELF_SERVICE user who is about to be refused still triggers it.
- **Wrong rows.** The input is every `session.assets` row, but only `expected` rows read the result, so unexpected assets use up the `maxRefreshes` budget for nothing.

**Fix:**
- Make the re-sign opt-in and pass it only from the scan loader and the mobile audit route. For example, a required `refreshExpectedAssetImages: boolean`, so every caller has to state which answer applies.
- Sign only rows where `auditAsset.expected` is true.
- Add a test that the layout, overview and activity paths make no `createSignedUrl` call.

apps/webapp/app/modules/asset/service.server.ts
```diff
@@ -5569,13 +5569,31 @@ export function isStorageObjectNotFound(error: unknown): boolean {
...
+ * The most expired asset images one response re-signs.
+ *
+ * Re-signing runs on the request path, one storage call per image. A list with
+ * no pagination (a booking, a kit, an audit) can hold hundreds of assets whose
+ * links have all lapsed, and an uncapped pass could outlast a client's request
+ * timeout. Pass this as `maxRefreshes` on such paths: rows past the cap keep
+ * their stored URL for that response, and because the re-signed rows are
+ * written back, each later load repairs the next slice.
+ */
+export const ASSET_IMAGE_RESIGNS_PER_RESPONSE = 300;
```
DonKoko: **Blocking: a count cap alone does not bound the response time.**

300 rows run as 30 sequential batches of 10, and each row signs a main image and a thumbnail. `createSignedUrl` (`utils/storage.server.ts`) makes up to 3 attempts: 1s pauses on edge or fetch failures, and 1s then 2s of backoff on a 429.

- **Healthy storage:** a few seconds added to booking detail, kit detail, audits and the asset list.
- **Rate-limited or degraded storage:** roughly 2–3s per batch, so well over a minute. That is far past the companion's 20s `REQUEST_TIMEOUT_MS` (`apps/companion/lib/api/client.ts:43`), and the whole screen fails instead of showing a blank photo. Up to 600 signing calls from one request is also a good way to *cause* the 429.

Before this PR these mobile reads made no storage calls at all, so this is a new failure mode for core screens.

**Minimum for this PR:**
- Give the re-sign a time budget. For example, stop starting new batches once ~2–3s have passed, so the worst case is one batch past the deadline. Rows not reached keep their stored URL and are repaired on the next load, the same as rows past the cap.
- Lower the count cap.
- Unit-test that a slow signer cannot hold the call past the budget.

**Preferred:** add a batch signer next to `createSignedUrl` in `utils/storage.server.ts`, built on Supabase's `createSignedUrls(paths, expiresIn)`, and use it inside `refreshExpiredAssetImages`. That is one request per page instead of up to 600. Map per-path errors onto the existing not-found and backoff handling. The web asset index and reports benefit too.

apps/webapp/app/modules/api/mobile-asset-images.server.ts
```diff
@@ -0,0 +1,144 @@
...
+ * this response and repairs them on the next open, which keeps the request well
+ * inside the companion's request timeout.
+ *
+ * @param assets - Rows carrying the image columns, in response order.
+ * @param organizationId - The workspace the rows belong to.
+ * @returns The same rows, with every lapsed URL replaced by a fresh one.
+ * @see {@link file://./../asset/service.server.ts} `refreshExpiredAssetImages`
+ * @see {@link file://./../../routes/api+/mobile+/asset.refresh-image.$assetId.ts}
+ */
+export async function refreshExpiredMobileAssetImages<
```
DonKoko: **Blocking (reuse): extend `refreshExpiredAssetImages` rather than wrapping it.**

This PR ends up with three ways to run the same repair:
1. `refreshExpiredMobileAssetImages` wraps `refreshExpiredAssetImages` to add `organizationId` and the cap.
2. `refreshExpiredMobileAssetImage` wraps (1) and repeats its "unchanged → return the original, otherwise merge" logic.
3. `modules/audit/service.server.ts:891` uses neither and calls the base helper directly, with a spread `organizationId` and its own `Map`.

Nothing here is specific to mobile; the wrapper exists only because these rows don't select `organizationId`.

**Fix:**
- Add the missing inputs to the existing helper, e.g. `refreshExpiredAssetImages(rows, { organizationId, maxRefreshes })`, where the option scopes the write-back for rows that don't carry `organizationId`. Type it so either the rows or the option must supply the org; the write-back must never run with an undefined org. Existing web and report callers stay unchanged.
- Delete this module. Call the base helper directly from every route, the audit service and the shared scanner step. A single row is just `const [asset] = await refreshExpiredAssetImages([row], { organizationId, maxRefreshes })`. That is also how #3029 calls `refreshExpiredKitImages`.

These go away with the module and should not be carried over:
- **The `asset.thumbnailImage === undefined` branch (line 83) and its test** cover a caller that doesn't exist: every select in this PR loads `thumbnailImage`. The base helper never adds that key by itself; this wrapper creates it by turning `undefined` into `null` on the way in (line 64).
- **The comment at lines 69–71 is incorrect.** The base helper returns `assets.map(...)`, so its output lines up with the input by index even when ids repeat. No `Map` is needed.

If any part of this module survives, it needs a file-level JSDoc block (repo rule; it currently starts with imports).

apps/webapp/app/modules/api/mobile-asset-images.server.ts
```diff
@@ -0,0 +1,144 @@
...
+ * @param organizationId - The workspace that OWNS the asset, which for a
+ *   cross-workspace scan is not the caller's.
+ * @returns The row, with a lapsed photo replaced by a fresh one.
+ */
+export async function refreshExpiredMobileAssetImage<
+  T extends {
+    id: string;
+    mainImage: string | null;
+    thumbnailImage: string | null;
+    mainImageExpiration?: Date | null;
```
DonKoko: **Blocking: don't make the expiry optional here.**

All three callers pass rows selected with `MOBILE_ASSET_SELECT`, which always includes `mainImageExpiration`. Marking it optional, and treating a missing value as "not lapsed", means that if the field is ever dropped from a select, the repair silently stops. The compiler says nothing and the photo is blank again.

The test `returns a scanned row with the keys it came in with when nothing lapsed` (`mobile-asset-images.server.test.ts`) locks in exactly that silent no-op, using a row with no expiry, a shape no production caller produces.

**Fix:**
- Require `mainImageExpiration: Date | null` on whatever the scanner step accepts. Once the base helper is extended (see the thread on line 53), that is the shared scan-asset helper.
- Replace that test with one that uses a real, not-yet-lapsed expiry.

apps/webapp/app/modules/api/mobile-code-resolve.server.ts
```diff
@@ -157,6 +158,12 @@ async function resolveSamShapedBarcode({
     return null;
   }
 
+  // The scan card shows the asset's photo, and the companion cannot repair a
+  // signed URL that has lapsed, so re-sign it before the card is built.
+  const barcodeAsset = barcode.asset
+    ? await refreshExpiredMobileAssetImage(barcode.asset, organizationId)
```
DonKoko: **Blocking (reuse): one shared "re-sign, then shape" step for scanned assets.**

The same pair of calls is pasted into four places:
- `modules/api/mobile-code-resolve.server.ts:164` (SAM-shaped barcode)
- `modules/api/mobile-code-resolve.server.ts:247` (SAM id)
- `modules/api/mobile-code-resolve.server.ts:330` (QR)
- `routes/api+/mobile+/barcode.$value.ts:203`

A fifth user of `MOBILE_ASSET_SELECT` + `shapeMobileAssetResponse` was missed: **`getMobileAssetForViewer`** (`modules/api/mobile-auth.server.ts:788`). `asset.adjust-quantity`, `custody.assign-quantity` and `custody.release-quantity` use it to return the asset in their success response, and it still returns the stored URL. If the deferred write-back hasn't landed, a lapsed photo comes back blank after those actions.

**Fix:**
- Add one async helper in `mobile-auth.server.ts`, next to `shapeMobileAssetResponse`. It takes a `MOBILE_ASSET_SELECT` row plus the owning `organizationId`, re-signs, and returns the shaped response.
- Use it at all five sites, so any future scanner path gets the repair automatically.

Optional, if it stays small: `barcode.$value.ts` repeats the lookup that `resolveSamShapedBarcode` already does, which is why this fix had to land twice. Worth merging the two, here or in a follow-up.

apps/webapp/app/routes/api+/mobile+/bookings.$bookingId.ts
```diff
@@ -299,10 +317,30 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
       }
     }
 
+    // A signed image URL stops loading once `mainImageExpiration` passes, and
+    // the companion cannot repair it, so an asset whose photo has lapsed shows
+    // as an empty tile on the booking screen. Re-sign those before the rows are
+    // shaped; the collapse above already reduced them to one row per asset.
+    const refreshedAssetById = new Map(
```
DonKoko: Two smaller things here (the same apply to `kits.$kitId.ts`):

1. **No `Map` needed.** The helper returns rows in input order, so match them by index. Take `const rows = Array.from(byAssetId.values())`, re-sign `rows.map((row) => row.first.asset)`, and read `refreshed[i]` in the `.map` below. That removes the `new Map(…)` + `.get(id) ?? original` block, which is currently copied here, in `kits.$kitId.ts:179` and in `modules/audit/service.server.ts:891`.
2. **Sign asset photos and kit images together (lower priority).** Asset photos are signed here; kit images are signed at line 656, after the `Promise.all` at line 614. Neither needs the other's result, so start this re-sign without awaiting it and await both together at line 656. The `assets` mapping moves down with it. Skip this if it makes the route hard to follow; the time budget on the cap matters more.

apps/webapp/app/routes/api+/mobile+/kits.$kitId.ts
```diff
@@ -169,8 +173,26 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
     // `location` field the companion's kit screen still reads (preserves
     // the existing mobile JSON contract).
     const { assetKits, ...kitData } = kit;
+
+    // Member photos are signed URLs the companion cannot repair either, so the
+    // lapsed ones are re-signed before the kit screen receives them.
+    const refreshedAssetById = new Map(
```
DonKoko: Same as the booking route:

1. **Match by index instead of rebuilding a `Map`.** The helper's output lines up with `assetKits`.
2. **Sign both at once.** The kit image (line 168) and the member photos (here) are signed one after the other. Both only need `storedKit`, so run them with `Promise.all`.

DonKoko (changes requested):
Thanks for picking this up. Re-signing on the server is the right fix for blank photos in the app, and CI is green.

I'm requesting changes for two performance problems and for how the repair is put together. The PR adds a mobile-named wrapper around an existing helper, a single-row wrapper around that wrapper, and a third hand-rolled version in the audit service. The goal is **one shared helper that every surface calls directly**.

Each item below has an inline thread with the evidence and a concrete fix. Suggested order:

### Blocking
1. **The audit re-sign runs on web pages that never use it.** `getAuditSessionDetails` has five callers and three throw the result away; the layout and child loaders sign the same photos at the same time. See `modules/audit/service.server.ts`.
2. **The 300-photo cap assumes healthy storage.** With the retry/backoff in `createSignedUrl`, a rate-limited or degraded Supabase pushes booking, kit, audit and asset-list responses past the companion's 20s timeout. These reads made no storage calls before. See `modules/asset/service.server.ts`.
3. **Extend `refreshExpiredAssetImages` instead of wrapping it**, and delete `modules/api/mobile-asset-images.server.ts`.
4. **One shared "re-sign, then shape" step for scanned assets.** It is pasted into four scanner sites and missing from `getMobileAssetForViewer`. See `modules/api/mobile-code-resolve.server.ts`.
5. **Don't make `mainImageExpiration` optional on the scanner path.** See `mobile-asset-images.server.ts:119`.

### Smaller, same PR
6. **Booking and kit routes:** use the helper's index-aligned output instead of rebuilding a `Map`, and sign kit images and asset photos together. Inline on both routes.
7. **Tests.** The audit change, the scanner paths and the barcode route have no coverage. At minimum:
   - the web audit layout, overview and activity loaders make no `createSignedUrl` call;
   - a lapsed photo on a scanned asset is re-signed through the shared step, including via `getMobileAssetForViewer`;
   - a slow or failing signer cannot hold a response past the time budget from item 2.
8. **Comments.** "The companion cannot repair a lapsed signed URL" is repeated at about 12 call sites. Say it once, in the helper's JSDoc; a call site needs a one-liner at most.

### Follow-up (separate PR, not blocking)
- `routes/api+/mobile+/asset.refresh-image.$assetId.ts` has its own copy of the re-sign logic: a 24h expiry (`oneDayFromNow()`) where everything else uses 72h, and an awaited `db.asset.update` without the guard the shared helper has. The companion never calls it (the only `mainImageExpiration` reference in `apps/companion` is a type). Once this lands, delete the route or reduce it to one call to the shared helper. The same question applies to `mainImageExpiration` on the list and detail responses, which nothing in the companion reads.
- **For awareness:** the deferred write-back is `db.asset.updateMany`, which bumps `Asset.updatedAt`. Opening a large booking, kit or audit on a phone now changes "updated at" on up to 300 assets. The web asset index already does this one page at a time; this PR extends it to unpaginated lists. Fine if intended, but worth one sentence in the helper's JSDoc.

### Checked — no change needed (security)
- The write-back is an `updateMany` guarded on `id` + `organizationId` + the `mainImage` that was read. The signed path comes from the stored URL, never from request input.
- Cross-workspace scans re-sign only after membership is proven (QR) or the lookup is limited to the caller's memberships (barcode), so no non-member receives a working link.
- Booking detail re-signs after draft visibility and the custody gate; the mobile audit route after `requireAuditAssignee`.

**Before re-requesting review:** load an audit's overview, activity and scan tabs on web with lapsed images and confirm only the scan tab signs anything; and run `pnpm webapp:validate`.

apps/webapp/app/modules/audit/service.server.ts
carlosvirreira: Done in 7e584d7e1. `getAuditSessionDetails` now takes a required `refreshExpectedAssetImages`. The scan loader and the mobile audit route pass `true`. The layout, overview and activity loaders pass `false` and make no storage call. Only `expected` rows are signed, so unexpected rows no longer use up the budget, and the result is matched by index.

Tests: `audits.$auditId.photo-signing.test.ts` pins the flag for all four web loaders, and the service test checks that `false` signs nothing and `true` signs only the expected row.

I also ran `getAuditSessionDetails` against the QA database and storage, with a lapsed photo on an expected asset of a real audit. With `false` it served the stored link and left the row untouched. With `true` it served a working link, and the write-back renewed it for 72h. I checked this at the service level, with the loader flags pinned by the route tests, rather than clicking through the tabs, since a web session there needs a human login.

apps/webapp/app/modules/asset/service.server.ts
carlosvirreira: Done in 7e584d7e1, the minimum version. `refreshExpiredAssetImages` takes `maxRefreshes` and `timeBudgetMs`, and every mobile read path plus the audit re-sign pass `ASSET_IMAGE_RESIGN_LIMITS`: 100 rows, and no new batch once 2.5s have passed, so the worst case is one batch past the budget. Rows not reached keep their stored URL and get no backoff, so the next load picks them up. The unit test uses a signer that takes 1s per call over three batches and checks that the third batch never starts.

I left the batch signer on `createSignedUrls` for a follow-up, listed in the description, so this PR does not change the retry handling under the web and report callers.

apps/webapp/app/modules/api/mobile-asset-images.server.ts
carlosvirreira: Done in 7e584d7e1. `refreshExpiredAssetImages(rows, { organizationId, maxRefreshes, timeBudgetMs })`: the overloads require a workspace from the rows or the option, and at runtime a row with neither is never signed, so no write-back runs unscoped. `mobile-asset-images.server.ts` and its test are deleted, including the `thumbnailImage === undefined` branch. The routes, the audit service and the scanner step call the base helper directly. Its JSDoc now says the output lines up with the input by index, and a test covers a repeated id.

apps/webapp/app/modules/api/mobile-asset-images.server.ts
carlosvirreira: Done in 7e584d7e1. The scanner step takes a `MobileAssetSelectRow`, which requires `mainImageExpiration: Date | null`, so dropping it from `MOBILE_ASSET_SELECT` fails to compile. The shaper keeps its own parameter optional only so it can strip the field for callers that build the argument by hand. The no-expiry test is gone, and `sends a photo that has not lapsed as stored` uses a real future expiry.

apps/webapp/app/modules/api/mobile-code-resolve.server.ts
carlosvirreira: Done in 7e584d7e1. `resignAndShapeMobileAsset(row, organizationId)` sits next to `shapeMobileAssetResponse` and is used at all five sites, including `getMobileAssetForViewer`, so the quantity and custody success responses return a working photo. Tests: `mobile-auth.server.test.ts` covers the step and `getMobileAssetForViewer`, the resolver test checks that each path hands the step the owning workspace, and the new `barcode.$value.test.ts` covers the current and a sibling workspace. Merging the duplicate barcode lookup is listed as a follow-up.

apps/webapp/app/routes/api+/mobile+/bookings.$bookingId.ts
carlosvirreira: 1. Done in 7e584d7e1: rows are matched by index.
2. Skipped, as you allowed. The asset rows that need the re-signed photos are built and mapped well before the kit-image await, so signing both together means moving that whole block. The time budget bounds this path instead.

apps/webapp/app/routes/api+/mobile+/kits.$kitId.ts
carlosvirreira: Done in 7e584d7e1: member rows are matched by index, and the kit image and member photos are signed together with `Promise.all`.

carlosvirreira:
@DonKoko thanks for the thorough review. Everything is in 7e584d7e1, and I replied in each thread.

- **Blocking 1 to 5 and item 6:** done as described in the threads. On the booking route I matched rows by index but skipped signing together with kit images, as you allowed.
- **7, tests:** the three you listed, plus the org option, `maxRefreshes` and index alignment on the helper. The time budget, the viewer path, the audit flag, the overview flag and the barcode scope each fail with their fix removed.
- **8, comments:** the companion constraint is stated once, in the helper JSDoc, and call sites keep a one-liner at most. The `updatedAt` effect is noted there too.
- **Follow-ups** are listed in the description and not in this PR: the refresh-image route, `mainImageExpiration` on the list and detail responses, a batch signer, and the duplicate barcode lookup.
- `pnpm webapp:validate` passes. Two older QR route tests mocked `mobile-auth.server` without the new step, so their mocks now include it.

apps/webapp/app/modules/api/mobile-code-resolve.server.test.ts
```diff
@@ -363,5 +376,42 @@ describe("resolveMobileScannedCode SAM-shaped barcode fallback", () => {
...
+describe("resolveMobileScannedCode photo re-sign on the QR path", () => {
+  it("hands the asset to the shared step with the workspace that owns the code", async () => {
+    // why: casts — narrow selected shapes, not full Prisma rows.
+    qrFindUnique.mockResolvedValue({
+      id: "qr-sibling",
+      assetId: "asset-7",
+      kitId: null,
+      organizationId: "org-sibling",
+    } as any);
+    membershipFindUnique.mockResolvedValue({ id: "membership-2" } as any);
```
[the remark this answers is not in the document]
carlosvirreira: Fixed in 7d30cca54. The resolver QR test now uses narrowed read handles instead of `as any`, and both route test fakes type the asset as the pivot fields the shaper flattens and the kit as `unknown`. The `opts: any` in the `ShelfError` mock of those two files predates this PR and is left as it was.

apps/webapp/test/routes-tests/api+/mobile.get-scanned-item.test.ts
```diff
@@ -34,22 +34,11 @@ vitest.mock("react-router", async () => {
...
-  MOBILE_ASSET_SELECT: {
-    id: true,
-    title: true,
-    status: true,
-    mainImage: true,
-    category: { select: { name: true } },
-    location: { select: { name: true } },
-  },
-  MOBILE_KIT_SELECT: { id: true, name: true },
+vitest.mock("~/modules/api/mobile-auth.server", () => {
```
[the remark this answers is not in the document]
carlosvirreira: Fixed in 7d30cca54. Each file's mock rationale now sits directly above `vitest.mock()`, and it also covers the photo re-sign step, which only shapes in these tests.

carlosvirreira:
@coderabbitai on the two outside-diff findings for 7e584d7e1:

- **Removing `mainImageExpiration` from the list, detail and picker payloads:** not in this PR. Those responses already sent the field before this change; this PR only avoids adding it where it newly selects the column. Removing it changes the response contract, which the reviewer asked for as a separate follow-up, listed in the description.
- **Repeated ids:** fixed in 7d30cca54, a slightly different way. A repeated id is the same asset, so the helper now signs it once and applies that result to every row with the id, which also keeps the returned URL equal to the one written back. The test hands out a different URL per signing call and asserts a single call.

carlosvirreira:
@DonKoko heads-up: after re-requesting your review I pushed 7d30cca54 and 53d32b4dd for CodeRabbit notes. They type the test fakes and fix their mock comments, and `refreshExpiredAssetImages` now signs a repeated asset id once. Nothing else changed, and `pnpm webapp:validate` passes. I also merged main into the branch (4787759f7, no conflicts) so it is up to date for merging.

apps/webapp/app/modules/asset/service.server.test.ts
```diff
@@ -655,6 +655,113 @@ describe("refreshExpiredAssetImages", () => {
...
+    // A second signing call would hand out a different URL. A counter rather
+    // than queued once-values, so nothing unused leaks into the next test.
+    let signingCalls = 0;
+    mockCreateSignedUrl.mockImplementation(() =>
+      Promise.resolve(
+        signingCalls++ === 0
+          ? "https://new-signed-url.com"
+          : "https://second-signed-url.com"
+      )
+    );
```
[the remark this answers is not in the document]
carlosvirreira: Fixed in 53d32b4dd: the comment now starts with `why:` and says the second URL is what exposes signing the repeated id twice.

carlosvirreira:
@coderabbitai on the outside-diff finding for 7d30cca54: `modules/asset/service.server.ts` predates this PR without a file-level docblock, and it is a large module with many responsibilities. Writing an accurate header for the whole of it is outside this change, so I left it; the helper this PR changes carries its own JSDoc. Its test file starts with imports, as the kit and audit service tests do, and is left the same way.

carlosvirreira:
Tested end to end on a local build of `4787759f7` against a workspace with six lapsed photo links:

- `GET /api/mobile/assets`: all six came back re-signed (+3 days) and every URL answers 200; the write-back is stored; a second call changes nothing.
- `GET /api/mobile/assets/:id` on an asset forced back to expired: fresh link, loads.
- Web audit pages with lapsed photos: the overview and activity tabs did not go through the server helper (the only refreshes seen were the page's own `refresh-main-image` loader, +1 day); the scan tab re-signed through the helper (+3 days).

Matches the description and the review items.

DonKoko (approved):
Approving. Every item from the last review is addressed, and the merge with main is clean.

**What changed since the review**
- **Audit signing is opt-in.** `getAuditSessionDetails` takes a required `refreshExpectedAssetImages`; only the scan tab and the mobile audit route pass `true`, and only `expected` rows are signed. The layout, overview and activity loaders sign nothing, pinned by `audits.$auditId.photo-signing.test.ts`.
- **Response time is bounded.** `ASSET_IMAGE_RESIGN_LIMITS` caps a response at 100 photos and starts no new batch after 2.5s, so the worst case is the budget plus one batch — inside the companion's 20s timeout even when storage is throttling. The fake-timer test proves the third batch never starts.
- **One helper.** `mobile-asset-images.server.ts` is gone; `refreshExpiredAssetImages(rows, { organizationId, ...ASSET_IMAGE_RESIGN_LIMITS })` is the only entry point, the overloads require a workspace for the write-back, and a repeated asset id is signed once.
- **One scanner step.** `resignAndShapeMobileAsset` covers all five `MOBILE_ASSET_SELECT` paths, including `getMobileAssetForViewer`, which the quantity and custody responses use.
- **The expiry is required** on scanned rows via `MobileAssetSelectRow`, so dropping it from a select fails to compile.

**Merge with main (#3025):** the only conflicts were the import groups in `assets.ts` and `kits.$kitId.ts`, plus the kit member destructure. Resolved by keeping both sides: main's display-code resolution and this branch's photo re-sign, which reads the re-signed member row by index. Both files now differ from `origin/main` by exactly this PR's changes. `pnpm webapp:validate` passes locally on the merged tree (6574 tests, no lint errors), and CI is green on `6af7ef839`.

**Verified:** every changed file including the test bodies, the unchanged callers of the changed helpers, and all 13 mobile routes that return asset photos — each one now goes through the shared helper. Not verified: photos rendering on a physical phone.

**Three small things, none blocking, fine as follow-ups:**
1. `audits.$auditId.scan.tsx` signs photos before its cancelled-audit redirect and before `requireAuditAssigneeForBaseSelfService`, so a refused user still pays for up to 100 signings (within the 2.5s budget, and nothing is sent to them). Fixing it means splitting the service, which is probably not worth it.
2. A few comments inside `refreshExpiredAssetImages` still narrate history ("would previously `await` inline", "Synchronous now"), which the comment rule asks to rewrite when the function is touched.
3. Nothing asserts that the mobile audit route passes `refreshExpectedAssetImages: true`; the required parameter only forces a value to be stated.

The four follow-ups in the description are the right list, especially the batch signer on `createSignedUrls` and deleting the `refresh-image` route.

PR #3059, opened 2026-09-18
fix(sentry): mask email addresses in every payload sent to Sentry

carlosvirreira:
## Summary

Found in the privacy review of `main` for the IT data-flow document. Sentry ingest is in the US region, and user email addresses reached it by several routes:

| Route into Sentry | Example |
| --- | --- |
| `ShelfError.additionalData` → `extra` | The SSO sign-in failure (`utils/sso.server.ts`) sends `{ email, domain }`. `utils/error.ts` warns about exactly this. |
| Error message / cause | `Failed to create SSO user: …`, or any message that quotes an address |
| Request URL | `/otp?email=jane%40acme.com&mode=login`, `/forgot-password?email=…` |
| Breadcrumbs | The request-log `console.log` lines and browser navigation breadcrumbs repeat those URLs |
| Structured logs (`enableLogs`) | `Logger.handledClientError` sends the ShelfError message |
| Transactions | Span descriptions and request URLs for the same pages |

## Change

- New `app/utils/sentry-email-mask.ts` (pure, client-safe):
  - `maskEmailAddresses(text)` replaces the local part and keeps the domain: `jane.doe@acme.com` → `[email]@acme.com`. It also handles the percent-encoded form found in URLs: `jane%40acme.com` → `[email]%40acme.com`.
  - `maskEmailsInSentryPayload(payload)` returns a masked deep copy of an event, transaction or log.
- It's wired into every hook that sends to Sentry:
  - **Server** (`server/instrument.server.ts`): `beforeSend` (after the existing drop/redact logic), `beforeSendTransaction`, and a new `beforeSendLog`.
  - **Browser** (`app/entry.client.tsx`): `beforeSend` (after `handleClientBeforeSend`) and `beforeSendTransaction`.

What is kept:

- **Nothing is dropped.** The error, its tags (`label`, `organizationId`, `shelf_trace_id`), `user.id`, trace id and the rest of `extra` are sent as before, with only the addresses masked.
- **The domain.** It's what triage needs (which workspace's SSO or mail setup is failing), and the SSO error already sends it separately as `domain`. If the data-flow document needs the domain gone too, it's a one-line change to the replacement.
- **Stack frame code locations** (file names, function names, source lines) and `debug_meta` pass through untouched, so symbolication and grouping are unaffected. Captured frame variables (`vars`, only present if `includeLocalVariables` is ever turned on) are masked.

The domain must be dotted and end in letters, so npm-style path segments like `.pnpm/@sentry+core@10.51.0/` aren't mistaken for addresses. `user@localhost` and `@handle` are left alone too.

Only Sentry payloads change. No config, env or migration changes.

I also rewrote a few comments in the touched hooks that narrated past bugs ("the old code…", "PR3's…") as standing constraints, per `.claude/rules/comments-describe-code-not-history.md`.

## Not in this PR

- Sentry's project-level server-side data scrubbing settings (ops/config, out of scope).
- The calendar feed token and invite token in Sentry `request.url` (credentials, not emails). The request-log side is handled in #3058.

## Test plan

- [x] `app/utils/sentry-email-mask.test.ts` (new, 16 tests) covers:
  - plain, percent-encoded, plus-tagged and multi-level-domain addresses
  - non-addresses left alone
  - an SSO-shaped error event with no raw address left anywhere and the rest of the error intact
  - stack frames unchanged, with frame `vars` masked
  - the input not mutated
  - transaction span descriptions and structured-log messages
  - cyclic structures
- [x] End-to-end check (throwaway, not committed): the real `instrument.server.ts` init, a `ShelfError` with the SSO `additionalData`, a request-log line and a `Sentry.logger.info`. The outgoing envelopes contain `[email]@acme.com` / `[email]%40acme.com` and never the original local part in any event, breadcrumb or log field. The exception is the stack frames' source lines, which are the test's own code read from disk.
- [x] `app/utils/sentry-filters.test.ts` still passes.
- [x] eslint, prettier and `tsc -b` are clean.

## Summary by CodeRabbit

* **Privacy**
  * Email addresses are now masked in Sentry error reports, performance transactions, and structured logs before data leaves the application.
  * Email domains remain visible while local parts are replaced, including addresses in URLs, metadata, breadcrumbs, and stack-frame variables.
  * Stack trace locations remain intact to support symbolication.

* **Bug Fixes**
  * Error and transaction reporting continues to exclude handled 4xx errors as expected.


DonKoko (changes requested):
@carlosvirreira Can you please take a look and resolve the coderabbit comments where necessary.

PR #3061, opened 2026-09-18
[Decision needed] chore(analytics): record Clarity sessions on public signed-out pages only

carlosvirreira:
> [!IMPORTANT]
> **Decision needed before this merges.** It's a product/marketing call, not only an engineering one. See [Trade-off](#trade-off) and [Open questions](#open-questions).

## Summary

Found in the privacy review of `main` for the IT data-flow document. Microsoft Clarity session recording (`components/marketing/clarity.tsx`) was initialised from the root layout (`app/root.tsx`), so it ran on **every page, including every logged-in page in the app**. Clarity records the page URL and what is on screen, which means customers' asset, booking and team data could appear in recordings sent to Microsoft.

This PR limits Clarity to public, signed-out pages.

## Trade-off

**No more session recordings of in-app usage.** Once this merges, Clarity will not record:
- anything behind login: assets, bookings, kits, audits, settings, reports, the admin dashboard, and so on
- onboarding: `/onboarding`, `/welcome` and `/select-plan` require a session, so the post-signup part of the funnel is no longer recorded either

What still records is the signed-out funnel and the public QR pages:

| Recorded | Not recorded (public, but excluded) |
| --- | --- |
| `/login`, `/join`, `/otp`, `/forgot-password`, `/sso-login` | `/accept-invite/:id`: the URL carries the signed invite token (`?token=`) |
| `/qr/:qrId/not-logged-in`, `/qr/:qrId/contact-owner` | `/oauth/callback`, `/oauth/callback/mobile`: the URL fragment carries the Supabase access and refresh tokens |
| | `/`, `/qr/:qrId`, `/logout`, `/send-otp`, `/resend-otp`: redirects or form actions only |

## Change

- `CLARITY_PAGES` lists the recordable pages. It's a subset of the server's `publicPaths` in `server/index.ts`. `isClarityPage(pathname)` matches against it.
- `<Clarity />` stays in the root `<head>` but now reads the location:
  - on the first recordable page it loads Clarity (`clarity.init`)
  - when the user navigates to a page that isn't on the list, it calls `clarity.stop()`
  - when they come back to a listed page, it calls `clarity.start()`

  The stop matters because navigation is client-side: after sign-in, the redirect from `/login` or `/otp` into the app doesn't reload the page, so a recording started on `/login` would otherwise continue into the app.
- A first page load inside the app never loads the Clarity script at all.

No config or env changes: `MICROSOFT_CLARITY_ID` is unchanged, and the component still does nothing when it's unset.

## Open questions

1. **Emails in URLs on recorded pages.** `/otp?email=…` and `/forgot-password?email=…` carry the user's email address in the query string, and Clarity records the page URL. They're on the recorded list here to keep the signup/login funnel visible. Removing them is a one-line change to `CLARITY_PAGES` if the data-flow document can't include email addresses going to Clarity.
2. **Onboarding funnel.** If the post-signup steps are what marketing actually uses Clarity for, the alternative is to add those specific routes back to the list explicitly. It's the same mechanism, but those pages are logged-in.
3. **Clarity dashboard filters/segments** that rely on in-app URLs will go empty after this ships.

## Test plan

- [x] `components/marketing/clarity.test.tsx` (new, 17 tests) uses a real router and real client-side navigation, with the Clarity module replaced by a recorder. It covers:
  - `init` on a public page
  - nothing when the first page is in-app
  - `stop` on navigating from `/login` → `/otp` → `/assets`
  - `start` again on return to a public page
  - no recording of the invite page
  - nothing without `MICROSOFT_CLARITY_ID`
  - `isClarityPage` for public, in-app and excluded paths
- [x] The navigation tests fail against the current component (it loads on `/assets` and never stops) and pass with this change.
- [x] eslint, prettier, `tsc -b` and `react-doctor --diff origin/main` are clean.
- [ ] Not verified: behaviour of the live Clarity script after `stop()` / `start()`. These are the calls `react-microsoft-clarity` exposes for pausing and resuming tracking. Worth a quick check in the Clarity dashboard on staging: sign in from `/login` and confirm the recording ends at the redirect.


carlosvirreira:
Closing this one. Nikolay decided to keep Microsoft Clarity recording inside the app: the team uses the recordings and the Clarity API to find and reproduce front-end problems, so limiting it to public pages would remove the part we rely on.

What happened instead:

- Strict masking is now switched on in the Clarity project, so recordings show layout, clicks and page flow rather than the content of assets, bookings or people.
- The privacy policy now names Microsoft Clarity, PostHog and Sentry, says recording is masked and why we keep it, and separates where data lives: website-v2 #316 and #319, both merged.

No consent banner and no per-account switch for now. The log and error-report changes in #3058 and #3059 are unaffected and still wanted.

